PR #2, opened 2012-11-29
Add README.rst

s0undt3ch:


thatch45:
This is awesome btw!
We should post about this to the mailing list

PR #3, opened 2012-12-03
M2Crypto failures when starting minion unless swig installed pre-salt

cedwards:
Installation works fine on FreeBSD, but I'm unable to start the salt minion unless swig was installed before M2Crypto.

### DIFF
--- a/bootstrap-salt-minion.sh
+++ b/bootstrap-salt-minion.sh
@@ -763,10 +763,12 @@ install_freebsd_git_deps() {
 }
 
 install_freebsd_90_stable() {
+    /usr/local/sbin/pkg install -y swig
     /usr/local/sbin/pkg install -y salt
 }
 
 install_freebsd_git() {
+    /usr/local/sbin/pkg install -y swig
     /usr/local/sbin/pkg install -y git salt
     /usr/local/sbin/pkg delete -y salt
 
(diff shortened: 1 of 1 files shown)


akoumjian:
Can you add it to the `_deps` functions instead of the main install function, since it appears to be a dependency?

s0undt3ch:
@cedwards This means it's working in FreeBSD 9 correct? At least with your fix?

cedwards:
CPU_VENDOR needs some massaging on my system (Xeon), which I may have a patch for later tonight, but beyond that the inclusion of the swig dependency fixes a major issue and I'm now able to start the minion.

s0undt3ch:
@cedwards Yes, while working on FreeBSD support I also noticed that the CPU_VENDOR was not that... clean...
That's because there's no `/proc/cpuinfo` on FreeBSD and we're using `sysctl`

s0undt3ch:
Anyway, a fix for that too would be awesome ;)

cedwards:
@s0undt3ch  I'll see what I can come up with tonight.

s0undt3ch:
@cedwards Thank You! Please submit another pull request if you end up cleaning the `CPU_VENDOR` code ;)

thatch45:
Thanks @cedwards, this is a great addition. I am really excited about this script, and the freebsd support.

PR #5, opened 2012-12-08
updated Arch Linux repo config to updated repository

cedwards:
intothesaltmine.org/archlinux is maintained by the
Salt package maintainer and is more current than
red45.org/archlinux.


thatch45:
Woah! Intothesaltmine.org is fantastic!

PR #6, opened 2012-12-10
Adding ppa:saltstack/salt-depends to Ubuntu daily

atrawog:
Ubuntu salt-daily won't install without adding ppa:saltstack/salt-depends


thatch45:
good catch, thanks!

PR #11, opened 2012-12-27
correcting epel install rpm location.

gwaland:
The minor version for epel changed which causes issues with the script making sure epel is installed.   This just updates epel to the current version.

### DIFF
--- a/bootstrap-salt-minion.sh
+++ b/bootstrap-salt-minion.sh
@@ -688,7 +688,7 @@ install_centos_63_stable_deps() {
     else
         local ARCH=$CPU_ARCH_L
     fi
-    rpm -Uvh --force http://mirrors.kernel.org/fedora-epel/6/${ARCH}/epel-release-6-7.noarch.rpm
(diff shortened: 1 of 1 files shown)


akoumjian:
Thanks for the update!

thatch45:
Thanks! I thought that there was a latest link for this rpm, if not we should request it from EPEL, is this the case @herlo ?

herlo:
I think it actually should be https://mirrors.fedoraproject.org/metalink?repo=epel-$releasever&arch=$basearch. This points to the metalink mirror, which in turn will give a good list of possible download points.

thatch45:
will that translate on the command line?

herlo:
Of course it will, try it out. :) Just edit /etc/yum.repos.d/epel.repo and replace the 6 (or 5 if you are using el5) with $releasever, type 'yum clean all' and perform a 'yum install znc' (or something useful). I just tested it here and it appears to work just fine.

PR #16, opened 2013-01-10
default bootstrap works on Centos 6.2

gotcha:
I only added `install_centos_62_stable_deps`, `install_centos_62_stable`, and `install_centos_62_stable_post`. (I did not test `daily` or `git` so I did not add support for it.)

It did work on my machine. Is there a test suite to confirm everything is ok ?

### DIFF
--- a/bootstrap-salt-minion.sh
+++ b/bootstrap-salt-minion.sh
@@ -701,6 +701,18 @@ install_centos_63_stable_post() {
     /etc/init.d/salt-minion start
 }
 
+install_centos_62_stable_deps() {
+    install_centos_63_stable_deps
+}
+
+install_centos_62_stable() {
+    install_centos_63_stable
+}
+
+install_centos_62_stable_post() {
+    install_centos_63_stable_post
+}
+
 install_centos_63_git_deps() {
     install_centos_63_stable_deps
     yum -y install git PyYAML m2crypto python-crypto python-msgpack python-zmq python-jinja2 --enablerepo=epel-testing


akoumjian:
I don't know if we have a test suite yet. @s0undt3ch has been working on a lot of that stuff. Do you have a pre-made 6.2 base box I can test this on? Thanks!

s0undt3ch:
@gotcha The only distro which has daily package available is Ubuntu.
@akoumjian No test suite for salt-boostrap yet! It's being worked on ;)

dlindquist:
FWIW, I verified that this changed worked on Centos 6.2 as well.

s0undt3ch:
Any reason why we can't have these same functions without the distro version?

``` bash
install_centos_stable_deps() {
    if [ $CPU_ARCH_L = "i686" ]; then
        local ARCH="i386"
    else
        local ARCH=$CPU_ARCH_L
    fi
    rpm -Uvh --force http://mirrors.kernel.org/fedora-epel/6/${ARCH}/epel-release-6-8.noarch.rpm
    yum -y update
}

install_centos_stable() {
    yum -y install salt-minion --enablerepo=epel-testing
}

install_centos_stable_post() {
    /sbin/chkconfig salt-minion on
    /etc/init.d/salt-minion start
}
```

akoumjian:
Probably wouldn't work in 5.x would be my guess.

s0undt3ch:
Hmm, yes...
Ok.

s0undt3ch:
Since we have 2 persons saying that it works, let's merge it! ;)

PR #26, opened 2013-01-24
Copy master configuration as well as minion

techhat:
The salt-cloud project has users (including me) who would like to be able to deploy masters, as well as minions. This should be all the support that salt-bootstrap needs to do so.

### DIFF
--- a/bootstrap-salt-minion.sh
+++ b/bootstrap-salt-minion.sh
@@ -907,14 +907,18 @@ config_minion() {
     fi
 
     PKI_DIR=/etc/salt/pki/minion
+    MASTER_PKI_DIR=/etc/salt/pki/master
 
     # Let's create the necessary directories
     [ -d /etc/salt ] || mkdir /etc/salt
     [ -d $PKI_DIR ] || mkdir -p $PKI_DIR && chmod 700 $PKI_DIR
+    [ -d $MASTER_PKI_DIR ] || mkdir -p $MASTER_PKI_DIR && chmod 700 $MASTER_PKI_DIR
 
     # Copy the minions configuration if found
     [ -f "$TEMP_CONFIG_DIR/minion" ] && mv "$TEMP_CONFIG_DIR/minion" /etc/salt
 
+    # Copy the masters configuration if found
+    [ -f "$TEMP_CONFIG_DIR/master" ] && mv "$TEMP_CONFIG_DIR/master" /etc/salt
 
     # Copy the minion's keys if found
     if [ -f "$TEMP_CONFIG_DIR/minion.pem" ]; then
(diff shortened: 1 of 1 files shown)


s0undt3ch:
Since the discussion is now public, why not also syndic support?

Also, @techhat, this is the first step, correct?
Since you're only copying the configuration files.

s0undt3ch:
We should also add a flag(`--master`) to **also** install the master right?

akoumjian:
That's the first step, but do we also want the script to install the master as well?

akoumjian:
Yes, then we would need the current functions to handle additional logic for the --master flag.

s0undt3ch:
@akoumjian Don't know if you missed my comment or you're replying to me :)

`--master` should also install the master.

s0undt3ch:
@akoumjian installing the master should be pretty simple, but yes, an `if` to check if master should be installed.

akoumjian:
It was a reply. :smile:

s0undt3ch:
@techhat want me to merge this as it is now? We'll add `--master` support afterwards. Or do you feal like you want to do some more `sh` coding and add that?

s0undt3ch:
Should our soon to be `master` branch already include this?

techhat:
Holy can of worms, Batman! I meant to include the master in this pull, and completely spaced it. However, unless this pull breaks something else, I'm for just merging it now... it still accomplishes the big part of what I need on my end.

s0undt3ch:
This change will also mean renaming the config function convention and also the script name.

```
config_salt() {
}
```

instead of:

```
config_minion() {
}
```

and also `bootstrap-salt-minion.sh` -> `bootstrap-salt.sh`

Agreed?

akoumjian:
Yes. Do we do the release before this?

techhat:
@s0undt3ch I didn't even think about the script name. It _will_ be nice to not have to do so much typing. ^_^ I also agree with changing the function name.

akoumjian:
@s0undt3ch : Also consider the case where we want a master but no minion. So `--master` and perhaps `--no-minion`, and leave minion as the default?

akoumjian:
Or perhaps that gets even more complicated with syndic. Perhaps require the flag for each...

s0undt3ch:
Hmm, now that we're expanding the idea, perhaps a flag for each install, but for the minion it should be a negative flag(for backwards compatibility).

s0undt3ch:
I already had the master flag setup locally, but I think I'll address the other ones before commiting

s0undt3ch:
@akoumjian About before of after the "master" branch. How about after? A full blown bootstrap script becomes master.

akoumjian:
Fair enough. I am personally in no immediate rush, and it would be silly to put out such a major change after a big release. I will have some time this evening to test out the master install functionality.

We will also have to remember to send out some kind of downstream notice. Obviously I can update salty-vagrant when necessary, @techhat can handle salt-cloud. Hopefully changing script name doesn't break anyone's automated tools.

s0undt3ch:
@akoumjian about the rename, we can leave a symlink in the repository to the new name?

s0undt3ch:
@akoumjian if you want to test the new master stuff, lemme push some changes I have locally :smile:

s0undt3ch:
Ok. current changes are in. Still need to add the minion and syndic flags. I'll see if I get to that once I get home.

techhat:
Do we really need syndic flags? That's just a part of the master configuration. In salt-cloud, I was fully intending to just provide docs on how to set up syndics.

akoumjian:
I haven't used syndic yet, I just knew that there was a separate package in Ubuntu for it. However, if it is just part of the master config, then perhaps not?

techhat:
It does have a separate daemon that runs, and now that you mention it, I think it does often have its own package. But it's just an extension of the salt-master functionality, so that's where its configuration goes.

Since it is sometimes a separate package that isn't necessarily needed (but will be in my case), I think I would like to have another flag for it, if it's not going to be too much of a pain.

s0undt3ch:
I think separate flags should be used since separate packages are often provided.

And since the discussion is now broader, it continues on #27.

Issue #27, opened 2013-01-25
Support master/syndic installations


s0undt3ch:
Carrying on the conversation from #26.
Support master/syndic installations

s0undt3ch:
The only distro we already support which still lacks master/syndic installs is FreeBSD. It will be taken care of when we have a repository for the salt packages and their dependencies.

s0undt3ch:
The FreeBSD issues mentioned above were addressed in #47 by @geoffgarside. Closing.

PR #31, opened 2013-01-26
Enable CentOS master/syndic installations. Refs #27

s0undt3ch:
This can be merged but it's not yet fully functional since starting the daemon(s), specially the minion hangs waiting for a master authentication.


s0undt3ch:
Check/waiting for saltstack/salt#3424 and saltstack/salt#3183

Maybe the bootstrap script should test for the presence of upstart support.

s0undt3ch:
There's no problem when using upstart which is the default for CentOS >= 6

PR #34, opened 2013-01-31
only use the first number of the distro for centos

visualphoenix:
https://github.com/saltstack/salt-bootstrap/commit/c2d641bdde06d89ddcd15f980542af540aeb786a removed the _63_ functions, which had fixed centos. this pull request makes it so that the epel can be downloaded properly.

### DIFF
--- a/bootstrap-salt-minion.sh
+++ b/bootstrap-salt-minion.sh
@@ -449,9 +449,11 @@ echo
 
 # Simplify version naming on functions
 if [ "x${DISTRO_VERSION}" = "x" ]; then
+    DISTRO_MAJOR_VERSION=""
     DISTRO_VERSION_NO_DOTS=""
     PREFIXED_DISTRO_VERSION_NO_DOTS=""
 else
+    DISTRO_MAJOR_VERSION="$(echo $DISTRO_VERSION | sed 's/^\([0-9]*\).*/\1/g')"
     DISTRO_VERSION_NO_DOTS="$(echo $DISTRO_VERSION | tr -d '.')"
     PREFIXED_DISTRO_VERSION_NO_DOTS="_${DISTRO_VERSION_NO_DOTS}"
 fi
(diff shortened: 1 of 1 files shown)


s0undt3ch:
First of all, Thank You for your interest in our project.

What did not match, or actually failed, in order for you to send this pull request?

I'm failing to see the benefit.

## 

Pedro Algarvio

----- Reply message -----
De: "Raymond Barbiero" notifications@github.com
Para: "saltstack/salt-bootstrap" salt-bootstrap@noreply.github.com
Assunto: [salt-bootstrap] only use the first number of the distro for centos (#34)
Data: qui, Jan 31, 2013 03:05
c2d641b removed the 63 functions, which had fixed centos. this pull request makes it so that the epel can be downloaded properly. 

You can merge this Pull Request by running 
git pull https://github.com/visualphoenix/salt-bootstrap develop 
Or view, comment on, or merge it at: 
https://github.com/saltstack/salt-bootstrap/pull/34 

Commit Summary 

only use the first number of the distro for centos 

File Changes 

M 
bootstrap-salt-minion.sh 
(4) 

Patch Links: 

https://github.com/saltstack/salt-bootstrap/pull/34.patch 
https://github.com/saltstack/salt-bootstrap/pull/34.diff

s0undt3ch:
Ok, I see your problem. So this does not work in CentOS 6.3 right?

visualphoenix:
Corrent. DISTRO_VERSION_NO_DOTS is 63 for CentOS 6.3. My patch makes it so that you're comparing only the first digit. It would probably be more robust to introduce a DISRO_MAJOR_VERSION variable which stripped all characters after and including the dot (ie: echo "981.4" | sed 's/.[0-9]*//g' ). This would handle cases when CentOS eventually reaches version 10.1. This will probably be a long time from now, but is probably more robust fix.

visualphoenix:
If you don't want to merge both these changes, I can open a new pull request

PR #37, opened 2013-01-31
add DISTRO_MAJOR_VERSION and use it for centos epel

visualphoenix:
replaced DISTRO_VERSION_NO_DOTS variable/usage with DISTRO_MAJOR_VERSION variable/usage. Centos 6.3 would have DISTRO_VERSION_NO_DOTS = 63. This is outside the range of the values used in the version comparison.


s0undt3ch:
Thank You and sorry it took so long. This was merged by hand since the script was renamed.

PR #44, opened 2013-02-08
Added support for an additional function to start daemons

### LINKED ISSUE #39: Start Daemons
I think we should have either an additional flag, or a function to start the installed daemons.

Some distributions start the daemons on their own, for example, Ubuntu. Others do not start their daemons, for example, RHEL/CentOS/Amazon Linux.

I'm in favor for having an optional function.

s0undt3ch:
Added support for an additional function to start daemons, in case needed. At least Ubuntu stable and ppa packages don't(for example).

Fixes #39.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-language: python
-
 before_install:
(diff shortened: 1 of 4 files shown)


s0undt3ch:
@saltstack Opinions?

Issue #20, opened 2013-01-17
FreeBSD seems to be broken on EC2


techhat:
Trying to spin up a VM on AWS EC2 using salt-cloud, using the following salt-cloud profile:

aws-freebsd:
    provider: aws
    image: ami-1d4c9874
    size: Micro Instance
    script: bootstrap-salt-minion
    ssh_username: root
    sudo: False

In this configuration, the script is uploaded as /tmp/deploy.sh, and executed with:

/tmp/deploy.sh -c /tmp

When it tries to run, I get:

/tmp/deploy.sh: Command not found.

When I edit the script and change the first line to use sh instead of bash, it executes. But then it fails to load all of the packages:
- System Information:
   CPU:          Intel(R) Xeon(R) CPU           E5645  @ 2.40GHz
   CPU Arch:     amd64
   OS Name:      FreeBSD
   OS Version:   9.0-RELEASE
   Distribution: FreeBSD 9.0
- INFO: Found function install_freebsd_90_stable_deps
- INFO: Found function install_freebsd_90_stable
- INFO: config_freebsd_90_stable_minion not found....
- INFO: config_freebsd_90_minon not found....
- INFO: config_freebsd_stable_minion not found....
- INFO: config_freebsd_minion not found....
- INFO: Found function config_minion
- INFO: Found function install_freebsd_90_stable_post
- Running install_freebsd_90_stable_deps()
  pkg.txz                                               1690 kB 1143 kBps
  Installing pkg-1.0.2...pkg-1.0.2 already installed

Failed to install the following 1 package(s): ./pkg.txz
Conversion has been successfully finished
Your old packages database has been moved to: /var/db/pkg.bak.
Updating repository catalogue
Repository catalogue is up-to-date, no need to fetch fresh copy
pkg: Package 'swig' was not found in the repositories
- Failed to run install_freebsd_90_stable_deps()!!!
- Removing the logging pipe /tmp/bootstrap-salt-minion.logpipe
- Killing logging pipe tee's with pid(s): 1092

akoumjian:
Not sure why it got switched to bash, should be sh at the top.

Don't know anything about freebsd packaging, unfortunately. Looks like @cedwards might know something.

s0undt3ch:
Yes, `sh` should be at the top.

s0undt3ch:
@techhat about the packages not found, look at @akoumjian's reply

ghost:
These should all be /bin/sh, not 'bash'.

[...]

PR #47, opened 2013-02-12
FreeBSD 9.1 support + start daemons

geoffgarside:
The following commits add FreeBSD 9.1 support along side FreeBSD 9.0. They are both pretty much the same, all they do is provide the `_freebsd_90_` and `_freebsd_91_` methods which call a `_freebsd_9x_` function.

I've also added /etc/rc.conf enabling of the daemons as per the [FreeBSD Installation Instructions](http://docs.saltstack.org/en/latest/topics/installation/freebsd.html).

Another pass has been made on the DISTRO_NAME_L setting, FreeBSD doesn't seem to like the `sed -re 's/(\s)+/_/g'` method either though `sed -re 's/[[:space:]]+/_/g'` does seem to be working in both FreeBSD and Ubuntu for me.

I've changed the name of the "salt" FreeBSD package to "sysutils/py-salt" to reflect the new package name. You could also use "py27-salt" but the fully qualified port name has the advantage of supporting any future python version defaults used in the package repository.

As a potential fix for #20 I've added the ability for the pkgng PACKAGESITE variable to be set by the environment, I'm not too sure if the `VAR=${VAR:="default"}` syntax is plain POSIX or not though.

Apologies for this being a rather disjointed set of commits, if you'd like me to break them up a bit into more logical chunks I'll happily do so.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1310,17 +1310,17 @@ __freebsd_get_packagesite() {
         BSD_ARCH="x86:32"
     fi
 
-    PACKAGESITE=${PACKAGESITE:="http://pkgbeta.freebsd.org/freebsd:9:${BSD_ARCH}/latest"}
+    BS_PACKAGESITE=${PACKAGESITE:-"http://pkgbeta.freebsd.org/freebsd:9:${BSD_ARCH}/latest"}
 }
 
 install_freebsd_9x_stable_deps() {
     __freebsd_get_packagesite
 
-    fetch "${PACKAGESITE}/Latest/pkg.txz"
+    fetch "${BS_PACKAGESITE}/Latest/pkg.txz"
     tar xf ./pkg.txz -s ",/.*/,,g" "*/pkg-static"
     ./pkg-static add ./pkg.txz
     /usr/local/sbin/pkg2ng
-    echo "PACKAGESITE: ${PACKAGESITE}" > /usr/local/etc/pkg.conf
+    echo "PACKAGESITE: ${BS_PACKAGESITE}" > /usr/local/etc/pkg.conf
 
     /usr/local/sbin/pkg install -y swig
 }
@@ -1336,11 +1336,11 @@ install_freebsd_91_stable_deps() {
 install_freebsd_git_deps() {
     __freebsd_get_packagesite
 
-    fetch "${PACKAGESITE}/Latest/pkg.txz"
+    fetch "${BS_PACKAGESITE}/Latest/pkg.txz"
     tar xf ./pkg.txz -s ",/.*/,,g" "*/pkg-static"
     ./pkg-static add ./pkg.txz
     /usr/local/sbin/pkg2ng
-    echo "PACKAGESITE: ${PACKAGESITE}" > /usr/local/etc/pkg.conf
+    echo "PACKAGESITE: ${BS_PACKAGESITE}" > /usr/local/etc/pkg.conf
 
     /usr/local/sbin/pkg install -y swig
 
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1299,7 +1299,7 @@ install_arch_start_daemons() {
 #
 #   FreeBSD Install Functions
 #
-install_freebsd_90_stable_deps() {
+install_freebsd_9x_stable_deps() {
     if [ $CPU_ARCH_L = "amd64" ]; then
         BSD_ARCH="x86:64"
     elif [ $CPU_ARCH_L = "x86_64" ]; then
@@ -1317,6 +1317,14 @@ install_freebsd_90_stable_deps() {
     echo "PACKAGESITE: http://pkgbeta.freebsd.org/freebsd:9:${BSD_ARCH}/latest" > /usr/local/etc/pkg.conf
 
     /usr/local/sbin/pkg install -y swig
+}
+
+install_freebsd_90_stable_deps() {
+    install_freebsd_9x_stable_deps
+}
+
+install_freebsd_91_stable_deps() {
+    install_freebsd_9x_stable_deps
 }
 
 install_freebsd_git_deps() {
@@ -1346,8 +1354,16 @@ install_freebsd_git_deps() {
     fi
 }
 
-install_freebsd_90_stable() {
+install_freebsd_9x_stable() {
     /usr/local/sbin/pkg install -y salt
+}
+
+install_freebsd_90_stable() {
+    install_freebsd_9x_stable
+}
+
+install_freebsd_91_stable() {
+    install_freebsd_9x_stable
 }
 
 install_freebsd_git() {
@@ -1378,6 +1394,23 @@ install_freebsd_start_daemons() {
         salt-$fname -d &
     done
 }
+
+install_freebsd_9x_stable_post() {
+    salt-minion -d &
+}
+
+install_freebsd_90_stable_post() {
+    install_freebsd_9x_stable_post
+}
+
+install_freebsd_91_stable_post() {
+    install_freebsd_9x_stable_post
+}
+
+install_freebsd_git_post() {
+    salt-minion -d &
+}
+
 #
 #   Ended FreeBSD Install Functions
 #
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1379,10 +1379,12 @@ install_freebsd_9x_stable_post() {
         [ $fname = "master" ] && [ $INSTALL_MASTER -eq $BS_FALSE ] && continue
         [ $fname = "syndic" ] && [ $INSTALL_SYNDIC -eq $BS_FALSE ] && continue
 
-        # XXX: Should we cp /usr/local/etc/salt/master.sample /usr/local/etc/salt/master and same for minion?
         enable_string="salt_${fname}_enable=\"YES\""
         grep "$enable_string" /etc/rc.conf >/dev/null 2>&1
         [ $? -eq 1 ] && echo "$enable_string" >> /etc/rc.conf
+
+        [ $fname = "minion" ] && cp /usr/local/etc/salt/${fname}.sample /usr/local/etc/salt/${fname}
+        [ $fname = "master" ] && cp /usr/local/etc/salt/${fname}.sample /usr/local/etc/salt/${fname}
 
         if [ $fname = "minion" ] ; then
             grep "salt_minion_paths" /etc/rc.conf >/dev/null 2>&1
@@ -1408,10 +1410,12 @@ install_freebsd_git_post() {
         [ $fname = "master" ] && [ $INSTALL_MASTER -eq $BS_FALSE ] && continue
         [ $fname = "syndic" ] && [ $INSTALL_SYNDIC -eq $BS_FALSE ] && continue
 
-        # XXX: Should we cp /usr/local/etc/salt/master.sample /usr/local/etc/salt/master and same for minion?
         enable_string="salt_${fname}_enable=\"YES\""
         grep "$enable_string" /etc/rc.conf >/dev/null 2>&1
         [ $? -eq 1 ] && echo "$enable_string" >> /etc/rc.conf
+
+        [ $fname = "minion" ] && cp /usr/local/etc/salt/${fname}.sample /usr/local/etc/salt/${fname}
+        [ $fname = "master" ] && cp /usr/local/etc/salt/${fname}.sample /usr/local/etc/salt/${fname}
 
         if [ $fname = "minion" ] ; then
             grep "salt_minion_paths" /etc/rc.conf >/dev/null 2>&1
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -517,7 +517,7 @@ else
     PREFIXED_DISTRO_VERSION_NO_DOTS="_${DISTRO_VERSION_NO_DOTS}"
 fi
 # Simplify distro name naming on functions
-DISTRO_NAME_L=$(echo $DISTRO_NAME | tr '[:upper:]' '[:lower:]' | sed 's/[^a-zA-Z0-9_ ]//g' | sed -re 's/[[:space:]]+/_/g')
+DISTRO_NAME_L=$(echo $DISTRO_NAME | tr '[:upper:]' '[:lower:]' | sed 's/[^a-zA-Z0-9_ ]//g' | sed -re 's/([[:space:]])+/_/g')
 
 
 # Only Ubuntu has daily packages, let's let users know about that
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1373,17 +1373,7 @@ install_freebsd_git() {
     /usr/local/bin/python setup.py install
 }
 
-install_freebsd_90_stable_post__() {
-    # XXX: What needs to be done for init.d support on FreeBSD
-    echo
-}
-
-install_freebsd_git_post__() {
-    # XXX: What needs to be done for init.d support on FreeBSD
-    echo
-}
-
-install_freebsd_start_daemons() {
+install_freebsd_9x_stable_post() {
     for fname in minion master syndic; do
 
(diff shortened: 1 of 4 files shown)


s0undt3ch:
@geoffgarside, first of all, Thanks!

I haven't looked at all your changes, though, the `DISTRO_NAME_L`, could you please verify that it works and change it to `sed -re 's/([[:space:]])+/_/g'`. That way we collapse multiple whitespace into a single underscore:

```
$ echo 'Foo Bar and Foo  Bar   FreeBSD' | sed -re 's/([[:space:]])+/_/g'
Foo_Bar_and_Foo_Bar_FreeBSD
```

s0undt3ch:
About the `POSIX` or not `POSIX` issues, we now have a test-suite. One of the first tests it does is `POSIX` compatibility, have a look [here](https://github.com/saltstack/salt-bootstrap/tree/develop/tests), more specifically, [here](https://github.com/saltstack/salt-bootstrap/blob/develop/tests/bootstrap/test_lint.py)

s0undt3ch:
If `install_freebsd_90_stable_deps` and `install_freebsd_91_stable_deps` do the exact same thing, couldn't we just define `install_freebsd_stable_deps`? Or will that bring problems with FreeBSD < 9.0 ?

s0undt3ch:
Sorry if I've sounded picky so far :smile:

geoffgarside:
Yes the `echo 'Foo Bar and Foo  Bar   FreeBSD' | sed -re 's/([[:space:]])+/_/g'` works, I'll change it to use that. I'll get a VM to run the tests in as it says it'll install/remove packages.

I believe the pkgng system should be supported in FreeBSD 8.2+ and 9.0+, it's possible people might bootstrap older versions but I could add a check to make sure the major version isn't less than 8. I simply opted for the most similar method of adding FreeBSD 9.1 based on what was already in the bootstrap.

Not at all, picky is a very good thing, especially for a multi-platform bootstrap project.

geoffgarside:
I'd need to get the Major version anyway in order to work out the pkgng ABI value as FreeBSD 8 will use "freebsd:8:${BSD_ARCH}" while the code at the moment assumes "freebsd:9:${BSD_ARCH}" so adding the check to make sure its not less than 8 shouldn't be a problem.

s0undt3ch:
#37 already mentions major and minor distro version support. Need to get that included.

s0undt3ch:
We can leave the 9.0 and 9.1 support as is. Maybe if someone complains, we later add 8 support?

geoffgarside:
Yeah can do FreeBSD 8 later if theres a demand for it, hopefully after #37 :)

I've run the Lint tests successfully and the Usage ones on a FreeBSD VM, I'll have to look further into the Installation tests and probably refamiliarise myself with Python :P

geoffgarside:
Yes good point, they do seem to be reliably installed by the port at the moment, but that could change.

s0undt3ch:
I showed you that code because it will only copy if the file exists and, if we're only installing a minion, the master file won't be copied over because it's skipped some lines above :)

s0undt3ch:
heh, I've read your changes only now :smile: 

A single:

```
[ -f /usr/local/etc/salt/${fname}.sample ] && cp /usr/local/etc/salt/${fname}.sample /usr/local/etc/salt/${fname}
```

is enough.

`[ $fname = "minion" ] &&` is not necessary since if we skip on the loop if not installing, see [here](https://github.com/geoffgarside/salt-bootstrap/blob/5545aa6cfdb162c83da2f1feeb8b5cb571bd98ac/bootstrap-salt.sh#L1378)

geoffgarside:
Ah yes, now I've got you, the additional tests for fname value are redundant.

s0undt3ch:
Yes. Also, [here](https://github.com/geoffgarside/salt-bootstrap/blob/5545aa6cfdb162c83da2f1feeb8b5cb571bd98ac/bootstrap-salt.sh#L1389) you only take into account the minion paths. The master and syndic paths are not important in FreeBSD?

s0undt3ch:
You noticed that now your sample copy lines are equal right? ;)

geoffgarside:
Only the `salt_minion` rc.d script has any paths configuration options.

s0undt3ch:
Ok, if those path changes makes it all work, I'm fine with it :)

geoffgarside:
Oh dear, school boy errors, its been an exhausting day.

geoffgarside:
The `salt_minion_paths` are lifted straight from the [FreeBSD Install guide](http://docs.saltstack.org/en/latest/topics/installation/freebsd.html) and include all the main paths. I suspect its required as without a shell most processes wouldn't get the /usr/local/{bin,sbin} paths which is where ports are installed.

s0undt3ch:
Well, I can tell you that I'm feeling bad for throwing all this at you :)

These are normal mistakes, I have just been devoting more time to this script, hence me picking this issues up.

s0undt3ch:
Ah! I think that the paths entry is there in order for the necessary binaries are found by the minion when running.

geoffgarside:
No no, don't feel bad about it, good to know its well curated :)

I've been looking through the tests/bootstrap/ext/os_data.py, it looks like just the os_family, oscodename, osfullname and osrelease grains aren't being detected for FreeBSD. Other than those are there any others I should be looking to populate. I'm not sure what the stubbed `_bsd_cpudata(grains)` ought to be doing.

bootstrap-salt.sh
```diff
@@ -1310,31 +1310,37 @@ install_freebsd_90_stable_deps() {
         BSD_ARCH="x86:32"
     fi
 
-    fetch http://pkgbeta.freebsd.org/freebsd:9:${BSD_ARCH}/latest/Latest/pkg.txz
+    PACKAGESITE=${PACKAGESITE:="http://pkgbeta.freebsd.org/freebsd:9:${BSD_ARCH}/latest"}
```
s0undt3ch: Last change :smile: 

You see any issues with making this line like:

```
PACKAGESITE=${BS_PACKAGESITE:="http://pkgbeta.freebsd.org/freebsd:9:${BSD_ARCH}/latest"}
```

I've prefixed the environment variable with `BS_`.

s0undt3ch:
The code in `tests/bootstrap/ext/os_data.py` was taken [directly from salt ](https://github.com/saltstack/salt/blob/develop/salt/grains/core.py#L548). I probably did not copy enough code for FreeBSD to be properly detected.

bootstrap-salt.sh
geoffgarside: Actually I think it probably needs to be

`BS_PACKAGESITE=${PACKAGESITE:-"http://pkgbeta.freebsd.org/freebsd:9:${BSD_ARCH}/latest"}`

so the `:-` format won't set $PACKAGESITE but if its unset $BS_PACKAGESITE will get the value or that of $PACKAGESITE if its defined. I'd probably try to keep $PACKAGESITE as the environment variable as thats in keeping with the environment variables you can use to override `pkg`.

s0undt3ch: I am using `:-` [here](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L41) and it does work.

Regarding the `PACKAGESITE` env variable, if FreeBSD already uses it, ok, let's not rename it. This is valuable information(because I would have renamed it :smiley:)

geoffgarside: Its probably better form to check for the existence of environment variables and give them internal names for continued use and setting defaults and the like though. I'll keep the environment variable as PACKAGESITE and internally store that (or use the determined default) as BS_PACKAGESITE.

geoffgarside:
I think I need to read through the tests and work out how much of the grain information is actually used before I go and copy the `_bsd_cpudata()` method across :-P

s0undt3ch:
Currently we're using the grains so we know how to un-install any salt packages and their dependencies, [have a look here](https://github.com/geoffgarside/salt-bootstrap/blob/5545aa6cfdb162c83da2f1feeb8b5cb571bd98ac/tests/bootstrap/test_install.py#L17)

s0undt3ch:
So, right now, only OS Family.

s0undt3ch:
WOW! You even added the required code for the tests! A BIG THANK YOU!

geoffgarside:
Nuts, the tests aren't working yet :(

s0undt3ch:
You're running?

```
PACKAGESITE=foo sudo -E tests/runtests.py
```

s0undt3ch:
Or you might be running as root :)

geoffgarside:
Yeah, running as root on a VM and I've got my PACKAGESITE set. It might be failing on the cleanup

```
ERunning cleanup command 'pkg delete -y sysutils/py-salt'
======================================================================
ERROR: test_install_daily (bootstrap.test_install.InstallationTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/salt-bootstrap/tests/bootstrap/test_install.py", line 139, in test_install_daily
    args=('daily',), timeout=15 * 60, stream_stds=True
  File "/root/salt-bootstrap/tests/bootstrap/__init__.py", line 138, in run_script
    while process.poll() is None:
  File "/root/salt-bootstrap/tests/bootstrap/__init__.py", line 74, in poll
    obuff = self.stdout.read()
IOError: [Errno 35] Resource temporarily unavailable
```

s0undt3ch:
I'm only taking into account errno 11 in the tests, that might be it.

Looking [here](http://fxr.watson.org/fxr/source/sys/errno.h#L92) the fix should be, in [line 78'](https://github.com/saltstack/salt-bootstrap/blob/develop/tests/bootstrap/__init__.py#L78) and [line 88](https://github.com/saltstack/salt-bootstrap/blob/develop/tests/bootstrap/__init__.py#L88) of [tests/bootstrap/**init**.py](https://github.com/saltstack/salt-bootstrap/blob/develop/tests/bootstrap/__init__.py):

``` python
            except IOError, err:
                if err.errno not in (11, 35):
```

geoffgarside:
Thanks, thats worked, two errors now, one in test_install_daily and test_install_using_bash, I'm guessing some part of the cleanup is failing. I'm re-running the tests now and capturing the output to see which cleanup command it running and then getting an "E".

```
======================================================================
ERROR: test_install_daily (bootstrap.test_install.InstallationTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/salt-bootstrap/tests/bootstrap/test_install.py", line 55, in tearDown
    stream_stds=True
  File "/root/salt-bootstrap/tests/bootstrap/__init__.py", line 191, in assert_script_result
    err_msg = '{0} No std{out,err} captured.'.format(err_msg)
KeyError: 'out,err'

======================================================================
ERROR: test_install_using_bash (bootstrap.test_install.InstallationTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/salt-bootstrap/tests/bootstrap/test_install.py", line 55, in tearDown
    stream_stds=True
  File "/root/salt-bootstrap/tests/bootstrap/__init__.py", line 191, in assert_script_result
    err_msg = '{0} No std{out,err} captured.'.format(err_msg)
KeyError: 'out,err'

----------------------------------------------------------------------
```

s0undt3ch:
Ops, that's also my fault.

Change:

``` python
err_msg = '{0} No std{out,err} captured.'.format(err_msg)
```

to

``` python
err_msg = '{0} No stdout nor stderr captured.'.format(err_msg)
```

s0undt3ch:
But apparently an error did occur, we just did not get any output

geoffgarside:
Ok, applied the change, running tests again. So is Python wanting to interpolate into the `{out,err}` bit then? My python skills are rather rusty.

geoffgarside:
Ok, got an error message :)

```
AssertionError: Failed to execute cleanup command 'pkg delete -y swig sysutils/py-salt':
 No stdout nor stderr captured.
```

though I've no idea why that wouldn't return any stdout or stderr, running them manually you get output whether they're installed or not.

s0undt3ch:
About the interpolation, yes, `{0}` are list indexes, `{foo}` is dictionary keys using `.format()` on strings.

Similiar to

``` python
'%s %(foo)s' % 'a' % dict(foo=b)
```

s0undt3ch:
What's the exit code when you run:

```
'pkg delete -y swig sysutils/py-salt'
```

ie:

```
echo $?
```

geoffgarside:
When the packages are installed and you run that, you get 0.
When the packages are missing and you run that, you get 65.

I think the 65 corresponds to the pkgng EPKG_\* codes.

Thanks for the info on the interpolation :)

s0undt3ch:
Ok, got another fix for you:

``` diff
diff --git a/tests/bootstrap/__init__.py b/tests/bootstrap/__init__.py
index ff26c15..0e02d7c 100644
--- a/tests/bootstrap/__init__.py
+++ b/tests/bootstrap/__init__.py
@@ -179,9 +179,12 @@ class BootstrapTestCase(TestCase):
                 # process already terminated
                 pass

-    def assert_script_result(self, fail_msg, expected_rc, process_details):
+    def assert_script_result(self, fail_msg, expected_rcs, process_details):
+        if not isinstance(expected_rcs, (tuple, list)):
+            expected_rcs = (expected_rcs,)
+
         rc, out, err = process_details
-        if rc != expected_rc:
+        if rc not in expected_rcs:
             err_msg = '{0}:\n'.format(fail_msg)
             if out:
                 err_msg = '{0}STDOUT:\n{1}\n'.format(err_msg, '\n'.join(out))
diff --git a/tests/bootstrap/test_install.py b/tests/bootstrap/test_install.py
index 096f506..6631488 100644
--- a/tests/bootstrap/test_install.py
+++ b/tests/bootstrap/test_install.py
@@ -43,7 +43,11 @@ class InstallationTestCase(BootstrapTestCase):
             print 'Running cleanup command {0!r}'.format(cleanup)
             self.assert_script_result(
                 'Failed to execute cleanup command {0!r}'.format(cleanup),
-                0,
+                (
+                    0,
+                    65  # FreeBSD throws this error code when the packages being
+                        # un-installed were not installed in the first place
+                ),
                 self.run_script(
                     script=None,
                     args=cleanup.split(),
```

geoffgarside:
Yay

```
Ran 8 tests in 176.167s

OK (skipped=1)
```

I suspect the bash one might have been skipped as /bin/bash doesn't exist on FreeBSD, ends up in /usr/local/bin/bash once the package is installed.

s0undt3ch:
That is correct!

Can I add your name and email to the script's header?

s0undt3ch:
You can also add it if you like :)

geoffgarside:
Shall I apply your fixes in my branch or do you want to apply them to master and I'll merge master into my branch? Yeah I can add that to the header :)

s0undt3ch:
Nah! You can add all fixes!

geoffgarside:
Well, after 3 hours of bug squashing and very kind assistance I think it might just be ready :)

s0undt3ch:
Well, I'm just finishing some more tweaks to the now included #37, which will allow future FreeBSD 8 support, if anyone asks for it, and I'll merge your changes.

Once again. Thank You!

geoffgarside:
Thats cool, by the looks of things, the simplest way to add FreeBSD 8 support would be to put `$DISTRO_MAJOR_VERSION` in the `BS_PACKAGESITE` variable value and then berate anyone who tried to do it on any earlier FreeBSD versions :P

s0undt3ch:
Yep!

s0undt3ch:
Ok, could you please pull latest changes and see if it still works?

geoffgarside:
Ok, will let you know shortly.

s0undt3ch:
There was an additional fix to `DISTRO_MINOR_VERSION`.

geoffgarside:
Yeah noticed some sed errors, fetched and running testsuite again :)

geoffgarside:
Yep, all tests passing, 1 skipped

s0undt3ch:
Awesome!!!!

Once again, Thanks!!!

## 

Pedro Algarvio

----- Reply message -----

PR #54, opened 2013-02-14
Develop

ixela:
Fixed bashisms to comply with unit testing.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -366,17 +366,13 @@ __gather_linux_system_info() {
     DISTRO_VERSION=""
 
     if [ -f /etc/lsb-release ]; then
-    #if [ $(lsb_release -a | grep Descr | awk '{ print $2 }') = "SUSE" ]; then
-     #   DISTRO_NAME="suse"
-     #   DISTRO_VERSION=$(lsb_release -a | grep Rel | awk '{ print $2}')
-#	else 
         DISTRO_NAME=$(grep DISTRIB_ID /etc/lsb-release | sed -e 's/.*=//')
         DISTRO_VERSION=$(__parse_version_string $(grep DISTRIB_RELEASE /etc/lsb-release | sed -e 's/.*=//'))
         fi
- #   fi
     if [ -f /etc/SuSE-release ]; then
           DISTRO_NAME="suse"
 	  DISRTRO_VERSION=$(grep VERSION /etc/SuSE-release | awk '{ print $3}'
+    fi
    if [ "x$DISTRO_NAME" != "x" ] && [ "x$DISTRO_VERSION" != "x" ]; then
         # We already have the distribution name and version
         return
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -366,15 +366,18 @@ __gather_linux_system_info() {
     DISTRO_VERSION=""
 
     if [ -f /etc/lsb-release ]; then
-    if [ $(lsb_release -a | grep Descr | awk '{ print $2 }') = "SUSE" ]; then
-        DISTRO_NAME="suse"
-        DISTRO_VERSION="$(lsb_release -a | grep Rel | awk '{ print $2}')"
-	else 
+    #if [ $(lsb_release -a | grep Descr | awk '{ print $2 }') = "SUSE" ]; then
+     #   DISTRO_NAME="suse"
+     #   DISTRO_VERSION=$(lsb_release -a | grep Rel | awk '{ print $2}')
+#	else 
         DISTRO_NAME=$(grep DISTRIB_ID /etc/lsb-release | sed -e 's/.*=//')
         DISTRO_VERSION=$(__parse_version_string $(grep DISTRIB_RELEASE /etc/lsb-release | sed -e 's/.*=//'))
         fi
-    fi
-    if [ "x$DISTRO_NAME" != "x" ] && [ "x$DISTRO_VERSION" != "x" ]; then
+ #   fi
+    if [ -f /etc/SuSE-release ]; then
+          DISTRO_NAME="suse"
+	  DISRTRO_VERSION=$(grep VERSION /etc/SuSE-release | awk '{ print $3}'
+   if [ "x$DISTRO_NAME" != "x" ] && [ "x$DISTRO_VERSION" != "x" ]; then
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -368,9 +368,12 @@ __gather_linux_system_info() {
     if [ -f /etc/lsb-release ]; then
         DISTRO_NAME=$(grep DISTRIB_ID /etc/lsb-release | sed -e 's/.*=//')
         DISTRO_VERSION=$(__parse_version_string $(grep DISTRIB_RELEASE /etc/lsb-release | sed -e 's/.*=//'))
+        fi
+    if [ -f /etc/SuSE-release ]; then
```
s0undt3ch, line 372: This check is unnecessary. Suse should be "catch" in [this for loop](https://github.com/ixela/salt-bootstrap/blob/fc9fd761e82e0c83c43b05e69ce2d96888dc017d/bootstrap-salt.sh#L381). If it doesn't it's a bug.

bootstrap-salt.sh
```diff
@@ -368,9 +368,12 @@ __gather_linux_system_info() {
...
+          DISTRO_NAME="suse"
+	  DISRTRO_VERSION=$(grep VERSION /etc/SuSE-release | awk '{ print $3}'
     fi
-
-    if [ "x$DISTRO_NAME" != "x" ] && [ "x$DISTRO_VERSION" != "x" ]; then
+   if [ "x$DISTRO_NAME" != "x" ] && [ "x$DISTRO_VERSION" != "x" ]; then
```
s0undt3ch, line 376: This little space removal is also unnecessary :)

bootstrap-salt.sh
```diff
@@ -1503,7 +1506,81 @@ install_smartos_start_daemons() {
...
+if [ ! -f /usr/local/bin/salt ]; then
+install_suse_11_git_deps
+__git_clone_and_checkout
+python setup.py install
+fi
+}
+install_suse_11_post(){
+if [ ! -f /etc/init.d/salt-master ]; then
+if [ $INSTALL_MASTER = 1 ]; then
+curl https://raw.github.com/ixela/salt/develop/pkg/rpm/salt-master > /etc/init.d/salt-master
```
s0undt3ch, line 1558: As you might understand, we can't depend on your own repository.

If there are necessary changes regarding the init scripts for SuSE, those should be filed against salt.

s0undt3ch:
Sorry if I seem picky by asking you to fix these issues.
If you prefer I can fix what I think is wrong.

bootstrap-salt.sh
ixela, line 1558: The changes have been rejected previously, so I'm stuck with either maintaining my own init files or building a spec file that changes file paths. SLES doesn't install salt to the same directory as RHEL or Fedora. Its a simple two line change in the file. I've been asked to provide an alternate spec file so that alternate init files can be generated. I'm currently working on that. I will rework it and have it parse the file instead and do the changes inside the bootstrap script if that is more appropriate.

bootstrap-salt.sh
ixela, line 372: It never catches SLES inside that for loop. Its why I wrote that simple if test, as lsb-release is not the same on SLES. This is the entire /etc/lsb-release for SLES 11:
LSB_VERSION="core-2.0-noarch:core-3.2-noarch:core-4.0-noarch:core-2.0-x86_64:core-3.2-x86_64:core-4.0-x86_64"

As you can see it does not contain the fields needed to parse out the version and distribution name. Instead, the /etc/SuSE-release file contains this data. Here is an example from one of my production machines:

SUSE Linux Enterprise Server 11 (x86_64)
VERSION = 11
PATCHLEVEL = 1

ixela:
I understand, I should have spent some more time looking at my diffs for the bashisms and the spacing. The other things are necessary due to SLES not acting like other RPM based distros.

bootstrap-salt.sh
s0undt3ch, line 372: @ixela I executed the script in a SUSE Linux Enterprise Server 11 SP2 and it was properly detected.

```
 *  INFO: PIPED THROUGH sh -- Version 1.4

 *  INFO: System Information:
 *  INFO:   CPU Arch:     i686
 *  INFO:   OS Version:   3.0.13-0.27-ec2
 *  INFO:   Distribution: SUSE 11
```

Contents of `/etc/SuSE-release`:

```
ip-10-158-69-6:~ # cat /etc/SuSE-release 
SUSE Linux Enterprise Server 11 (i586)
VERSION = 11
PATCHLEVEL = 2
```

bootstrap-salt.sh
s0undt3ch, line 1558: Perhaps some inline `sed` could be used on the original init scripts?

ixela, line 1558: I've got a working model going, however I also just bought a house and have been busy. I'll try to send a new pull request Monday or Tuesday.

bootstrap-salt.sh
ixela, line 372: Is this with my loop change or not? If you're using the code I submitted, then it should catch SLES 11 just fine. If you remove the extra if statement, then it should not as /etc/lsb-release does not reflect the necessary data for the rest of the logic below.

bootstrap-salt.sh
s0undt3ch, line 1558: @ixela Sure, no need to feel pressured ;)

bootstrap-salt.sh
s0undt3ch, line 372: The problem you were encountering, which led you to create that `if` is now also fixed.

The output from above is from a SLES 11 which did not had an `lsb-release` file. After I added one like yours, I then had to fix some parsings.

You don't need that `if` now.

ixela, line 372: I have attempted to verify that the current bootstrap-salt.sh script correctly identifies a SLES 11 environment. It does not. Here is the output:

```
*  INFO: /bin/sh ./bootstrap-salt.sh -- Version 1.4

 *  INFO: System Information:
 *  INFO:   CPU:          AuthenticAMD
 *  INFO:   OS Version:   2.6.32.59-0.7-default
 *  INFO:   Distribution: ASC 22

 *  INFO: Installing minion
```

PR #59, opened 2013-02-22
Install a `pyzmq` package compiled against `libzmq3`. Fixes #57.

### LINKED ISSUE #57: libzmq3 packages for debian 6
Debian 6 (and 7) comes with libzmq1, not libzmq3. This version doesn't support keepalive and is thus pretty worthless (we loose half our minions in a day). Ubuntu doesn't have this problem since it comes with libzmq3.

I have created 2 debian 6 packages:
libzmq3_3.2.2_amd64.deb
python-zmq_2.2.0.1_amd64.deb

After installing those, I run the bootstrap and the minion is successfully installed using libzmq3. 
Who should I send those (or a howto for making them) ?

s0undt3ch:
Although Debian experimental has `libzmq3`, it does not have `pyzmq` compiled against it. We install `pyzmq` from `pip` having `libzmq3` already installed.

**Should we really do this?**
**Even the stable builds are installing `pyzmq` from pip.**

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -922,6 +922,9 @@ install_debian_6_0_deps() {
     echowarn "PyZMQ will be installed from PyPi in order to compile it against ZMQ3"
     echowarn "This is required for long term stable minion connections to the master."
 
+    # No user interaction, libc6 restart services for example
+    export DEBIAN_FRONTEND=noninteractive
+
     if [ "x$(grep -R 'backports.debian.org' /etc/apt)" = "x" ]; then
         echo "deb http://backports.debian.org/debian-backports squeeze-backports main" >> \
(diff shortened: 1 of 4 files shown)


s0undt3ch:
@saltstack opinions?

techhat:
Not being a Debian user, and therefore probably not qualified to say anything, I am okay with this.

ghost:
Wow.. so you're melding the available libzmq3 from experimental with a pip-installed pyzmq that can support zmq3? Good workaround!

s0undt3ch:
Well, yes, it seems that a pyzmq compiled against ZMQ3 is going to take a long time...

I'm also having the users pass an extra flag just for that so they know what their doin'...

s0undt3ch:
Let's solve one of Debian's support current issues.

PR #62, opened 2013-02-26
Removed '&' symbol from the start function for CentOS. Fixes Issue 61.

tateeskew:
I took a look at the script and modified the script for the restart daemons function. Simply removing the & to force to the background lets the script work as expected.


thatch45:
Nice catch, thanks!

techhat:
Ah, that would do it.

s0undt3ch:
And are you sure the script will not hang waiting for salt to start?

tateeskew:
s0undt3ch;  yeah, the script will not hang. should be good to go.

PR #63, opened 2013-02-26
Fix FreeBSD function detection

geoffgarside:
The addition of the `__strip_duplicates` function seems to have broken installation on FreeBSD as the `sed` call ends up replacing all of the spaces with 'n' which is obviously not the intended result.

I've replaced the use of `sed` with `tr` which can match the '[:space:]' character class and, seemingly more portably, replace it with a newline before piping the resulting output through `awk`.

I've run the tests and they pass on FreeBSD.


s0undt3ch:
Thanks @geoffgarside !

PR #68, opened 2013-03-01
Added fetch & reset to refresh existing repo.

rca:
When bootstrap is run more than once on a machine, the existing git
checkout is now refreshed instead of just used as is.


s0undt3ch:
Thank You @rca!

PR #72, opened 2013-03-06
Added INSTALL_DEBIAN_INIT_SCRIPT test.

rca:
This checks for the INSTALL_DEBIAN_INIT_SCRIPT environment variable and
will skip trying to install upstart.  This is a method of installing the
salt-minion init script on an Ubuntu LXC because upstart wasn't working.
The method for bootstrapping salt in the LXC is like so:

```
lxc_bootstrap_salt:
  cmd.run:
    - name: INSTALL_DEBIAN_INIT_SCRIPT=1 lxc-unshare -s MOUNT -- chroot "{{ container['rootfs'] }}" sh {{ bootstrap_salt }} git develop
    - unless: test -e {{ container['rootfs'] }}/usr/bin/salt
    - require:
      - file: {{ container['rootfs'] }}{{ bootstrap_salt }}
```

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -869,7 +869,7 @@ install_ubuntu_git_post() {
         [ $fname = "master" ] && [ $INSTALL_MASTER -eq $BS_FALSE ] && continue
         [ $fname = "syndic" ] && [ $INSTALL_SYNDIC -eq $BS_FALSE ] && continue
 
-        if [ -f /sbin/initctl ]; then
+        if [ -z ${INSTALL_DEBIAN_INIT_SCRIPT} ] && [ -f /sbin/initctl ]; then
             # We have upstart support
             /sbin/initctl status salt-$fname > /dev/null 2>&1
             if [ $? -eq 1 ]; then


s0undt3ch:
At first glance, I think i have to reject this pull request.

It's tailored for your specific needs, and the bootstrap script **should** setup the init scripts and have the daemons start a boot time.

If we detect upstart support, that's what **will** be used since that's what the distribution chose.

So, is the script not doing something it should in order for the installation to properly work?
Was your upstart issue not related to `salt-bootstrap` at all?

Also, the test suite is failing.

rca:
There is likely a better way to do this, so no worries on rejecting this patch.  Here's the scenario.

I'm trying to get salt working within a LXC (linux container).  It appears that upstart does not work within the container environment and starting up services like SSH fallback to using scripts in `/etc/init.d`.  This is speculation though, I'm still learning LXC.

If you have a chance to try, it'd be great to know if the standard bootstrap script properly installs the salt minion for you in an LXC.  Making a container is pretty straight forward on Ubuntu 12.04:

```
sudo apt-get update
sudo apt-get install -y lxc
sudo lxc-create -t ubuntu -n salttest
sudo lxc-start -n salttest -d
```

At that point you have a container running.  You can console in by running `sudo lxc-console -n salttest` (get out of the console with `ctrl+a q`.  Alternatively, dig out the IP and SSH in: `ssh ubuntu@$(host salttest 10.0.3.1)`.

Thanks for telling me about the failing test, I'll look into that.

s0undt3ch:
I think that in order for the bootstrap to properly install salt, you'd have to either uninstall upstart(should not be possible), or, rename `/sbin/initctl` so the script does not detect it.

All this, before running the script of course :smile:

rca:
The idea of hacking the system in order to make the bootstrap script "do the right thing" sounds kinda hokey.  Are you open to the idea of having a configurable switch; maybe a legitimate flag, e.g. `--debian`, or maybe something along the lines of my initial patch if it didn't break the build?

s0undt3ch:
I say the same thing, making the script do a specific thing because a system is crippled(because upstart has issues running within an lxc container) sounds kinda hokey.

I think this fall's out of the scope of the script.

PR #77, opened 2013-03-08
added python2-yaml as dependency in archLinux

maru1034sk:
I've added python2-yaml as dependency for archLinux.

salt-minion doesn't start without it.

```
salt-minion[10163]: ImportError: No module named yaml
```


s0undt3ch:
@plueschopath Thanks!

PR #82, opened 2013-03-10
Update test docs

rca:
Mention the need for `sudo` in order for the tests to run locally.


s0undt3ch:
@rca Thanks for the update!

PR #84, opened 2013-03-11
Disable auto exit for service status

rca:
With `set -e`, testing for the init script by calling `status` will cause the bootstrap script to exit.  Similarly, so will trying to stop a stopped service, which has happened to me when I restore a VM snapshot, but do not clear out the key on the salt master.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -890,8 +890,11 @@ install_ubuntu_git_post() {
 
         if [ -f /sbin/initctl ]; then
             # We have upstart support
+            set +e
             /sbin/initctl status salt-$fname > /dev/null 2>&1
-            if [ $? -eq 1 ]; then
+            status=$?
+            set -e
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -890,8 +890,11 @@ install_ubuntu_git_post() {
 
         if [ -f /sbin/initctl ]; then
             # We have upstart support
+            set +e
             /sbin/initctl status salt-$fname > /dev/null 2>&1
-            if [ $? -eq 1 ]; then
+            status=$?
+            set -e
+            if [ $status -eq 1 ]; then
```
rca, line 897: I also realized that if the init script is ever updated, re-bootstrapping a system will not install the new version.

s0undt3ch:
Sorry for your trouble, but the bootstrap script no longer uses `set -e`, there were too many places where we needed to check the exit codes.

PR #89, opened 2013-03-15
Merge current develop branch into Stable

s0undt3ch:
Distro Support Added:
- SuSE 11 SP1/11 SP2
- OpenSUSE 12.x

Distro Support Fixed:
- Debian 6
- Ubuntu 12.10
- CentOS

Additional improvements/fixes:
- Improved unittesting.
- Starting daemons.
- Make sure that daemons are really running.
- For the users to make the choice if installing from PIP(if required since there aren't system pacakges).
- Fixed salt's git cloning when the salt git tree is already present on the system.


s0undt3ch:
Hmm, this merge needs to be done by hand :frowning: 

Any commented before a "hand" merge?

s0undt3ch:
I'll create a branch based on stable, do the merge there, fix the conflicts and I'll send a pull req

PR #99, opened 2013-03-21
fixed a few pipelines where awk could do the job of several tools

gcw:
AWK can do all of that stuff more efficiently than calling grep, cut, cat, etc...


s0undt3ch:
Thanks!

PR #102, opened 2013-03-31
OS X 10.8 (Darwin) Support

jasonrm:
Adding support is a little tricky as a fresh install of OS X will not have the "Command Line Tools for Xcode" installed by default. I dislike adding an option switch, as it seems, unclean. However, not being able to automate the install of salt on OS X is even more annoying. At least to me. If there is a better way of passing this information I'd be fine with changing it.

I'm not 100% certain that the script is still pure sh despite `sh -n salt-bootstrap.sh` returns no errors due to sh being bash in sh compatibility mode.

When testing it might be best to installing with `./salt-bootstrap.sh git v0.13.3` as there appears to be a few issues in v0.14.0 and develop at the time of this commit. I think the install process works fine, and the keys are register with the salt-master however I am not able to do a test.ping successfully.

Finally, I'm also made a pull request to include the launched plist files for the services so until/if that pull request is accepted the services won't install/run.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -27,6 +27,7 @@ ScriptName="bootstrap-salt.sh"
 #   * BS_ECHO_DEBUG:      If 1 enable debug echo which can also be set by -D
 #   * BS_SALT_ETC_DIR:    Defaults to /etc/salt
 #   * BS_FORCE_OVERWRITE: Force overriding copied files(config, init.d, etc)
+#   * BS_XCODE_CLT_DMG:   URL to "Command Line Tools for Xcode" installer dmg
 #===============================================================================
 
 
@@ -162,6 +163,7 @@ CONFIG_ONLY=$BS_FALSE
 PIP_ALLOWED=${BS_PIP_ALLOWED:-$BS_FALSE}
 SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/etc/salt}
 FORCE_OVERWRITE=${BS_FORCE_OVERWRITE:-$BS_FALSE}
+XCODE_CLT_DMG=${BS_XCODE_CLT_DMG:-null}
 
 while getopts ":hvnDc:MSNCP" opt
 do
@@ -635,6 +637,15 @@ __gather_bsd_system_info() {
     DISTRO_VERSION=$(echo "${OS_VERSION}" | sed -e 's;[()];;' -e 's/-.*$//')
 }
 
+#---  FUNCTION  ----------------------------------------------------------------
+#          NAME:  __gather_darwin_system_info
+#   DESCRIPTION:  Discover Darwin system information
+#-------------------------------------------------------------------------------
+__gather_darwin_system_info() {
+    DISTRO_NAME=${OS_NAME}
+    DISTRO_VERSION=$(sw_vers -productVersion)
+}
+
 
 #---  FUNCTION  ----------------------------------------------------------------
 #          NAME:  __gather_system_info
@@ -658,6 +669,9 @@ __gather_system_info() {
             ;;
         openbsd|freebsd|netbsd )
             __gather_bsd_system_info
+            ;;
+        darwin )
+            __gather_darwin_system_info
             ;;
         * )
             echoerror "${OS_NAME} not supported.";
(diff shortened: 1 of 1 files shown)


s0undt3ch:
W00t!!!! Awesome!

s0undt3ch:
I'm sorry it has taken me so long to reply to this awesome pull request.

I have some issues I'd like to go through with you to see if we get the code up to par to what's required/expected of the bootstrap script.
- I don't favor adding a command line flag just for a specific system. To overcome this issue, could we use an environment variable instead? [Here's](https://github.com/saltstack/salt-bootstrap/blob/v1.5.3/bootstrap-salt.sh#L1853) a usage example. We could call it `BS_XCODE_CLT_DMG` on the script and try to get it as `XCODE_CLT_DMG` from the environment?
- Since you used `easy_install`(could we switch it to pip?), you also need to add a check, [see here](https://github.com/saltstack/salt-bootstrap/blob/v1.5.3/bootstrap-salt.sh#L1952). This is for the users to explicitly allow for pip based installations.
- When using `pip`, you can install requirements like `pip -r requirements.txt`

For now, that's about it, which is a lot.

Since I don't have access to OSX, could you please update your pull request with my suggestions?

Thanks!

jasonrm:
Not a problem, I've been crazy busy with many things as well.
1. I agree that a command line flag is not best. At one point I did have it passed as an environment variable, however I seem to recall running into an issue. I'll make the changes for it to work that way again and take note of what, if any, issues I encounter.
2. I can certainly add the check. As for switching, what should happen when the `-P` option is passed, but `pip` is not installed? Should `pip` be installed via `easy_install` first so long as `pip` is allowed?
3. `pip` would indeed make installing requirements much cleaner

Thanks for looking it over and the feedback.

s0undt3ch:
1. Let me know if I can help out
2. If pip is allowed, it should be installed as a dependency
3. Yes it does ;)

To install distribute and pip, from their docs:

``` sh
$ curl -O http://python-distribute.org/distribute_setup.py
$ python distribute_setup.py
$ curl -O https://raw.github.com/pypa/pip/master/contrib/get-pip.py
$ python get-pip.py
```

seems to be a simple way.

jasonrm:
OK, I've made those changes… though I'm still not quite happy with the install process as it seems there are a number of dependencies that can't be avoided. The one thing I liked about `easy_install` was that it would install binary packages.

It bugs me that both the standard pip install of salt, and the git install have to install the same amount of dependencies, so I'm contemplating the removal of non-git install until, if ever, there is a native salt installer for OS X.

bootstrap-salt.sh
```diff
@@ -1828,6 +1850,210 @@ install_arch_linux_restart_daemons() {
 
 ##############################################################################
 #
+#   OS X 10.8.x Install Functions
+#
+install_darwin_10_8_deps() {
+    [ $PIP_ALLOWED -eq $BS_FALSE ] && pip_not_allowed
+    echowarn "salt dependencies will be installed using pip"
+
+    command -v clang >/dev/null 2>&1
```
s0undt3ch, line 1859: `command -v` is considered a bashism, it should not be used. `command -p` is the only one considered not to be a bashsism.

s0undt3ch:
Hmm, now that you brought to my attention that `easy_install` vs `pip` binary installations "issue", I think there's no real reason why I should say you needed to use `pip`(we just need to say that the `-P` flag is also for `easy_install` installations).

You can use `easy_install`, though, please also require the `-P` flag and it's check because, although binary, you're still not installing a distro package.

About having `pip/easy_install` or `git` installations, I think both should be available. `pip/easy_install` and `git` are able to install stable releases, but sometimes one also wants a specific revision or the latest develop branch to be installed and that can only be achieved with `git`.

Did I make any sense?

bootstrap-salt.sh
jasonrm, line 1859: Ok, I'll try and get those checks modified, although `command -p` doesn't appear to work with the return code checking (or at least it won't work the same way) so I'll have to figure it out. When I looked for other existence checks in the file and found the one in `__function_defined` and I incorrectly assumed that it wasn't a bashism.

s0undt3ch, line 1859: Within your cloned fork, issue:

``` bash
python tests/runtests.py -L
```

That'll check for bashisms...

jasonrm, line 1859: Awesome. Sorry I missed that previously. I had searched for ways to do that kind of check in general and all I had found was `sh -n`.

dplarson:
Any news on this?

s0undt3ch:
@dplarson not much, no.
I don't have a system to test the script, a lot has changed and @jasonrm said right up front that time was not that much.

We can only wait for @jasonrm to find some time to update the script, or any other user which has the proper system to test this pull request.

dplarson:
@s0undt3ch Thanks for the quick reply.

I have access to machines with OS X 10.8.4, 10.7.5 and (I think) 10.6.8. They are all currently used systems (i.e. not fresh/clean installs), but they should work for the purposes of testing. I'll test this script and if needed, revise the scripts (via a separate pull request).

s0undt3ch:
Awesome!

dplarson:
The script ran on OSX 10.8.4 (Mountain Lion) with sudo, but threw some errors when it came to the `/Library/LaunchDaemons/com.saltstack.salt.*.plist` files (related to line 2032, 2034, 2046, 2047). I don't have access to a second machine with OSX 10.8 and so I can only try running salt masterless. I'm looking into whether I can run a virtual machine of OSX so I can test this on a clean install.

On a related note, I think it would be best to keep a separate OSX bootstrap script, both to reduce the complexity of the script and to make it easier to deal with OSX's sometimes non-standard configurations of common UNIX tools.

s0undt3ch:
You know you can run both a master and a minion on the same machine right :smile: 
About the separate script, that defeats this project goal, one script bootstraps _all_ systems :)

dplarson:
Thanks for the clarification. Any suggestions on tutorials or documentation for running a master and minion on the same machine?

s0undt3ch:
Don't know if there are any particularities while running on OSX. But it's just a matter of having salt installed and running both :)

Lemme bring a OSX user along, @basepi, can you help out?

basepi:
Heh, I've actually had a tab with this pull req open in it for weeks, just haven't gotten around to looking at it.

The biggest problem is that though I use OSX, I generally develop for linux and so am not up to date on all its idiosyncrasies.  I'm happy to try to help, though.  =)

s0undt3ch:
@basepi yeah, @dplarson seems to have the stamina to take this on, but I can't help out much because I've never had an OSX(and most likely never will but that does not matter :smile:)

dplarson:
The installation script _works_ (as far as I can tell), but I'm unable to run salt-master (it gets stuck on "Worker binding to socket ipc:///var/run/salt/master/workers.ipc").

basepi:
I never got a chance to really look over this, but now that salt is in both Homebrew and Macports, we should probably pull those install methods into the bootstrap script.

s0undt3ch:
@basepi True! :smile:

s0undt3ch:
While the script does not yet support OS X bootstraps, there are now options to do this. Closing.

PR #109, opened 2013-04-14
lsb_release distronames for openSUSE 12.3 didn't work

twerkmeister:
shorten distroname of openSUSE from 'opensuse_project' to 'opensuse' to match install function names

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -723,6 +723,11 @@ if ([ "${DISTRO_NAME_L}" != "ubuntu" ] && [ $ITYPE = "daily" ]) && \
     echoerror "${DISTRO_NAME} does not have daily packages support"
     exit 1
 fi
+ 
+#lsb_release -si returns "openSUSE project" on openSUSE 12.3
+if [ "${DISTRO_NAME_L}" == "opensuse_project" ]; then
+    DISTRO_NAME_L="opensuse"
+fi
 
 
 #---  FUNCTION  ----------------------------------------------------------------
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -725,7 +725,7 @@ if ([ "${DISTRO_NAME_L}" != "ubuntu" ] && [ $ITYPE = "daily" ]) && \
 fi
  
 #lsb_release -si returns "openSUSE project" on openSUSE 12.3
-if [ "${DISTRO_NAME_L}" == "opensuse_project" ]; then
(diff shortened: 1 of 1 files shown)


s0undt3ch:
I can have a better reply for you in afew hours, though, double equal signs is a bashism. Please use a single one.

## 

Pedro Algarvio @ phone

----- Reply message -----
De: "Thomas" notifications@github.com
Para: "saltstack/salt-bootstrap" salt-bootstrap@noreply.github.com
Assunto: [salt-bootstrap] lsb_release distronames for openSUSE 12.3 didn&apos;t work (#109)
Data: dom, Abr 14, 2013 20:41
shorten distroname of openSUSE from 'opensuse_project' to 'opensuse' to match install function names 

You can merge this Pull Request by running 
git pull https://github.com/lesnail/salt-bootstrap develop 
Or view, comment on, or merge it at: 
https://github.com/saltstack/salt-bootstrap/pull/109 

Commit Summary 

shorten distroname of openSUSE from 'opensuse_project' to 'opensuse' to match install function names 

File Changes 

M 
bootstrap-salt.sh 
(5) 

Patch Links: 

https://github.com/saltstack/salt-bootstrap/pull/109.patch 
https://github.com/saltstack/salt-bootstrap/pull/109.diff

s0undt3ch:
Please don't mix 2 different things in the same pull request. (That `__install_debian_stable` was prefixed for a reason, for it **not** to be found unless explicitly :smile: )

twerkmeister:
Oh sorry, I didn't know following commits will also go into the pull request. Will fix it. The reason I removed the prefix is debian squeeze wouldn't find any installer function unless this prefix is removed, because the Install func naming schemes don't have install_${DISTRO_NAME_L}${PREFIXED_DISTRO_MAJOR_VERSION}. Maybe I should add this naming scheme instead?

s0undt3ch:
Ok, now, could you move the check to be just before [line 480](https://github.com/lesnail/salt-bootstrap/blob/7bf8434eef855dfa0262609fad4959b9625fb3c2/bootstrap-salt.sh#L480).

s0undt3ch:
About the prefixed function, debian stable packages are not yet being produced, hence we only allow some debian versions, and we might even remove those. You should try installing from git, for example:

```
curl -L http://bootstrap.saltstack.org | sudo sh -s -- git v0.14.1
```

s0undt3ch:
@lesnail Thanks for all the time and trouble.

PR #118, opened 2013-05-03
Debian installations now use SaltStack's repository

s0undt3ch:
Debian installations now use SaltStack's repository.

ZMQ3 support is optional and enabled when passing `-P`.


s0undt3ch:
@joehealy I'll be merging this soon. Could you please check if I haven't missed(or messed) anything please?

PR #125, opened 2013-05-13
force the use of the $SALT_ETC_DIR instead of hardcoded /etc/salt

mguegan:
Impact on :
- config_salt()
- SmartOS install functions


s0undt3ch:
Thanks @mguegan!

PR #127, opened 2013-05-14
smartos : add detection of GZ and make manifest persists across reboots

mguegan:
When started on a Global Zone (GZ) (or baremetal), SmartOS imports
manifests from /opt/custom/smf/ automatically.

This patch use dedicated SmartOS manifest from Salt.


s0undt3ch:
Thanks for this awesome contribution!

mguegan:
Ooh no no : it's this project which is awesome !

PR #128, opened 2013-05-14
daemons_running() : add a proper way to check SmartOS running services

mguegan:


bootstrap-salt.sh
```diff
@@ -2540,7 +2540,12 @@ daemons_running() {
         [ $fname = "master" ] && [ $INSTALL_MASTER -eq $BS_FALSE ] && continue
         [ $fname = "syndic" ] && [ $INSTALL_SYNDIC -eq $BS_FALSE ] && continue
 
-        if [ "x$(ps aux | grep -v grep | grep salt-$fname)" = "x" ]; then
+        if [ ${DISTRO_NAME} == "SmartOS" ]; then
```
s0undt3ch: Could you please replace `==` with `=`? It's more portable between shells

mguegan: Yes sure ! It was a mistake.

s0undt3ch:
Thank You for this contribution too!

PR #129, opened 2013-05-15
Set the FreeBSD /etc/salt dir to PREFIX/etc/salt

geoffgarside:
The FreeBSD port version of saltstack uses /usr/local/etc/salt rather
than /etc/salt for its configuration. This change fixes the use of a configuration directory when bootstrapping FreeBSD.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1976,6 +1976,12 @@ install_freebsd_9_stable_deps() {
     echo "PACKAGESITE: ${BS_PACKAGESITE}" > /usr/local/etc/pkg.conf
 
     /usr/local/sbin/pkg install -y swig || return 1
+
+    # Lets set SALT_ETC_DIR to ports default
+    if [ "$TEMP_CONFIG_DIR" != "null" ]; then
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -1976,6 +1976,12 @@ install_freebsd_9_stable_deps() {
     echo "PACKAGESITE: ${BS_PACKAGESITE}" > /usr/local/etc/pkg.conf
 
     /usr/local/sbin/pkg install -y swig || return 1
+
+    # Lets set SALT_ETC_DIR to ports default
+    if [ "$TEMP_CONFIG_DIR" != "null" ]; then
+        SALT_ETC_DIR="/usr/local/${SALT_ETC_DIR}"
```
s0undt3ch: Why not:

``` bash
SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/usr/local/etc/salt}
```

geoffgarside: Ah yes, as that will respect the environment variable as well won't it.

s0undt3ch: Yes, it will.

s0undt3ch:
Thanks!

PR #130, opened 2013-05-15
Fix daemons_running() on small window installations.

geoffgarside:
Handily ps(1) output is truncated to the size of the window, this causes the running check to fail on consoles with small windows as the salt-{minion,master,syndic} strings will not appear properly in the output which is passed to grep(1). The use of multiple -w passed to ps(1) prevents the truncation of the process lines.


s0undt3ch:
Thanks!

PR #132, opened 2013-05-15
Support for passing an HTTP or FTP URL for TEMP_CONFIG_DIR

geoffgarside:
Uses one of curl, wget or fetch to download the provided URL and then decompresses gzip, bzip2 and xz tarballs. The file is extracted to /tmp and the extracted path is set for TEMP_CONFIG_PATH.

I've managed to test this insofar as running it on a BSD VM, I'll setup and test on a Linux VM this evening.

I'm not quite sure what to do about writing tests for it to make sure it reliably works though.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -174,6 +174,7 @@ __check_config_dir() {
         *)
             if [ ! -e "${CC_DIR_NAME}" ]; then
                 echo "null"
+                return 0
             fi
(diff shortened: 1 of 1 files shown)


s0undt3ch:
This is awesome stuff!

s0undt3ch:
About the tests, we might try to get a local http server started with the test suite to do some unit testing...

PR #152, opened 2013-05-31
Fix FreeBSD git installation.

geoffgarside:
Move git installation so that it is installed before the git clone is performed.


s0undt3ch:
Hehehe.

Thanks @geoffgarside!

PR #162, opened 2013-06-13
Fixing options not recognizing -F

nevins-b:
Currently the script doesn't match -F in the command line arguments correctly. This change should add it to the options match.


s0undt3ch:
Doh! I forgot to add it!!!???

@nevins-b Thanks for the fix!

PR #166, opened 2013-06-18
order of operations should be install then config

teburd:
This fixes a bug with archlinux bootstrapping

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2991,18 +2991,6 @@ if [ $CONFIG_ONLY -eq $BS_FALSE ]; then
     fi
 fi
 
-
-# Configure Salt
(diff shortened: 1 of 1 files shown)


s0undt3ch:
Need to test to see if this does not break all other distributions. Still, Thanks for the contribution.

akoumjian:
Don't the debian/ubuntu packages start the services immediately after installing? This would break on those, I believe.

s0undt3ch:
That was what I was thinking. So I think I'm back to my original though. Arch might need to be handled alone regarding this subject.

s0undt3ch:
Err, got the tickets messed up.

This is not the proper fix. Sorry @bfrog. This is now handled in #168.

Issue #144, opened 2013-05-26
Support OmniOS (IllumOS derivative)


robinsmidsrod:
It'd be nice if OmniOS could be supported in the installer. It is a Solaris-based system that uses IPS packaging, same as OpenIndiana.

See http://omnios.omniti.com/wiki.php/Packaging for more details on how to package for OmniOS.

I believe the information below is what you need to at least detect the distro correctly.

```
# /sbin/uname -X
System = SunOS
Node = nas
Release = 5.11
KernelID = omnios-d3950d8
Machine = i86pc
BusType = <unknown>
Serial = <unknown>
Users = <unknown>
OEM# = 0
Origin# = 1
NumCPU = 8

# cat /etc/release
  OmniOS v11 r151006
  Copyright 2012-2013 OmniTI Computer Consulting, Inc. All rights reserved.
  Use is subject to license terms.
```

s0undt3ch:
Is there any cloud service providing an image of OmniOS?

robinsmidsrod:
@s0undt3ch Yes, I believe there are EC2 AMIs available. Details here: http://omnios.omniti.com/wiki.php/Installation#IntheCloud

s0undt3ch:
Cool!!! I'll see if I spin one up to at least have the script properly detect OmniOs.

## 

Pedro Algarvio @ phone

----- Reply message -----
De: "Robin Smidsrød" notifications@github.com
Para: "saltstack/salt-bootstrap" salt-bootstrap@noreply.github.com
Cc: "Pedro Algarvio" pedro@algarvio.me
Assunto: [salt-bootstrap] Support OmniOS (IllumOS derivative) (#144)
Data: qua, Jun 19, 2013 20:59
@s0undt3ch Yes, I believe there are EC2 AMIs available. Details here: http://omnios.omniti.com/wiki.php/Installation#IntheCloud 

&mdash;

s0undt3ch:
Quoting @robinsmidsrod on #169 so the comment does not get lost.

---

@s0undt3ch Well, technically it's using the same system as OpenIndiana (and traditional Solaris I guess), which are IPS packages. More details here: http://omnios.omniti.com/wiki.php/Packaging

[...]

PR #169, opened 2013-06-20
Properly detect OmniOS. Refs #144

s0undt3ch:
```
 *  INFO: sh bs.sh -- Version 1.5.5

 *  INFO: System Information:
 *  INFO:   CPU Arch:     i86pc
 *  INFO:   OS Name:      SunOS
 *  INFO:   OS Version:   5.11
 *  INFO:   Distribution: OmniOS r151006

 *  INFO: Installing minion
 * DEBUG: install_omnios_stable_deps not found....
 * DEBUG: install_omnios_deps not found....
 * DEBUG: install_omnios_stable not found....
 * DEBUG: install_omnios_stable_post not found....
 * DEBUG: install_omnios_post not found....
 * DEBUG: install_omnios_stable_restart_daemons not found....
 * DEBUG: install_omnios_restart_daemons not found....
 * DEBUG: daemons_running_omnios_stable not found....
 * DEBUG: daemons_running_omnios not found....
 * DEBUG: Removing the logging pipe /tmp/bootstrap-salt.logpipe
```

All we need now is someone to add the install functions :)

_hint hint_ @robinsmidsrod.... :smiley:

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -710,7 +710,7 @@ __gather_sunos_system_info() {
                     ;;
                 *OmniOS*)
                     DISTRO_NAME="OmniOS"
-                    DISTRO_VERSION=$(echo "$line" | awk '{print $3}'
+                    DISTRO_VERSION=$(echo "$line" | awk '{print $3}')
                     __SIMPLIFY_VERSION=$BS_FALSE
                     break
                     ;;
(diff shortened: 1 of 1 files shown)


robinsmidsrod:
@s0undt3ch Well, technically it's using the same system as OpenIndiana (and traditional Solaris I guess), which are IPS packages. More details here: http://omnios.omniti.com/wiki.php/Packaging

It can also run any traditional Solaris 10 binary, so if you already have that in your script, I believe you should be able to point it at those. If not, a typical from source build that works on Solaris and OpenIndiana _should_ also work here. The packages available from opencsw.org also work using the opencsw.org installer. Not sure exactly what kind of package manager you'd want to support, as the IPS package manager is for the core OS, but opencsw seems to have a lot more packages available for Solaris, which also work on OmniOS.

I've never built salt from source and I'm quite a newbie with it comes to all things Solaris/OpenIndiana/OmniOS, but I am fairly decent at testing. Could you give me a quick overview of what I'd need to test and how?

The install functions are all shell code, right? In terms of installing services, it's all standard Solaris SMF (which I also suck at). In terms of installing packages from the standard repos, you can read more about that here: http://omnios.omniti.com/wiki.php/GeneralAdministration#PackageManagement

Hopefully some of this is helpful in getting something done.

PR #171, opened 2013-06-27
Tiny fix TEMP_CONFIG_DIR -> TEMP_KEYS_DIR in preseed function definition

kozhukalov:


s0undt3ch:
Do we really need the debug calls? The script already says what's the function running.

kozhukalov:
Additional debug outputs can help because it says the calculated function name before it is called and even if it is null. There are some additional conditions checked before actual function launching. They are something like the following 

```
[ "$TEMP_KEYS_DIR" != "null" ] &&[ "$PRESEED_MASTER_FUNC" != "null" ]
```

So if condition is not true, then the function will not be called and you never will know the exact reason of why the function would not been called.

I added another tiny fix. I moved definition of PKI_DIR into the head of script because when you run script with -k option but without -c option 'preseed_master' function reports  unbound PKI_DIR variable.

s0undt3ch:
Ok. You convinced me :)

Thanks!

PR #181, opened 2013-07-09
SLES 11 SP3 support added to bootstrap script

aboe76:
Needed to tweak the script because of lsb_release output changed
to 'SUSE LINUX' instead of 'SUSE'


s0undt3ch:
Thanks!

PR #194, opened 2013-08-02
Emergency workaround for salt issue 6502

terminalmage:
This workaround will create the cachedir that 0.16.2 fails to (see
https://github.com/saltstack/salt/issues/6502)

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -3026,6 +3026,13 @@ if [ $CONFIG_ONLY -eq $BS_FALSE ]; then
     fi
 fi
 
+# Ensure that the cachedir exists
+# (Workaround for https://github.com/saltstack/salt/issues/6502)
+if [ $INSTALL_MINION -eq $BS_TRUE ]; then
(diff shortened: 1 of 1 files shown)


s0undt3ch:
I'd have to see the full changed source code and my phone is not the best tool for that. This seems like a fix for some distributions only, those using /vaR correct? Or is this supposed to fix all distros installing 0.16.2?

terminalmage:
Yeah, just those using /var. I think this can probably be reverted, to be honest. Addressing this with the packagers (as I have done separately via email) seems to be a better solution.

s0undt3ch:
Let's maintain the workaround until either 0.16.3 or 0.17.0 has been
released.

Pedro Algarvio  Pedro Algarvio
Contact me via mail(@algarvio.me mailto:pedro@algarvio.me), twitter
https://twitter.com/s0undt3ch, github https://github.com/s0undt3ch
or bitbucket https://bitbucket.org/s0undt3ch. Tip me via gittip
https://www.gittip.com/s0undt3ch/.

PR #197, opened 2013-08-05
Fix #189: Don't call yum update

### LINKED ISSUE #189: yum-based systems call 'yum update'
The issue is that `yum update` isn't like `apt-get update`. It actually goes and upgrades packages. If there are updates pending, or some other package has a broken dependency, then the whole install fails because of elements unrelated to salt.

Unless I've missed something, the goal of that call is to make sure the EPEL repository is updated, for which `yum makecache` should be sufficient.

cwacek:
This addresses the issues raised in Issue #189. 

`yum update` automatically upgrades all packages that can be updated, which is either annoying or catastrophic if it breaks running software. All Yum commands automatically refresh the cache before running, so this shouldn't be necessary.

Tested on Centos64 and found no issues.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1593,7 +1593,7 @@ install_centos_stable_deps() {
         return 1
     fi
 
-    yum -y update || return 1
+    yum -y makecache || return 1
 
     if [ $DISTRO_MAJOR_VERSION -eq 5 ]; then
         yum -y install PyYAML python26-m2crypto m2crypto python26 \
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1593,8 +1593,6 @@ install_centos_stable_deps() {
         return 1
     fi
 
-    yum -y makecache || return 1
-
     if [ $DISTRO_MAJOR_VERSION -eq 5 ]; then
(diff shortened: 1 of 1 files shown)


techhat:
I still maintain that this should be an option, not removed entirely.

cwacek:
I disagree. `yum update` upgrades all upgradeable packages on the system and is called _before_ Salt is even installed. What is the use case for the Salt installer to be updating random packages on the system?

I could see an argument for running `yum update <salt_dependencies>`, since you might conceivably want to make sure the things Salt requires are fully up to date. That's not what it's doing currently though.

techhat:
My strategy: spin up a new VM, upgrade it fully and install the packages I'm going to use (including Salt), and then use the VM for a short period of time. Then destroy the instance, and spin up a new one. I would never install new packages or upgrade an existing instance after it has entered production. That's just asking for trouble.

I don't care if the default is to not do a `yum update`; that's fine with me. The option should be there, for those who want to use it. Removing existing functionality just because you don't use it is a good way to alienate users who do use it.

s0undt3ch:
Ok, I'm sorry @cwacek but I'm not going to merge this pull request. I'll continue the discussion on #189

PR #209, opened 2013-08-19
updated FreeBSD PACKAGESITE to use freebsd.saltstack.com

cedwards:
I have generated and published FreeBSD salt packages at freebsd.saltstack.com. This request updates the bootstrap to use this infrastructure.


s0undt3ch:
Thanks!!!!

Issue #198, opened 2013-08-05
Bootstrap fails to install properly on RHEL systems if optional repo is not enabled.


gravyboat:
Currently when trying to run the bootstrap script (without additional options) on a rhel 6 64 bit machine salt as well as Jinja2 fail to install because the only package available in EPEL is [jinja2-26](http://dl.fedoraproject.org/pub/epel/6/x86_64/repoview/python-jinja2-26.html). This was previously brought up as a failure for the salt master to install [here](https://groups.google.com/forum/#!msg/salt-users/ZyatYnb95QY/ADngjIjVF8cJ), and [bug 844710](https://bugzilla.redhat.com/show_bug.cgi?id=844710) was filed, but not considered to be an actual problem as the optional channel for RHEL can be enabled. 

If we're unable to address this issue in the packages themselves (and really that should NOT need to be done, the RPM available in EPEL should be corrected), then the RHEL function could be updated to run the following command <code>rhn-chanel --add --channel=rhel-$ARCH-server-optional-$DISTRO_VERSION</code>, so for x86_64, on rhel6 it would look like <code>rhn-chanel --add --channel=rhel-x86_64-server-optional-6</code>.

s0undt3ch:
Thanks @gravyboat !

s0undt3ch:
@herlo can you please tell us what you think should be done regarding this issue?

herlo:
There are two problems with this suggestion. First, if I recall correctly, putting the package in as a Requires for salt would mean we'd be using the wrong jinja2 package. Secondly, enabling the repository seems like a good idea, except for the fact that some subscriptions do not include the optional repository. 

I think the second one is more problematic. Maybe salt-bootstrap could have a distro/OS specific set of flags which can be passed to enable things such as this?

herlo

gravyboat:
@herlo Yes I agree that putting the package in as a requires is a bad idea. The optional repo is also problematic, since then individuals who don't have an optional repo subscription are completely out of luck when it comes to installing salt via the bootstrap unless they decide to pull from git. Is there no way we can contact Thomas Moschny and request that he fix the package in EPEL? His email is freely available with a quick Google search (thomas.moschny@gmx.de) , and I don't believe the bugzilla was closed with a response that took the situation you provided into account.

[...]

PR #211, opened 2013-08-19
Update to install_red_hat_linux_stable_deps function to resolve issue 198.

gravyboat:
This pull addresses concerns I raised in [issue 198](https://github.com/saltstack/salt-bootstrap/issues/198) regarding how the bootstrap errors out without a useful message for redhat users. Due to the way that the rhn-channel command operates (requiring a user/password to join to a channel) I've simply added in an error explaining that users need to enable the optional repo. At some point it might be good to expand on this, but I don't see a good way to do it right now.


gravyboat:
Updated to reduce duplication.

PR #228, opened 2013-09-24
Improve Gentoo support

eliasp:
This PR improves some Gentoo related items:
- Don't hardcode all dependencies in the bootstrapping script, but instead let Gentoo's package manager (Portage) handle this on its own by writing to the corresponding /etc/portage/package.\* files
- Use `eix` where available
- Use the package category when installing 'salt' to prevent possible naming collisions
- Add support for systemd on Gentoo (the Gentoo ebuild still needs to catch up and install the systemd units)
- Don't let the bootstrapping script pollute the worldfile with salt dependencies

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2800,7 +2800,6 @@ install_gentoo_deps() {
 }
 
 install_gentoo_git_deps() {
-    __emerge git
     __gentoo_pre_dep || return 1
     echo "=app-admin/salt-9999 **" >> ${GENTOO_ACKEYS}
     __gentoo_post_dep
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2789,9 +2789,9 @@ __gentoo_pre_dep() {
 }
 __gentoo_post_dep() {
     # ensures dev-lib/crypto++ compiles happily
-    __emerge libtool
+    __emerge 'sys-devel/libtool'
     # the -o option asks it to emerge the deps but not the package.
-    __emerge -vo salt
+    __emerge -vo 'app-admin/salt'
 }
 
 install_gentoo_deps() {
@@ -2805,7 +2805,7 @@ install_gentoo_git_deps() {
 }
 
 install_gentoo_stable() {
-    __emerge -v salt || return 1
+    __emerge -v 'app-admin/salt' || return 1
 }
 
 install_gentoo_git() {
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2779,9 +2779,17 @@ __emerge() {
 
 __gentoo_pre_dep() {
     if [ $_ECHO_DEBUG -eq $BS_TRUE ]; then
-        emerge --sync
+        if which eix &> /dev/null; then
+            eix-sync
+        else
+            emerge --sync
+        fi
     else
-        emerge --sync --quiet
+        if which eix &> /dev/null; then
+            eix-sync -q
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -2772,82 +2772,64 @@ install_suse_11_restart_daemons() {
...
+    # this point, manually merge the changes using etc-update/dispatch-conf/
+    # cfg-update and then restart the bootstrapping script, so instead we allow
+    # at this point to modify certain config files directly
+    export CONFIG_PROTECT_MASK="$CONFIG_PROTECT_MASK /etc/portage/package.keywords /etc/portage/package.unmask /etc/portage/package.use /etc/portage/package.license"
 }
 
 __gentoo_pre_dep() {
     if [ $_ECHO_DEBUG -eq $BS_TRUE ]; then
-        emerge --sync
+        if which eix &> /dev/null; then
```
s0undt3ch: Please update to:

``` bash
if [ "x$(which eix)" != "x" ]; then
```

bootstrap-salt.sh
```diff
@@ -2772,82 +2772,64 @@ install_suse_11_restart_daemons() {
...
+            eix-sync
+        else
+            emerge --sync
+        fi
     else
-        emerge --sync --quiet
+        if which eix &> /dev/null; then
```
s0undt3ch: Please update to:

``` bash
if [ "x$(which eix)" != "x" ]; then
```

s0undt3ch:
WOW! Massive update on the Gentoo front! Thanks!!!!!

bootstrap-salt.sh
eliasp: I understand what this will do and will change it, I'm just wondering what's the reason to do this?

s0undt3ch: This, 

``` bash
if which eix &> /dev/null; then
```

is considered a bashism, ie, code expected to run under bash but not under plain sh. We enforce plain sh, even though Gentoo might default to bash being the default shell.

s0undt3ch:
You removed the hard-coded flags and it's been quite some time since I messed with a Gentoo box, so, is it ok not to hard code the flags? `--autounmask-write` will take care of that right?

eliasp:
Yes, since quite some time Portage supports `--autounmask-write` which will take care of this completely (set required entries in `package.keywords`, `package.use` and even `package.unmask` and `package.license`).

No more checking for the entries in `/etc/portage/package.{keywords,use,…}` whether they're directories or files etc.
No more keeping track of the dependencies for Salt in the bootstrapping script.
All this workarounds are a thing of the past now.

By default the `--autounmask-write` feature respects `CONFIG_PROTECT` which means bootstrapping would have to be interrupted, changes to the files merged manually using `etc-update`/`dispatch-conf`/`cfg-update`, but I'm bypassing this by setting a matching `CONFIG_PROTECT_MASK`.

bootstrap-salt.sh
eliasp: Thanks for the explanation. I knew I had to avoid bashisms, but didn't know `&>` is one…

s0undt3ch: That's not it, the missing square brackets is, I think :smile:

s0undt3ch:
Thanks for the explanation!

bootstrap-salt.sh
eliasp: Ah, well… whatever solution works :)

Issue #224, opened 2013-09-10
broken on SmartOS


robinbowes:
I tried using salt-bootstrap on SmartOS with pkgin configured and found that it's broken.

I dug in and found a few issues:
1. There is no gcc-compiler package. Not sure why this is missing, or if it ever existed (presumably did?)
2. logic error when installing gcc47. This:

```
pkg_info gcc47 > /dev/null 2>&1 && pkgin -y in gcc47
```

should be:

```
pkg_info gcc47 > /dev/null 2>&1 || pkgin -y in gcc47
```
1. gcc47 doesn't leave gcc in the path, it installs to /opt/local/gcc47/bin/gcc
2. I added /opt/local/gcc47/bin/ to the path but the zeromq build fails with:

```
Making all in src
make[1]: Entering directory `/root/zeromq-3.2.3/src'
make  all-am
make[2]: Entering directory `/root/zeromq-3.2.3/src'
  CXX    libzmq_la-xpub.lo
make[2]: *** [libzmq_la-xpub.lo] Error 1
make[2]: Leaving directory `/root/zeromq-3.2.3/src'
make[1]: *** [all] Error 2
make[1]: Leaving directory `/root/zeromq-3.2.3/src'
make: *** [all-recursive] Error 1
 * ERROR: Failed to run install_smartos_git_deps()!!!
```

I noticed that there is a salt package in pkgin but it is 0.15.3. Might be worth getting that updated? I believe the maintainer is Jonathan Perkin: http://www.perkin.org.uk/pages/pkgsrc-binary-packages-for-illumos.html

R.

techhat:
Hi @robinbowes. Indeed there was a gcc-compiler package at one point. My first thought is that either the system that you're on isn't pointing to a repo with gcc-compiler, or it is but the package was removed from the repo. Do you know which pkgin repo your system is pointed to?

robinbowes:
I'm using the 2013Q2 repo:

http://pkgsrc.joyent.com/packages/SmartOS/2013Q2/x86_64/All

R.

techhat:
Thanks, @robinbowes, I'll take a look at it. Is it safe to assume this is a VM on JoyEnt? If so, can you tell me which image you spun up?

robinbowes:
Jo,

I am trying to install salt in the global zone of a SmartOS machine, version joyent_20130905T204057Z

Actually, I only need to install the minion.

R.

mguegan:
Hi ! 
[...]

PR #239, opened 2013-09-26
Issue #224 : update SmartOS installation procedure

mguegan:
This fix intend to resolve the Issue #224.
- Make use of 2013Q2 package set from Joyent : this would require
  a recent image of SmartOS
- NO compat with previous 2013Q1 and before
- Tested with 'stable' version of Salt (provided by Joyent : 0.15) on a global zone.
- As git can't be installed on the global zone via pkgin, 'git develop' version is currently broken on a global zone.
- Should be working on a VM / Zone (not yet tested)


s0undt3ch:
Thanks!!!

PR #246, opened 2013-10-01
 Added oracle linux funcs based of CentOS, only tested git develop with OracleVMServer 3.2

adonm:
Stable should work, on Oracle Linux 5 & 6, to be tested
git develop branch is tricky due to python2.7 requirement, works now but actually builds python2.7 and installs to opt (only tested on oraclevm server)


adonm:
Doh scratch that - stable works but git doesn't, was fooled by stable minion still checking in

Issue #245, opened 2013-09-30
Bootstrap does not install git-core (on stable), or gitpython (on stable or git).


gravyboat:
It was noted by xuru in the IRC today that the bootstrap script is not installing git-core/gitpython when performing an installation using the stable option (or no option), this makes it so you are unable to use gitfs. If calling the script using the git option it does install git-core, however gitpython is still not installed. Is this by design?

xuru:
When using the install type of 'stable' I have to do the following in order to have gitfs working (from the vagrant file):
  config.vm.provision :shell, :inline => "sudo apt-get install git-core"
  config.vm.provision :shell, :inline => "sudo apt-get install python-setuptools"
  config.vm.provision :shell, :inline => "sudo easy_install GitPython"

Seems like we can can install these packages by default.  The only concern would be any security issues installing git on a production system.  However, we could add an argument to have it install git tools...

s0undt3ch:
The bootstrap script only installs what's required for salt to run, ie, it's hard dependencies.
- Git, although it's not a hard dependency to install Salt, it **is** required if you're trying to install salt from git
- Git Python is not a hard dependency to Salt, as such, it won't be installed

I know it's not what you were after but, you can install those packages once salt is installed using salt.

I'm closing as _won't fix_, at least for now.

v6:
// , This does not imply to start out with some sort of shell provisioning, because maybe one could make the first few salt states that run install Git and Python prerequisites for the other states?

PR #248, opened 2013-10-02
Adds GitFS dependencies for Ubuntu

ifnull:


gravyboat:
The install_ubuntu_git_deps function is for installing packages that are required to do a checkout from git for Salt itself. It is not used for installing pip or gitpython as neither of those are required for salt to function. I previously asked a question regarding this over at issue https://github.com/saltstack/salt-bootstrap/issues/245 and an answer was provided regarding why changes such as this will not be approved.

s0undt3ch:
I think I've mentioned this before on an issue or another pull request, but here it goes again.

The same way Salt supports git as a back-end to it's file roots and pillar, it also supports mercurial, subversion, amaon's S3, and this list might grow as time goes by.
So, I'm going to reject this pull request but the discussion is still open.

We have two choices.
1. If all distros support something like "optional", or "suggestions", which is unlikely, we can add a flag which installs those, all of them
2. We add a flag which accepts a list of packages to be installed, additionally.

I'm thinking of 2.

Thoughts?

gravyboat:
I like the idea of accepting a list of packages to be installed. This would allow it to continue to be OS agnostic, and reduce the workload on constantly maintaining the names of rpms/debs or whatever else.

s0undt3ch:
Yep!

Issue #255, opened 2013-10-30
[FreeBSD] Repository "salt" not always available


wunki:
In `install_freebsd_9_stable_deps` you are trying to install `swig` from the salt repository. See [line 2360](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L2360).

But that repository is only setup when your box doesn't have pkg already installed. See [line 2343](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L2343).

I'm running on a vagrant box which already has pkgng setup and I do have `swig` in my repository, but not in a special "salt" repository, thus receiving an error `pkg: Unknown repository: salt`.

Maybe a check to only add `-r salt` if the repository is added by the script. Otherwise fetch it from the repository which is already setup?

mortn:
Thank you! That did the trick on FreeBSD 9.1 i386

s0undt3ch:
Closing as the fix is now part of the script.

PR #268, opened 2013-11-21
Fix bootstrap.sh for FreeBSD 9.2

millerdev:
`pkg` is installed by default on FreeBSD 9.2, and Salt is in the FreeBSD package repositories.

Without this fix I was getting the following error:

```
pkg: Unknown repository: salt
 * ERROR: Failed to run install_freebsd_9_stable_deps()!!!
```


millerdev:
This may be a duplicate of #255, but this is more complete because it also patches `install_freebsd_git`.

cedwards:
Looks OK to me

s0undt3ch:
Awesome!

PR #273, opened 2013-12-16
salt-bootstrap only runs with sudo. Changed README file to reflect that.

thedrow:


techhat:
Did salt-bootstrap stop working with fully logged in as root? Almost all of the cloud providers in salt-cloud use root as the default login.

thedrow:
I'm installing salt on a desktop. sudo should be specified by default imo for consistency.

techhat:
I must disagree. The purpose of those examples is to demonstrate the versatility in which the script can operate. Ignoring the default operating mode of the majority of the environments that this will run in is not correct; if anything, fewer of the examples should have sudo.

PR #276, opened 2013-12-22
Add mint 16 eqiuvalence to Ubuntu 13.10

nmbooker:
This adds linux mint 16 support by declaring that it's Ubuntu base is 13.10.

Tested with no switches.  Not tried installing git version or anything else.


s0undt3ch:
Thanks!

PR #336, opened 2014-03-05
Verify if the Upstart config file for Salt are installed. If not, use

diegows:
SysV scripts.

Upstart config files are not packages yet.


s0undt3ch:
This is using a ["bashism"](https://travis-ci.org/saltstack/salt-bootstrap/builds/20162376)... 

Please replace:

``` bash
if [ -f /sbin/initctl -a -f /etc/init/salt-$fname ]; then
```

with

``` bash
if [ -f /sbin/initctl ] && [ -f /etc/init/salt-$fname ]; then
```

PR #364, opened 2014-04-04
Avoid leaving unnecessary file after bootstrap install. 

oreh:
pip-python treats the version requirement, ">=", as an output redirection. As a result, a "=0.1.4.0" file was created when installing libcloud. Added quotation to avoid this problem.


s0undt3ch:
Test FAILed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap-pr/7/

s0undt3ch:
Thank You for catching this!

PR #402, opened 2014-06-11
Changing errors to warning and removing return statements. Why return?

nmadhok:
There's no need to return. Instead just display a warning message to users and continue execution. This solves the problem where the instance creation using cloud module fails giving the following error:

```
test-master:
----------
          ID: test.minion
    Function: cloud.present
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python2.6/site-packages/salt/state.py", line 1492, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.6/site-packages/salt/states/cloud.py", line 106, in present
                  info = __salt__['cloud.create'](cloud_provider, name, **kwargs)
                File "/usr/lib/python2.6/site-packages/salt/modules/cloud.py", line 202, in create
                  info = client.create(provider, names, **kwargs)
                File "/usr/lib/python2.6/site-packages/salt/cloud/__init__.py", line 362, in create
                  mapper.create(vm_))
                File "/usr/lib/python2.6/site-packages/salt/cloud/__init__.py", line 1162, in create
                  output = self.clouds[func](vm_)
                File "/usr/lib/python2.6/site-packages/salt/cloud/clouds/openstack.py", line 773, in create
                  ret = salt.utils.cloud.bootstrap(vm_, __opts__)
                File "/usr/lib/python2.6/site-packages/salt/utils/cloud.py", line 418, in bootstrap
                  deployed = deploy_script(**deploy_kwargs)
                File "/usr/lib/python2.6/site-packages/salt/utils/cloud.py", line 1204, in deploy_script
                  deploy_command
              SaltCloudSystemExit: Executing the command '/tmp/.saltcloud/deploy.sh -c /tmp/.saltcloud' failed
     Started: 14:41:56.077120
     Duration: 867 ms
     Changes:   

Summary
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
```

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2542,11 +2542,9 @@ install_red_hat_linux_stable_deps() {
     fi
     if [ $DISTRO_MAJOR_VERSION -eq 6 ]; then
       if rhn-channel -b >/dev/null 2>&1 && case "X$(rhn-channel -l | grep optional)" in Xrhel-${OPTIONAL_ARCH}-server-optional-${DISTRO_MAJOR_VERSION}* ) false ;; * ) true ;; esac ; then
-        echoerror "Failed to find RHN optional repo, please enable it using the GUI or rhn-channel command."
-        return 1
+        echowarn "Failed to find RHN optional repo, please enable it using the GUI or rhn-channel command."
       elif ! (yum repolist | grep -s -q server-optional-rpms); then
-        echoerror "Failed to find server-optional-rpms repo, please enable it using yum-config-manager."
-        return 1
+        echowarn "Failed to find server-optional-rpms repo, please enable it using yum-config-manager."
       fi
     fi
     install_centos_stable_deps || return 1


ghost:
Test FAILed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/30/

s0undt3ch:
This is not the way to go and we return because execution must stop. I'll submit the fix in some mins.

s0undt3ch:
https://github.com/saltstack/salt-bootstrap/commit/4912150f6b3c6983550b7ebcfa1793ec5d9a1df3

techhat:
@s0undt3ch, why is it that execution must stop?

techhat:
@s0undt3ch in production, it is very common for orgs to set up their own repos instead of using RHN, even if they have a license. This is the use case with @nmadhok's org. What would you suggest to allow this use case? Would it be reasonable to add an option to allow the script to continue past this point instead of returning?

nmadhok:
@s0undt3ch What happens if we don't return? Is it destructive if we remove the returns? We need to work together on this and cover the different use cases. It'll be helpful for me to contribute if i know the behavior in other cases as well if the returns are removed. 

@techhat True. Currently there's no option to use packages from private repository. The execution will fail and it would return if server-optional-repo doesn't exist or if the server doesn't use the rhn channel. I think adding an option would be a good idea and give us and other organizations that use private repositories more control over it.

s0undt3ch:
The script's goal is to install Salt, as such, if it "detects" that it's not going to be possible, execution stops.
The earlier we stop, the less "crippled" we leave a system.

That's why we stop. 'return 1' is a "full stop".

Now I understand why it is so important to bypass the RHN check, but we can't just simply ignore it.....

We could search for the packages which commonly live on the optional repository as opposed to the actual repository.
With Cent and RHEL 5 we could get away with this but @techhat pointed out that with RHEL >= 6 we're able so search for packages on repositories for which we don't have the required license, so we need to address this differently.

I'll be doin' some more investigation latter this evening(UTC)

ghost:
Test FAILed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/31/

PR #406, opened 2014-06-19
Added support for http proxy

### LINKED ISSUE #397: Add support for http proxy
When bootstrapping an instance in a private aws VPC, the script fails running the first apt-get update, because the instance has no public IP and it is not behind NAT.
It would be nice to have an option to set an http/ftp proxy to the package manager configuration.

giannello:
Fixes #397
Allows to provision an instance without having direct access to internet, using a proxy.

Invoke the script with "-H http://proxy:port"


ghost:
Test FAILed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/34/

s0undt3ch:
Thank You!

PR #432, opened 2014-07-24
SLES 11 SP3 Fixes

s0undt3ch:
SLES 11 SP3 ships with both python-M2Crypto-0.22.\* and python-m2crypto-0.21 and we will be asked which we want to install, even with --non-interactive.
Let's try to install the higher version first and then the lower one in case of failure.


ghost:
Test PASSed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/50/

PR #445, opened 2014-08-01
Increase default sleep period to avoid stepping on upstart.

cachedout:
The default sleep period would cause a race condition with upstart
wherein the package installation would call an upstart start and before
it could complete, bootstrap would call another. The result was _two_
copies of salt running which ended up causing a most stubborn bug that's
documented in saltstack/salt#12248


ghost:
Test PASSed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/62/

PR #453, opened 2014-08-29
Updated EPEL url for CentOS 7 to match the new EPEL release

dagvl:
The bootstrap script currently fails on CentOS 7 because the EPEL url for EPEL 7 has changed. I've updated the location, and tested that it works for me now.


ghost:
Test PASSed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/66/

s0undt3ch:
Thanks!

PR #476, opened 2014-10-16
improve __git_clone_and_checkout to support saltrepo srcurl wit...

leomeih:
There is a bug in current -g option and __git_clone_and_checkout function. For example:  
./bootstrap-salt.sh -g "http://example.com/git/saltstack" -M -F -L git develop  
will encounter a fatal error and break the installation complaining the '/tmp/git/salt' repository cannot be found.  
That's because the git clone command lines in __git_clone_and_checkout use the default name 'saltstack' as the reponame instead of __SALT_GIT_CHECKOUT_DIR base name.  
Manually setting environment variable BS_SALT_GIT_CHECKOUT_DIR="/tmp/git/saltstack" is a way to bypass the error.  
However, the improved __git_clone_and_checkout is a more robust solution.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1124,7 +1124,10 @@ __git_clone_and_checkout() {
 
     echodebug "Installed git version: $(git --version | awk '{ print $3 }')"
 
-    __SALT_GIT_CHECKOUT_PARENT_DIR=$(dirname "${__SALT_GIT_CHECKOUT_DIR}")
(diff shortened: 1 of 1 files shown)


ghost:
Test PASSed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/71/

s0undt3ch:
Thanks!

PR #480, opened 2014-10-19
Check services without using runlevel (fixes #479)

### LINKED ISSUE #479: Checking Debian services fails when chrooted from a systemd system
[Line 1547](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L1547) uses `runlevel` command to determine whether it is enabled in the current runlevel. However, when chrooted from a systemd OS (ArchLinux in my case) the utility just doesn't work:

```
# runlevel
unknown
```

Wouldn't it be more sensible to check whether the services are enabled in the usual runlevels 2345?

[Here](https://gist.github.com/angel333/927d11a7ad96b9662f46) is whole output.

angel333:
The bootstrap script should not rely on the "runlevel" utility as it
reports "unknown" when used in a systemd OS. One use case is a Debian
system inside of a ArchLinux chroot.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1544,7 +1544,7 @@ __check_services_debian() {
     echodebug "Checking if service ${servicename} is enabled"
 
     # shellcheck disable=SC2086,SC2046
-    if [ -f /etc/rc$(runlevel | awk '{ print $2 }').d/S*${servicename} ]; then
+    if [ $(ls /etc/rc[2345].d/S*${servicename} | wc -l) -eq 4 ]; then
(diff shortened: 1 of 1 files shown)


ghost:
Test PASSed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/74/

mbrgm:
+1

s0undt3ch:
The current develop version of the script can now skip `-d` checking services. Closing as a proposed fix for this.
Please let me know if you prefer a different fix.

PR #490, opened 2014-11-07
Enabled binary installation on RHEL 7

ggillies:
RHEL 7 is now available and the packages in EPEL 7 for salt work
correctly. We can now do epel installs onto RHEL 7

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2921,11 +2921,6 @@ install_red_hat_linux_git_deps() {
     return 0
 }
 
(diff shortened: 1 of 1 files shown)


ghost:
Test PASSed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/78/

ggillies:
I have tested this change on both RHEL 7 (with active subscriptions) and CENTOS 7 and it works correctly

s0undt3ch:
Thanks!

PR #493, opened 2014-11-14
Fix install on Debian Docker images (fix runlevel problem #479)

ross-p:
An alternate approach to fixing the runlevel problem https://github.com/saltstack/salt-bootstrap/issues/479 as compared to PR https://github.com/saltstack/salt-bootstrap/pull/480 (though that also looks like it works).

In this approach:
- No changes for most users
- In Debian Docker images (and perhaps other environments) when `runlevel` == "unknown":
  - DO NOT check that the services will start at boot.
  - DO NOT exit with a fatal error caused by not being able to check that the services will start at the current runlevel.

In Docker the startup scripts don't get run so this is a pointless check in that event anyway, and without this fix Salt cannot be installed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+/.idea/
 /.project
 *.swp
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -208,6 +208,7 @@ _SALT_MINION_ID="null"
 # __SIMPLIFY_VERSION is mostly used in Solaris based distributions
 __SIMPLIFY_VERSION=$BS_TRUE
 _LIBCLOUD_MIN_VERSION="0.14.0"
+_PY_REQUESTS_MIN_VERSION="2.4.3"
 _EXTRA_PACKAGES=""
 _HTTP_PROXY=""
 __SALT_GIT_CHECKOUT_DIR=${BS_SALT_GIT_CHECKOUT_DIR:-/tmp/git/salt}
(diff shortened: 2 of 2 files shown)


ghost:
Test PASSed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/79/

s0undt3ch:
I like this approach better.

The script does attempt to enable services, thus bypasses that too right?

ross-p:
The services are still "enabled".

If you look in /etc/rc_.d/S_salt-\* there are symlinks there, so if the startup scripts were to run, the services would be started.

The error was just that the script didn't expect $(runlevel) to return "unknown", which it does under Docker and apparently in some chroot environments.

In those cases the startup scripts will never be run by the OS anyway, so it doesn't matter if they are enabled or not.

ghost:
Test PASSed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/83/

slafs:
Looks nice. But I guess the issue still exists when building from ubuntu images, right?

ross-p:
Ubuntu installed from packages may or may not have the problem.  I'm not sure what version of python-requests they're distributing.

slafs:
Actually I was asking about the docker/runlevel thing.

ross-p:
@slafs Ubuntu uses the Debian procedure, so this should fix Ubuntu as well.

I notice that Github decided to pull in updates I made to my branch after I submitted the pull request instead of locking the sha of my commit.

if you want to merge this, I'll make a new pull request that contains only the fix intended for this issue.

s0undt3ch:
@ross-p yeah, please rebase just for the initial fix.

PR #494, opened 2014-11-14
Fix lsb_release detection for Debian Jessie

tjuberg:
This allows salt to correctly detect Debian Jessie and sets the correct version number.
Non lsb (release file detection) still fails due to /etc/os-release being a symlink and it not containing VERSION.
I see no reason to start touching that as there is no telling how os-release will behave for future debian testing releases. lsb detection should be reliable and now warn if an unknown debian testing release is found and allow to easilly add support for those in the future.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -718,6 +718,20 @@ __gather_linux_system_info() {
             return
         fi
         rv=$(lsb_release -sr)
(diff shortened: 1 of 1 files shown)


ghost:
Test PASSed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/80/

tjuberg:
No longer relevant.
Closing.

PR #501, opened 2014-11-21
Fix centos 7 systemd start from git install

### LINKED ISSUE #499: Centos 7: Salt bootstrap fails when using Git as a source
```
 *  INFO: Running install_centos_git_post()
Note: Forwarding request to 'systemctl enable salt-minion.service'.
Note: Forwarding request to 'systemctl enable salt-master.service'.
 *  INFO: Running install_centos_check_services()

Note: This output shows SysV services only and does not include native
      systemd services. SysV configuration data might be overridden by native
      systemd configuration.

      If you want to list systemd services use 'systemctl list-unit-files'.
      To see services enabled on particular target use
      'systemctl list-dependencies [target]'.

 * ERROR: Failed to run install_centos_check_services()!!!
```

This is because install_centos_git_post is using init.d scripts but the RPMs for the stable version are using systemd service files.

highlyunavailable:
Since the RPMs are now using systemd so should we. Added systemd-python as a
dependency for centos_git.

Fixes #499.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2689,9 +2689,9 @@ install_centos_git_deps() {
     install_centos_stable_deps || return 1
     if [ "$DISTRO_NAME_L" = "oracle_linux" ]; then
         # try both ways --enablerepo=X disables ALL OTHER REPOS!!!!
-        yum -y install git || yum -y install git --enablerepo=${_EPEL_REPO} || return 1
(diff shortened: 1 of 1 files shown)


ghost:
Test PASSed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/91/

s0undt3ch:
Awesome! Thanks!

PR #504, opened 2014-11-22
opensuse 13.2: fix installation

rominf:
zypper in openSUSE 13.2 now checks for file conflicts. When you install a package that replaces files from other (already installed) package it warns you and asks what to do. It breaks bootstrap script because `python-pyzmq` conflicts with `libzmq3` (I don't know why).

I've fixed this by specifying additional argument that forces file overwrite. Also I put all `zypper` arguments into two variables to follow DRY and make code more readable and maintainable.

I'm not an expert in bootstrap script, so I didn't know where to put variables definition. If there is a better way, please fix it. Probably you would like to change version checking too (to `sed` for example). In my opinion this way is the cleanest, moreover openSUSE and SUSE always used GNU version of `sort`, so it's not a problem.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -17,7 +17,7 @@
 #       CREATED: 10/15/2012 09:49:37 PM WEST
 #======================================================================================================================
 set -o nounset                              # Treat unset variables as an error
-__ScriptVersion="2014.10.30"
+__ScriptVersion="2014.11.22"
 __ScriptName="bootstrap-salt.sh"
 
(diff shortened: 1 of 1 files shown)


ghost:
Test PASSed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/94/

ghost:
Test PASSed.
Refer to this link for build results: http://cookingwithsalt.org/job/bootstrap/job/pr/job/build/95/

PR #507, opened 2014-12-02
Add python-gnupg dependency install for gpg render support.

christianchristensen:
xref: http://docs.saltstack.com/en/latest/ref/renderers/all/salt.renderers.gpg.html

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1788,7 +1788,7 @@ install_ubuntu_daily_deps() {
 install_ubuntu_git_deps() {
     install_ubuntu_deps || return 1
     __apt_get_install_noinput git-core python-yaml python-m2crypto python-crypto \
-        msgpack-python python-zmq python-jinja2 || return 1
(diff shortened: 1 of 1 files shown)


christianchristensen:
Idea here being support for [all renders](http://docs.saltstack.com/en/latest/ref/renderers/all/) should be supported at bootstrap.

When using a `#! jinja | yaml | gpg` template on initial run the `python-gnupg` dependency is needed.

christianchristensen:
Note: this needs work for other systems deps.

s0undt3ch:
I'm sorry for the trouble you had but this is not a hard dependency for salt, and, as such, it's not the script's responsibility to install it. Once salt is installed, you can use it to install the optional dependencies.

Again, I'm sorry for the trouble but I'm closing this PR.

s0undt3ch:
There is a switch which allows you to install extra pip dependencies at bootstrap time.

christianchristensen:
Thanks @s0undt3ch ! Saw the `-p` flag, for packages, and that seems to be a good way to handle this case.

PR #511, opened 2014-12-09
Force install of wget in debian

ptonelli:
When you use containers or minimal debian machines, wget is not available (specifically the debian lxc containers). This patch makes sure the wget commands work.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1993,6 +1993,9 @@ install_debian_6_deps() {
 
     apt-get update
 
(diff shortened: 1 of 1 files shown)


s0undt3ch:
Is curl or fetch available on minimal machines?

ptonelli:
No, curl or fetch are not installed. This is the standard container created under debian wheezy with the command::

`# lxc-create -n name -t debian`

s0undt3ch:
Sorry it took so long. Thanks!

PR #514, opened 2014-12-11
EPEL RPM file is now 7.5 instead of 7.2

cro:


s0undt3ch:
I'll release a stable and will PR salt with it..

PR #518, opened 2014-12-14
Make -U work for FreeBSD

eshagh:
When using http://files.wunki.org/freebsd-9.2-amd64-wunki.box or
http://files.wunki.org/freebsd-10.0-amd64-wunki.box with Vagrant,
bootstrap would succeed, but a number of states would fail without
a package upgrade prior to running highstate. The -U switch currently
does not trigger an upgrade when bootstraping FreeBSD. This patch adds
a package upgrade to the end of the FreeBSD dependency install
function.


s0undt3ch:
Thanks! Sorry for the delay.

PR #523, opened 2014-12-23
libzmq3(.2.3) has moved from unstable to backports in debian 7

devaos:
Pulling in unstable should no longer be necessary - https://packages.debian.org/wheezy-backports/libzmq3


s0undt3ch:
Sorry for the delay.

Thanks!

Issue #535, opened 2015-01-27
Fix for failing to install repository when wget is not available


The-Loeki:
On certain systems bootstrap failed because wget wasn't there.

``` shell
__install_saltstack_copr_zeromq_repository() {
    echoinfo "Installing Zeromq >=4 and PyZMQ>=14 from SaltStack's COPR repository"
    if [ ! -f /etc/yum.repos.d/saltstack-zeromq4.repo ]; then
        if [ "${DISTRO_NAME_L}" = "fedora" ]; then
            __REPOTYPE="${DISTRO_NAME_L}"
        else
            __REPOTYPE="epel"
        fi
-        wget -O /etc/yum.repos.d/saltstack-zeromq4.repo \
+       __fetch_url /etc/yum.repos.d/saltstack-zeromq4.repo \
                         "http://copr.fedoraproject.org/coprs/saltstack/zeromq4/repo/${__REPOTYPE}-${DISTRO_MAJOR_VERSION}/saltstack-zeromq4-${__REPOTYPE}-${DISTRO_MAJOR_VERSION}.repo" || return 1
    fi
    return 0
}
```

s0undt3ch:
Can you please make this change as a pull request so we can give proper credits to you?

Thanks!

s0undt3ch:
Fixed by your PR. Thanks!

PR #537, opened 2015-01-27
Fix for #535 failing to install repository when wget is not available on Red Hat systems

The-Loeki:
PR for #535


s0undt3ch:
Thanks!

PR #541, opened 2015-02-05
Minion keys and /etc/salt/minion should be overwritten on -C

cro:
If user passes -C to configure only, overwrite any existing minion keys or /etc/salt/minion.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4468,24 +4468,24 @@ config_salt() {
         CONFIGURED_ANYTHING=$BS_TRUE
     fi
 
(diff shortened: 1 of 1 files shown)


s0undt3ch:
There's a flag to force overwrites on existing files. Web shouldn't just overwrite without users being aware...

s0undt3ch:
@cro any thoughts about my comment above?

cro:
Sorry, I lost track of this issue.

All -C does _is_ overwrite files, though, correct?  Isn't it redundant to have someone pass --force and -C?

s0undt3ch:
I think I'll default overwrite to True when -C is passed and wait some seconds for user interruption with a big fat warning.

PR #542, opened 2015-02-06
Fixes for CentOS 7 w/LibCloud & Syndic

The-Loeki:
Two fixes for installation on RHEL/CentOS 6/7 systems;
- 7 doesn't have 'pip-python' symlink to pip; just modified it to run 'pip'; which exists on both
- Syndic is in it's own package in EPEL. It won't be installed currently, ever, because of this. Bootstrap fails on it.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2591,7 +2591,7 @@ install_centos_stable_deps() {
         if [ "$DISTRO_MAJOR_VERSION" -eq 5 ]; then
             easy_install-2.6 "apache-libcloud>=$_LIBCLOUD_MIN_VERSION"
         else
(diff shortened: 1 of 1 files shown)


s0undt3ch:
Does `pip` also exist on CentOS 5?

The-Loeki:
Not sure. Seeing pkg python-pip is not in EPEL for 5 and bootstrap using easy_install-2.6 to get libcloud when  "$DISTRO_MAJOR_VERSION" -eq 5, I'd venture no though ;) 
That's also the reason why that won't matter for this specific patch though

s0undt3ch:
Forget what I said :)

s0undt3ch:
Thanks!

The-Loeki:
Thank you ;)

PR #545, opened 2015-02-15
Enabled server-optional package for RHEL 7

freimer:
salt-bootstrap.sh fails on RHEL because python-jinja2 was moved to the server-optional repo.  Updated to enable this repo for RHEL 7.


freimer:
Woops, I had 6 in there instead of 7, fixing..

PR #546, opened 2015-02-15
Enable server-optional repo for RHEL for python-jinja2

freimer:
RHEL 7 moved python-jinja2 to server-optional repo. Added a check and enable if on RHEL 7.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2869,6 +2869,9 @@ __test_rhel_optionals_packages() {
         # shellcheck disable=SC2043
         for package in python-jinja2; do
             echoinfo "  - ${package}"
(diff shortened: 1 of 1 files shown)


s0undt3ch:
Are you sure there's no need for user interaction to enable this? Last time I checked there was, hence, the system needed to have it enabled beforehand and we only tested the installation process to confirm...

freimer:
Not sure what interaction you are talking about.  I've used that command in CloudFormation user-data scripts, automated boot scripts, and in this bootstrap.sh script itself.

s0undt3ch:
It was either that or `yum-config-manager` not being available...

Anyway, if it works, merged...

Thanks!

PR #558, opened 2015-03-20
Force LC_ALL=C for chkconfig, since that's LANG dependant

mscherer:
On EL 6 and with LANG=fr_FR.utf8 ( as transmitted by ssh ), the script
fail. It turn out that running salt-cloud remotely with ssh mean that
LANG is transmitted, and chkconfig output is translated. So the grep fail,
and so the deploy fail.

See also https://github.com/saltstack/salt/pull/21858


s0undt3ch:
Thanks!

PR #561, opened 2015-03-25
possible fix for Issue 560

bradthurber:
Ref Issue 560: param to avoid git install on CentOS/RHEL?

s0undt3ch suggested we could detect git binary prior to installation. 

I just wrapped the install with the git check as suggested. I'm not sure about the necessity for "systemd-python" though. This wrapper will also prevent install of that package if git is already installed.

I'm new to all of this so scrutiny appreciated.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2678,16 +2678,16 @@ install_centos_stable_post() {
 
 install_centos_git_deps() {
     install_centos_stable_deps || return 1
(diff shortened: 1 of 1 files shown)


s0undt3ch:
`systemd-python` should always be installed. Could you please do the git check separate from the `systemd-python` install step?
Also, please see the lint issue [here](https://jenkins-testing.saltstack.com/job/bootstrap/job/pr/job/561/job/lint/1/violations/file/bootstrap-salt.sh/)

PR #567, opened 2015-04-01
Fix wrong config paths for FreeBSD (#552)

### LINKED ISSUE #552: Bootstrapping FreeBSD puts config files in wrong location
Given 

`bootstrap.sh -D -G -Z git upstream/2014.7`

Most config files (not the minion_id) are put in `/etc/salt`, but for FreeBSD these belong in `/usr/local/etc/salt`, or they won't be picked up.

The-Loeki:
Fixes #552


s0undt3ch:
Thanks!

The-Loeki:
you welcome as always :+1:

Issue #557, opened 2015-03-20
Wheezy bootstrap error - GPG error:


pitatus:
lastest bootstrap downloaded today, system update performed today ( debian version 7.8 ), attempting to bootstrap results:

```
sh install_salt.sh -P git 2015.2
...
Fetched 1,059 kB in 2s (440 kB/s)
Reading package lists...
W: GPG error: http://http.debian.net wheezy-backports Release: The following signatures were invalid: BADSIG 8B48AD6246925553 Debian Archive Automatic Signing Key (7.0/wheezy) <ftpmaster@debian.org>
Reading package lists...
Building dependency tree...
Reading state information...
python-apt is already the newest version.
The following extra packages will be installed:
  libpgm-5.1-0 python-six python-urllib3
The following NEW packages will be installed:
  libpgm-5.1-0 libzmq3 libzmq3-dev python-requests python-six python-urllib3
  python-zmq
0 upgraded, 7 newly installed, 0 to remove and 30 not upgraded.
Need to get 1,570 kB of archives.
After this operation, 4,286 kB of additional disk space will be used.
WARNING: The following packages cannot be authenticated!
  libzmq3 libzmq3-dev python-six python-urllib3 python-requests python-zmq
E: There are problems and -y was used without --force-yes
 * ERROR: Failed to run install_debian_7_git_deps()!!!
```

erikankrom:
I've run into this multiple times with bootstrapped debian wheezy images.  Workaround is to install the gpg key manually for wheezy backports:

```
gpg --keyserver pgpkeys.mit.edu --recv-key  8B48AD6246925553      
gpg -a --export 8B48AD6246925553 | sudo apt-key add -
```

Would it be possible to add a check to the bootstrap script to grep the output of `apt-key -L`, and add it if missing?

s0undt3ch:
If that solves this issue, I see no problems with the check. Want to submit a PR?

erikankrom:
I'm much more of a sysadmin, and beginner when it comes to git.  I'm a newb when it comes to git pull requests.  Gotta start somewhere though right?

s0undt3ch:
There's no problem if you're not comfortable with this. If you find yourself in trouble, ping me, or, eventually I'll hack it :smile:

s0undt3ch:
Closing as the PR was merged. Thanks!

erikankrom:
[...]

PR #573, opened 2015-04-09
add Debian Archive Automatic Signing Key 

erikankrom:
See #557


erikankrom:
Looks like I could have removed that first edit and reversion.  I'l submit a new PR.

PR #577, opened 2015-04-10
Fix issue where bootstrap script fails if a private epel repo has been configured

### LINKED ISSUE #576: bootstrap script fails if a private epel repo has already been configured and if the server cannot access external network
The bootstrap script should first check if epel repo is already enabled and shows under `yum repolist``before checking if``epel-release`` package is installed. This becomes a problem if the servers in the private network cannot access the networks outside the organization and cannot download a package from public mirrors/repos. If the server has private repos/mirrors already present and enabled (including epel), use the private epel repo. 

Script version: 2015.03.15

Here's the fix:

```
###############################
#
#   CentOS Install Functions
#
__install_epel_repository() {
    if [ ${__EPEL_REPOS_INSTALLED} -eq $BS_TRUE ]; then 
        return 0
    fi

    # Check if epel repo is already enabled and flag it accordingly
    yum repolist | grep -i "epel" > /dev/null 2>&1 
    if [ $? -eq 0 ]; then 
        __EPEL_REPOS_INSTALLED=${BS_TRUE}
        return 0
    fi 

    # Check if epel-release is already installed and flag it accordingly
    rpm --nodigest --nosignature -q epel-release > /dev/null 2>&1
    if [ $? -eq 0 ]; then 
        __EPEL_REPOS_INSTALLED=${BS_TRUE}
        return 0
    fi
```

nmadhok:
Fix issue where bootstrap script fails if a private epel repo has been configured and if the server cannot access external network

Fixes #576

@s0undt3ch

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2608,6 +2608,13 @@ __install_epel_repository() {
         return 0
     fi
 
+    # Check if epel repo is already enabled and flag it accordingly
+    yum repolist | grep -i "epel" > /dev/null 2>&1
+    if [ $? -eq 0 ]; then
+        __EPEL_REPOS_INSTALLED=${BS_TRUE}
+        return 0
+    fi
+
     # Check if epel-release is already installed and flag it accordingly
     rpm --nodigest --nosignature -q epel-release > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


s0undt3ch:
Sorry it took this long.

PR #579, opened 2015-04-13
Fixes #578, set env DEBIAN_FRONTEND=noninteractive

### LINKED ISSUE #578: Salt bootstrap with -U is interactive when an upgradable conf file has been changed
Below are the commands that I'm running for salt bootstrapping our Ubuntu (12 & 14) and Cent (6 & 7) VMs

```
 wget -O install_salt.sh https://bootstrap.saltstack.com --no-check-certificate &&
 sh install_salt.sh -A bdr-salt-master.eng.solidfire.net -P -U -I -F git v2014.7.4
```

Many of these VMs are NOT fresh installs, and have changes to default config files.

On Ubuntu, running with the -U option then requires an interactive confirmation of a conf file change that I've pasted in below

image

I think that the salt bootstrap script should run non-interactively by including the --force-confold or --force-confdef option when using the -U option on Debian-based distros

sparrc:
Some packages ignore the dpkg options --force-confold and --force-confdef (see issue #578 for an example) 

Setting the DEBIAN_FRONTEND to noninteractive guarantees that an interactive prompt won't come up during the package upgrade process.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1704,6 +1704,8 @@ install_ubuntu_deps() {
     if [ $_START_DAEMONS -eq $BS_FALSE ]; then
         echowarn "Not starting daemons on Debian based distributions is not working mostly because starting them is the default behaviour."
     fi
+    # No user interaction, libc6 restart services for example
+    export DEBIAN_FRONTEND=noninteractive
 
     apt-get update
 


sparrc:
Adding `export DEBIAN_FRONTEND=noninteractive` to the start of the install_ubuntu_deps() to mirror install_debian_deps()

s0undt3ch:
Thanks!

PR #589, opened 2015-04-27
Fix Debian Squeeze backports mirror

panticz:
http://http.debian.net/debian-backports is not available anymore

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2040,7 +2040,7 @@ _eof
 
     # Debian Backports
     if [ "$(grep -R 'squeeze-backports' /etc/apt | grep -v "^#")" = "" ]; then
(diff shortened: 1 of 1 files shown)


panticz:
Alternate Debian Squeeze backport mirrors can by found here:
http://backports.debian.org/Mirrors/

s0undt3ch:
We should not "bind" to the German mirror...

Could we switch to `deb http://httpredir.debian.org/debian-backports squeeze-backports main`?

s0undt3ch:
Wow, 404!

s0undt3ch:
Isn't there a main dns entry which will then "redirect" to the closes mirror for Debian?

s0undt3ch:
Thanks!

PR #591, opened 2015-04-29
Added to readme: supported OS SLES 11 SP3.

ghost:
I have tested salt-bootstrap on SUSE Linux Enterprise 11 SP3 and it works.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -332,7 +332,7 @@ Supported Operating Systems
 - Red Hat Enterprise 5/6
 - Scientific Linux 5/6
 - SmartOS
-- SuSE 11 SP1/11 SP2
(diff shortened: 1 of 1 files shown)


s0undt3ch:
Is it OK to assume that enterprise sp1 and sp2 also works? If so, superb!

s0undt3ch:
Nevermind. :smile:

s0undt3ch:
Thanks!

ghost:
Actually, I just noticed in https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L1324 that the script does not support 11 SP1 (contrary to what the README mentioned). We should remove SP1.

PR #606, opened 2015-06-03
Switch to httpredir.debian.org as default Debian mirror

wwentland:
http.debian.net is now hosted on debian.org infrastructure under the name of
httpredir.debian.org which also uses 302 redirects which allows users of Google
DNS servers to use it.

See https://lists.debian.org/debian-devel-announce/2015/05/msg00003.html for
further information.


s0undt3ch:
Thanks!

PR #610, opened 2015-06-23
scmgit package outdated in pkgin, use git instead

BretFisher:
Since SmartOS pkgin 2013Q3 release, `scmgit` is depreciated. We're "strongly advised" to use `git` instead. For now, both yield same result.

More info: https://github.com/joyent/pkgsrc-joyent/blob/master/scmgit/DESCR


jfindlay:
@BretFisher, thanks for your work on this.

Issue #395, opened 2014-05-20
salt-bootstrap does not support Windows guests


marnovdm:
Now that Vagrant supports Windows guests I've been testing the Salt minion on Windows. I've created the following pull request for the Vagrant project which includes a very basic Salt bootstrap script in Powershell: https://github.com/mitchellh/vagrant/pull/3825/files

Is this something the salt-bootstrap project would like to include (and probably greatly improve on) or is salt-bootstrap only meant for Linux deployments?

s0undt3ch:
Sorry @marnovdm for the slow response.

It would be very cool to additionally host a windows powershell script which automated the installation process!!!

Would you be able to maintain it? I can try to take care of small issues that we find in the future but I've never powershell'ed before :smile:

techhat:
+1 on this, @marnovdm! I would love to see this get into Salt Cloud too, in (or instead of) `salt.utils.cloud.deploy_windows()`.

marnovdm:
@s0undt3ch no problem, I've been busy with other things as well. To be honest I'm not really a Windows/Powershell guy either, this is my first Powershell script, was hoping other people could improve it :) Just wanted to get this working so I don't have to manually deploy Windows VMs anymore. 

I've just started playing with salt-cloud as well, once the vSphere provider is supported (Helium I believe) in salt-cloud I could try to integrate this in salt cloud and create a pull reqeust, however I feel like someone more experienced with Powershell would probably do a much better job. Thoughts?

s0undt3ch:
Well, if it's in a working condition, make a PR against the bootstrap script repo, others might pick up on it and develop it even further, hopefully :smile:

marnovdm:
It's definitely in working condition with Vagrant and has been merged, will be released in 1.6.3. I would prepare a pull request for salt-bootstrap as well but don't have a provider supporting Windows in salt-cloud untill the vSphere provider is released for it.

s0undt3ch:
On 27 de Maio de 2014 16:06:20 WEST, Marno notifications@github.com wrote:

> It's definitely in working condition with Vagrant and has been merged,
> will be released in 1.6.3. I would prepare a pull request for
> salt-bootstrap as well but don't have a provider supporting Windows in
> salt-cloud untill the vSphere providers is released for it.
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/saltstack/salt-bootstrap/issues/395#issuecomment-44288685

[...]

PR #618, opened 2015-07-08
Windows bootstrap script

PatOShea:
This would complete #395 
It will install a minion locally on windows. Option parameters will accept version, minion id, master id, and a flag for stopping and setting the salt-minion service to manual for local testing. Finally, help is displayed via:
man ./bootstrap-salt.ps1
help ./bootstrap-salt.ps1 

Passing in a verbose flag will increase logging.
./bootstrap-salt.ps1 -verbose

Finally, I'd like to see the vagrant plugin in mitchellh/vagrant use this too. I would be happy to submit the PR to point mitchellh/vagrant at the file in this location if the SaltStack team thought it was a good idea. 
This would be consistent with how bootstrap-salt.sh is handled currently.
https://github.com/mitchellh/vagrant/blob/master/plugins/provisioners/salt/bootstrap-salt.ps1

### DIFF
--- a/bootstrap-salt.ps1
+++ b/bootstrap-salt.ps1
@@ -1,0 +1,154 @@
+<#
+.SYNOPSIS
+A simple Powershell script to download and install a salt minion on windows.
+.DESCRIPTION
+The script will download the official salt package from saltstack. It will install a specific
+package version and accept parameters for the master and minion ids. Finally, it can stop and
+set the windows service to "manual" for local testing. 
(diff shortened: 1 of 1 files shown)


UtahDave:
@PatOShea thanks for this!  I had forgotten about that open issue for this. 

We'll likely merge your work here with what we're working on in the other location you mentioned.

It definitely makes sense to have a boostrap script here for Windows as well.

@twangboy

Issue #619, opened 2015-07-11
python-jinja2 package moved to optional-releases repo in AWS/EC2 RHUI for RHEL6


lomeroe:
The python-jinja2 package has been moved to the optional-releases repo in AWS/EC2's RHUI for RedHat AWS Marketplace AMIs.

This could be the cause for issue #587 as well.

PR #621, opened 2015-07-11
python-jinja2 has been moved to rhui-...server-releases-optional repo…

lomeroe:
… in EC2's RHUI for RHEL6

Resolves issue #619


s0undt3ch:
I assume you tested this and it actually works right?

lomeroe:
I've tested on RHEL6.6 and RHEL7.1 on both EC2 and GCE

s0undt3ch:
Thanks!

PR #623, opened 2015-07-13
moving install_centos_stable_deps before __test_rhel_optionals_pacakg…

### LINKED ISSUE #622: RHEL attempts to use dependency before checking dependency install
Function install_red_hat_linux_stable_deps calls __test_rhel_optionals_packages, which calls yum-config-manager; however, if yum-utils has not been installed, __test_rhel_optionals_packages will throw errors that yum-config-manager is not available.

yum-utils is verified/installed in the install_centos_stable_deps function, which is called just after the code that calls __test_rhel_optionals_packages.

I only noticed this b/c at this time, yum-utils is not installed by default on a default RHEL7.1 image in GCE.

lomeroe:
…es call in install_red_hat_linux_stable_deps to verify that yum-config-manager is available (it is installed in install_centos_stable_deps) for use (called in __test_rhel_optionals_packages)

resolves #622 

There could be a reason the code was originally ordered this way, I couldn't see a reason while looking at the problem...

Tested on RHEL7.1 and RHEL6.6 to bootstrap new minions

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -3186,6 +3186,8 @@ __test_rhel_optionals_packages() {
 
 
 install_red_hat_linux_stable_deps() {
+    install_centos_stable_deps || return 1
+
     if [ "${DISTRO_MAJOR_VERSION}" -ge 6 ]; then
         # Wait at most 60 seconds for the repository subscriptions to register
         __ATTEMPTS=6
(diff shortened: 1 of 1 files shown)


s0undt3ch:
The order kind of matters, here's a better fix https://github.com/saltstack/salt-bootstrap/commit/f63c8fb2826d1324a3377dbc9445d1ca7c76a75f

Still, Thank You for looking into this.

lomeroe:
Thanks @s0undt3ch

PR #624, opened 2015-07-16
fix config and etc path on SmartOS

### LINKED ISSUE #615: SmartOS base-64 15.1.1    /etc/salt should be /opt/local/etc/salt
#### SMF manifest installed by salt-bootstrap specifies correct directory

```
# svccfg export salt-minion | grep name=\'config_dir\'
      <propval name='config_dir' type='astring' value='/opt/local/etc/salt'/>
```
#### Salt configuration at /etc/salt but not /opt/local/etc/salt

```
# ls -ld /etc/salt /opt/local/etc/salt
ls: cannot access /opt/local/etc/salt: No such file or directory
drwxr-xr-x 3 root root 5 Jun 26 08:36 /etc/salt
```
#### temporary work-around

```
ln -s /etc/salt /opt/local/etc/salt
```

BretFisher:
Issue: On SmartOS, if BS_SALT_ETC_DIR is not given, `bootstrap-salt.sh` [uses wrong etc path](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L194)

Fixes https://github.com/saltstack/salt-bootstrap/issues/615

Background:
Salt-bootstrap hardcodes all OS's to /etc/salt unless they have their own manually later in the script (like freebsd), or the BS_SALT_ETC_DIR environment variable is given at runtime. SmartOS uses /opt/local/etc so lets set that to default for it.

This is killing two issues on SmartOS that this script presents with default etc path.
1. First is we just need to set the proper default [like freebsd does.](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L3976)
2. Next, We've reciently fixed salt's syspaths.py for SmartOS since 2015.5.0, but that doesn't help anyone using this bootstrap to install 2014.x etc. which will still put parts of salt in the wrong place, so unlike how [fedora falls back to syspath.py](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L2677-L2684) I think we should still prefer our new default unless we have a environment variable override.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4229,7 +4229,7 @@ install_smartos_stable() {
 install_smartos_git() {
     # Use setuptools in order to also install dependencies
     # lets force our config path on the setup for now, since salt/syspaths.py only  got fixed in 2015.5.0
-    USE_SETUPTOOLS=1 /opt/local/bin/python setup.py install --salt-config-dir="$_SALT_ETC_DIR" || return 1
++++ USE_SETUPTOOLS=1 /opt/local/bin/python setup.py install --salt-config-dir="$_SALT_ETC_DIR" || USE_SETUPTOOLS=1 /opt/local/bin/python setup.py --salt-config-dir="$_SALT_ETC_DIR" install || return 1
     return 0
 }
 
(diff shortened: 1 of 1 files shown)


s0undt3ch:
Please make the following change:

``` diff
--- USE_SETUPTOOLS=1 /opt/local/bin/python setup.py install --salt-config-dir="$_SALT_ETC_DIR" || return 1
+++ USE_SETUPTOOLS=1 /opt/local/bin/python setup.py install --salt-config-dir="$_SALT_ETC_DIR" || USE_SETUPTOOLS=1 /opt/local/bin/python setup.py --salt-config-dir="$_SALT_ETC_DIR" install || return 1
```

Since the `--salt-config-dir` option became a global one but we still need to try both.

BretFisher:
done.

s0undt3ch:
Thanks!

PR #634, opened 2015-07-23
bugfix: exit git root before removing it

### LINKED ISSUE #613: On SmartOS, /tmp/git/salt isn't removed after install
As mentioned in issue #473 and in my own use across dozens of SmartOS servers, after running salt-bootstrap.sh the `/tmp/git/salt` directory is empty, but remains. On next run of salt-bootstrap.sh, it will fail as it can't clone to an existing folder.

Here's where I focused my troubleshooting:

``` bash
__exit_cleanup() {
    EXIT_CODE=$?

    if [ "$ITYPE" = "git" ] && [ -d "${__SALT_GIT_CHECKOUT_DIR}" ]; then
        if [ $_KEEP_TEMP_FILES -eq $BS_FALSE ]; then
            # Clean up the checked out repository
            echodebug "Cleaning up the Salt Temporary Git Repository"
            rm -rf "${__SALT_GIT_CHECKOUT_DIR}"
        else
```

https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L496

On using `ll --full-time /tmp/git` during and after install, the dir has different dates, which to me means it's actually getting recreated even after `__exit_cleanup()` runs. I put some logging in for every `mkdir` and even tried -D but couldn't see where it would recreate it.
### Options
- Someone with more script-fu find where it's getting recreated
- We add a pre-clone check to see if `salt` dir is there and if so delete it so we don't fail install

Happy to submit PR if someone wants to recommend a course of action.

BretFisher:
Fixes #613

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -504,6 +504,7 @@ __exit_cleanup() {
         if [ $_KEEP_TEMP_FILES -eq $BS_FALSE ]; then
             # Clean up the checked out repository
             echodebug "Cleaning up the Salt Temporary Git Repository"
+            cd "${__SALT_GIT_CHECKOUT_PARENT_DIR}"
             rm -rf "${__SALT_GIT_CHECKOUT_DIR}"
         else
             echowarn "Not cleaning up the Salt Temporary git repository on request"


jfindlay:
Nice, thanks @BretFisher.

PR #653, opened 2015-09-05
Make bootstrap work with FreeBSD 11-CURRENT.

cbuechler:
This change makes bootstrap work with FreeBSD 11-CURRENT.


s0undt3ch:
Thank You!

PR #654, opened 2015-09-09
Fixes error finding python-jinja2 in RHEL 7

hedinfaok:
The installer prints the following on RHEL 7 when the server optional package is not enabled:

```
 *  INFO: Testing if packages usually on the optionals repository are available:
 *  INFO:   - python-jinja2
 * ERROR: Failed to find an installable 'python-jinja2' package. The optional repository or its subscription might be missing.
 *  INFO: Sleeping 10 seconds while waiting for the optional repository subscription to be externally configured
```

This change fixes the name of the server-optional repo to be the correct name.

The command `sudo subscription-manager repos --list` displays all available repos and their names.


s0undt3ch:
Thanks!

PR #660, opened 2015-09-22
Add -f option to force shallow cloning

eyj:
We're currently deploying some instances from our own salt repo with quick fixes and feature cherrypicks, using a deploy branch on that repo to control which version should be deployed.

The -f flag forces the git clone to use shallow cloning, accepting that the NOC part of the version number is "n/a" (see https://github.com/saltstack/salt/issues/14855 ).

It is probably also interesting for other users which want to deploy bleeding edge or self-modified salt versions at medium scale.


jfindlay:
Thanks, @eyj.

PR #662, opened 2015-09-29
RedHat based distros now point to SaltStack corporate repo. 

grep4linux:
Added SLE 12 support. Fixed OpenSUSE.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1,4 +1,4 @@
-#!/bin/sh -
+h!/bin/sh -
 #======================================================================================================================
 # vim: softtabstop=4 shiftwidth=4 expandtab fenc=utf-8 spell spelllang=en cc=120
 #======================================================================================================================
@@ -862,6 +862,10 @@ __gather_linux_system_info() {
                     debian      )
                         n="Debian"
                         v=$(__derive_debian_numeric_version "$v")
+                        ;;
+                    sles        )
+                        n="SUSE"
+                        v="${rv}"
                         ;;
                     *           )
                         n=${nn}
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh -
+h!/bin/sh -
```
s0undt3ch, line 1: h?

s0undt3ch:
I need to review the rest locally since I already have the new repo stuff in a branch on my fork.

bootstrap-salt.sh
```diff
@@ -4457,6 +4469,9 @@ install_opensuse_stable_deps() {
     # shellcheck disable=SC2086
     __zypper_install ${__PACKAGES} || return 1
 
+    # Fix for OpenSUSE 13.2 and 2015.8 - gcc should not be required. Work around until package is fixed by SuSE
+    _EXTRA_PACKAGES="${_EXTRA_PACKAGES} gcc python-devel libgit2-devel"
```
s0undt3ch, line 4473: Why do we need these? Specially pygit2 which is not a hard dependency for Salt.

What type of installation triggered this? Stable packages? Git based installation?

bootstrap-salt.sh
```diff
@@ -4624,8 +4639,200 @@ install_opensuse_check_services() {
...
+    return 0
+}
+
+install_suse_12_stable() {
+    if [ "$SUSE_PATCHLEVEL" -gt 1 ]; then
+        install_opensuse_stable || return 1
+    else
+        # USE_SETUPTOOLS=1 To work around
+        # error: option --single-version-externally-managed not recognized
+        USE_SETUPTOOLS=1 pip install salt || return 1
```
s0undt3ch, line 4770: The openSUSE stable packages don't work in SuSE? Why are we pip installing here?

bootstrap-salt.sh
grep4linux, line 1: oops! Sorry about that. keep the normal shebang.

bootstrap-salt.sh
grep4linux, line 4473: We looked at this with Tom. He said that it looked like the SUSE salt-minion rpm was requiring git event though we don't need it.  I'll have to re-create the problem, but one of the dependencies, (I think git 2) is being required by SUSE. However, it has a dependency that requires a compiler. My plan is to look into it more, submit a bug to SUSE. I've met the packager Olaf when I worked there and wanted to send him a note.  

All I know at this point is that the only way it works is with those packages added. This is just a standard rpm-based installation.

jfindlay:
@grep4linux, @s0undt3ch, any updates here?

clarkperkins:
Would be nice if this could get merged, as `bootstrap-salt.sh` does not currently work with CentOS 6 due to the bad version of python-crypto.

jfindlay:
@clarkperkins, try https://bootstrap.saltstack.com/develop.

vutny:
@grep4linux Hey, looks like your branch is outdated and requires some rebase/rework.
Is this still an issue after all? As I see it, bootstrapping on Suse has nothing to do with SaltStack corporate repo.

jfindlay:
@vutny, I have talked with @grep4linux and am working on making the changes you suggest and some others.

vutny:
Thank you @jfindlay. Just wanted to say that looks like this PR is no longer relevant and could be closed.

PR #663, opened 2015-10-01
Install stable version with the insecure example

alexjurkiewicz:
For anyone who does copy & paste the insecure one-liner, they probably should get the stable version rather than development.


s0undt3ch:
The comment above your change mentions the latest develop.

My suggestion would be to include your stable example above the latest develop comment, with a comment referring something like, to install latest stable..... and then the develop comment and example would come bellow....

PR #665, opened 2015-10-01
Change to 'dnf' as package manager for Fedora 22->

mbologna:
If you are installing Saltstack on Fedora 22 onwards, the default package manager
is dnf (instead of yum).
This patch decide which package manager to use based on Fedora version and the
availability of dnf on the current system. If found, uses dnf to install packages
needed by saltstack.

Using this patch final user will not see yum deprecation warnings like:
Yum command has been deprecated, redirecting to '/usr/bin/dnf install -y yum-utils..
See 'man dnf' and 'man yum2dnf' for more information.
To transfer transaction metadata from yum to DNF, run:
'dnf install python-dnf-plugins-extras-migrate && dnf-2 migrate'

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2629,14 +2629,14 @@ install_debian_check_services() {
 
 FEDORA_PACKAGE_MANAGER="yum"
 
-fedora_get_package_manager() {
+__fedora_get_package_manager() {
   if [ "$DISTRO_MAJOR_VERSION" -lt 22 ] || [ "$(which dnf)" != "" ]; then
     FEDORA_PACKAGE_MANAGER="dnf"
   fi
 }
 
 install_fedora_deps() {
-    fedora_get_package_manager
+    __fedora_get_package_manager
     if [ "$_ENABLE_EXTERNAL_ZMQ_REPOS" -eq $BS_TRUE ]; then
         __install_saltstack_copr_zeromq_repository || return 1
     fi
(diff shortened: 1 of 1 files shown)


s0undt3ch:
What happened to the symlinks?

mbologna:
@s0undt3ch you're right. Corrected and re-push'ed -f

bootstrap-salt.sh
```diff
@@ -2626,7 +2626,17 @@ install_debian_check_services() {
 #
 #   Fedora Install Functions
 #
+
+FEDORA_PACKAGE_MANAGER="yum"
+
+fedora_get_package_manager() {
```
jfindlay: @mbologna, would you mind renaming this to `__fedora_get_package_manager`?  Thanks.

mbologna: @jfindlay done in 10bcdf3

bootstrap-salt.sh
```diff
@@ -2626,7 +2626,17 @@ install_debian_check_services() {
 #
 #   Fedora Install Functions
 #
+
+FEDORA_PACKAGE_MANAGER="yum"
+
+__fedora_get_package_manager() {
+  if [ "$DISTRO_MAJOR_VERSION" -lt 22 ] || [ "$(which dnf)" != "" ]; then
```
jfindlay: @mbologna, sorry for another comment, but shouldn't this line be
`if [ "$DISTRO_MAJOR_VERSION" -ge 22 ] || [ "$(which dnf)" != "" ]; then`

mbologna: Yep, you're right. Correcting it right now and pushing.

jfindlay:
Thanks, @mbologna.

Issue #539, opened 2015-01-31
Bootstrap fails on Raspbian


mventimi:
I cannot install using salt-bootstrap on a Raspbian system:

```
pi@raspberrypi ~ $ sudo sh install_salt.sh -D
 *  INFO: PIPED THROUGH install_salt.sh -- Version 2015.01.12

 *  INFO: System Information:
 *  INFO:   CPU:
 *  INFO:   CPU Arch:     armv6l
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   3.12.28+
 *  INFO:   Distribution: Raspbian 7.6

 * DEBUG: Binaries will be searched using the following $PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 *  INFO: Installing minion
 * DEBUG: install_raspbian_7_stable_deps not found....
 * DEBUG: install_raspbian_7_6_stable_deps not found....
 * DEBUG: install_raspbian_7_deps not found....
 * DEBUG: install_raspbian_7_6_deps not found....
 * DEBUG: install_raspbian_stable_deps not found....
 * DEBUG: install_raspbian_deps not found....
 * DEBUG: DEPS_INSTALL_FUNC=null
     ...
 * ERROR: No dependencies installation function found. Exiting...
 * DEBUG: Removing the logging pipe /tmp/bootstrap-salt.logpipe
```

I updated `__gather_linux_system_info()` with the following code:

``` python
        elif [ "${DISTRO_NAME}" = "Arch" ]; then
            DISTRO_NAME="Arch Linux"
            return
+       elif [ "${DISTRO_NAME}" = "Raspbian" ]; then
+           DISTRO_NAME="Debian"
+       fi
        rv=$(lsb_release -sr)
```

Running the updated script seems to fix the problem:

```
pi@raspberrypi ~ $ sudo sh install_salt.sh -D
 *  INFO: PIPED THROUGH install_salt.sh -- Version 2015.01.12

 *  INFO: System Information:
 *  INFO:   CPU:
 *  INFO:   CPU Arch:     armv6l
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   3.12.28+
 *  INFO:   Distribution: Debian 7.6

 * DEBUG: Binaries will be searched using the following $PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 *  INFO: Installing minion
[...]

PR #670, opened 2015-10-12
Update bootstrap-salt.sh

virtualguy:
Resolves: Bootstrap fails on Raspbian #539 

Add alias of Raspbian to match the debian family. Tested working on Raspbian jessie (released 2015-09-24)


jfindlay:
Thanks, @virtualguy.

Issue #668, opened 2015-10-05
CentOS 6 points to old repo


rallytime:
Similar to #659, the bootstrap script points to the [COPR repo](https://copr.fedoraproject.org/coprs/saltstack/salt/) instead of the new updated address http://repo.saltstack.com. This means that 2015.5.2 is getting installed for CentOS 6 (and likely other versions as well) instead of 2015.8.0 or newer.

jfindlay:
This should be fixed by #674.

s0undt3ch:
#674 was merged. Closing. Please re-open if necessary.

PR #671, opened 2015-10-12
Migrate from COPR repo to Saltstack Repo

timwsuqld:
Fix for Issue #668
We no longer need COPR repository now that Saltstack have offical repo's.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2627,11 +2627,7 @@ install_debian_check_services() {
 #   Fedora Install Functions
 #
 install_fedora_deps() {
-    if [ "$_ENABLE_EXTERNAL_ZMQ_REPOS" -eq $BS_TRUE ]; then
-        __install_saltstack_copr_zeromq_repository || return 1
-    fi
-
-    __install_saltstack_copr_salt_repository || return 1
+    __install_saltstack_repository || return 1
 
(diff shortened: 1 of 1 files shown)


rallytime:
@timwsuqld Thanks for this fix! Looks good to me. @s0undt3ch Can you review this when you get a moment?

jfindlay:
@timwsuqld, this is great, although I'm loth to remove copr support (Will we ever need it again in the future?).

timwsuqld:
@jfindlay One way we look at this kind of thing at work, is it's in git, so not hard to got back and get the few functions back. Happy to modify this to include the copr repo, but they'll be functions that aren't being called by anything, so may end up not working as no one will be using them, so they won't be getting tested.

jfindlay:
@timwsuqld, yes, you are right, sorry for the dithering.

jfindlay:
@timwsuqld, since fedora packages have not yet been integrated into repo.saltstack.com, would you mind reworking this so that fedora still has access to the COPR repo?  If you don't have time for this, then I can work on it.  Thanks.

timwsuqld:
@jfindlay Before I try and work this more, am I correct in assuming that https://github.com/saltstack/salt-bootstrap/pull/674 brings in these changes anyway and we can now close this?

jfindlay:
@timwsuqld, that is correct.

timwsuqld:
Closing as #674 fixes this

PR #685, opened 2015-11-04
Add support for pegged versions on YUM based OS'ses through repo.saltstack.com

The-Loeki:
This patch adds major version pegging support for YUM based OS'es by leveraging Salt's new repo layout.
See the discussion for further details.

~~This admittedly hacky patch (blame your own repo infrastructure for that ;) ) adds version pegging support to repo.saltstack.com repos for any YUM based OS that properly supports glob matching in `includepkgs/exclude=`.~~
~~I've only tested it on CentOS 7, will probably test CentOS 6/7 in the very near future if this gets committed, but the `includepkgs/exclude=` syntax seems universal.~~

~~It does this by adding~~

```
includepkgs=[0-9]:[!s][!a][!l][!t]* salt*${STABLE_REV}*
```

~~to the repo definition if `STABLE_REV` isn't `latest`.
(In case you're wondering, the `[0-9]:` is because YUM also matches on a string with an 'epoch' in it)~~

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1875,7 +1875,7 @@ install_ubuntu_stable_deps() {
 
         # Saltstack's Stable Ubuntu repository
         if [ "$(grep -ER 'latest .+ main' /etc/apt)" = "" ]; then
-            echo "deb http://repo.saltstack.com/apt/ubuntu/$DISTRO_VERSION/$CPU_ARCH/$STABLE_REV $DISTRO_CODENAME main" >> \
+            echo "deb http://repo.saltstack.com/apt/ubuntu/$DISTRO_VERSION/$CPU_ARCH_L/$STABLE_REV $DISTRO_CODENAME main" >> \
                 /etc/apt/sources.list.d/saltstack.list
         fi
 
@@ -1884,7 +1884,7 @@ install_ubuntu_stable_deps() {
         __apt_get_install_noinput wget
 
         # shellcheck disable=SC2086
-        wget $_WGET_ARGS -q https://repo.saltstack.com/apt/ubuntu/$DISTRO_VERSION/$CPU_ARCH/$STABLE_REV/SALTSTACK-GPG-KEY.pub -O - | apt-key add - || return 1
+        wget $_WGET_ARGS -q https://repo.saltstack.com/apt/ubuntu/$DISTRO_VERSION/$CPU_ARCH_L/$STABLE_REV/SALTSTACK-GPG-KEY.pub -O - | apt-key add - || return 1
 
     else
         # Alternate PPAs: salt16, salt17, salt2014-1, salt2014-7
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2942,7 +2942,7 @@ enabled_metadata=1
 _eof
         if [ "$ITYPE" = "stable" -a "$STABLE_REV" != "latest" ]; then
             cat << _eof >> "/etc/yum.repos.d/repo-saltstack-el${DISTRO_MAJOR_VERSION}.repo"
-includepkgs=[0-9]:[!s][!a][!l][!t]* salt*${STABLE_REV}*
+includepkgs=?:[!s][!a][!l][!t]* salt*${STABLE_REV}*
 _eof
         fi
 
@@ -2967,7 +2967,7 @@ enabled_metadata=1
 _eof
         if [ "$ITYPE" = "stable" -a "$STABLE_REV" != "latest" ]; then
             cat << _eof >> "/etc/yum.repos.d/repo-saltstack-el${DISTRO_MAJOR_VERSION}.repo"
-includepkgs=[0-9]:[!s][!a][!l][!t]* salt*${STABLE_REV}*
+includepkgs=?:[!s][!a][!l][!t]* salt*${STABLE_REV}*
 _eof
         fi
 
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -3168,7 +3168,7 @@ install_centos_git() {
         _PYEXE=python2
     fi
     if [ -f "${__SALT_GIT_CHECKOUT_DIR}/salt/syspaths.py" ]; then
-        $_PYEXE setup.py --prefix=/usr --salt-config-dir="$_SALT_ETC_DIR" install || return 1
+        $_PYEXE setup.py install --prefix=/usr --salt-config-dir="$_SALT_ETC_DIR" || return 1
     else
         $_PYEXE setup.py install --prefix=/usr || return 1
     fi
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2925,10 +2925,12 @@ __install_saltstack_copr_zeromq_repository() {
 
 __install_saltstack_rhel_repository() {
     base_url="https://repo.saltstack.com/yum/redhat/\$releasever/\$basearch/${STABLE_REV}/"
+    fetch_url="https://repo.saltstack.com/yum/redhat/${DISTRO_MAJOR_VERSION}/${CPU_ARCH_L}/"
+
     if [ "${DISTRO_MAJOR_VERSION}" -eq 5 ]; then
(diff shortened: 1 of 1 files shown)


The-Loeki:
now tested & verified on CentOS6 as well

vutny:
@The-Loeki As for me, version pegging is not very useful for https://repo.saltstack.com/yum/ since repository is holding only latest released version of stable packages. Also, there are a lot of required packages, like python libs, so including just packages beginning with `salt` looks as not good idea.

But I definitely like to download GPG key with https. Maybe you cosider to revert your changes for RHEL-based system and leave the fix only for Ubuntu?

The-Loeki:
@vutny version pegging seems kinda mandatory IMHO; officially salt minions & salt masters SHOULD match versions and minions MUST NOT run @ higher major versions than salt masters.

Therefore to prevent breakage you need to be able to 'pin' on a major release. Since the repo's do not reflect that, you need to peg the major version.

Also, note that the includepkgs line translates to 'anything NOT named salt AND everything salt-VERSION'

dmurphy18:
@vutny @The-Loeki You are actually able to pin on a release in the salt repo's for example, Redhat:
baseurl:=http://repo.saltstack.com/yum/redhat/7/x86_64/archive/2015.8.4

there are similar abilities for Debian  and Ubuntu
The archive locations are not subject to change.

Use of yum/redhat/version/arch is the preferred method to use when using the repo,
yum/rhel7, yum/rhel6 and yum/rhel5 are only there to preserve backwards compatibility with the earliest versions of repo.saltstack.com.

The-Loeki:
@dmurphy18 thanks for that tip, hadn't noticed that yet :) I'll modify my PR ASAP

The-Loeki:
@dmurphy18 There you go :) Looking further through the (fast improving, kudos!) layout of the repository, I noticed there's now repo rpm's, which IMHO is the preferred way of doing things; the installation could simply be shortened to installing that one RPM.

However the repo RPM's are _not_ pegging yet, always resorting to 'latest' rather than the chosen major.

dmurphy18:
@The-Loeki the install repo are designed to install the latest based on OS release and architecture for the particular branch of salt, e.g. 2015.8, otherwise they would have to be adjusted every time a new release release occurs.  if pinning is desired, then the user can adjust the installed repo file for the particular archived version of salt.

The-Loeki:
All better now :) ready for merge again 

@dmurphy18 that's exactly what I mean; the repo RPM's have 'latest' in their URL as opposed to '2015.8/2015.5' etc

dmurphy18:
@The-Loeki the install repo rpm currently on the site is specific for branch 2015.8. There is another for 2015.5 but it has not had full time to go through QA process, but hopefully it will be available soon.  The 'latest' is by design, not error and is similar to that used in AWS for OS release version.  It was decided that providing a baseurl that points to the latest version of salt in the repo for a specific branch was the preferred method to use.

The-Loeki:
@dmurphy18 while I can understand the desire for a 'latest' repo which just refers to the latest & greatest release of salt, the point there is that you'd have a `salt-repo-latest.rpm` & a `salt-repo-2015.8.rpm` and a `salt-repo-2015.5.rpm` etc.

I would be rather unpleasantly suprised if the 2015.8 rpm actually turns out to be 'latest' and dozens if not hundreds of minions start upgrading their packages to Boron during their maintenance window and suddenly are not able to connect to a Beryllium master anymore, wreak havoc on modules that aren't compatible with the new codebase etc. ;) 
Again, Salt's own documentation specifically outlines these risks and strongly recommends against newer major version minions connecting to older major masters.

That kind of otherwise unavoidable breakage is IMHO the very reason you'd want to peg to a specific major version.

jfindlay:
Also see similar discussion at https://github.com/saltstack/salt/issues/27385.

vutny:
@The-Loeki Now I understood your concern. The updated PR looks really good as for me. I even would like to test it. Only one line I found suspicious (2952):

```
enabled_metadata=1gpg
```

Accidental paste?

bootstrap-salt.sh
```diff
@@ -2923,40 +2923,36 @@ __install_saltstack_copr_zeromq_repository() {
...
+[salt-${repo_rev}]
+name=SaltStack ${repo_rev} Release Channel for RHEL/CentOS \$releasever
+baseurl=$base_url
 skip_if_unavailable=True
 gpgcheck=1
-gpgkey=https://repo.saltstack.com/yum/rhel${DISTRO_MAJOR_VERSION}/SALTSTACK-GPG-KEY.pub
+gpgkey=${base_url}${gpg_key}
 enabled=1
-enabled_metadata=1
+enabled_metadata=1gpg
```
vutny: Strange... Accidental paste?

bootstrap-salt.sh
```diff
@@ -1884,7 +1884,7 @@ install_ubuntu_stable_deps() {
         __apt_get_install_noinput wget
 
         # shellcheck disable=SC2086
-        wget $_WGET_ARGS -q http://repo.saltstack.com/apt/ubuntu/ubuntu$DISTRO_MAJOR_VERSION/$STABLE_REV/SALTSTACK-GPG-KEY.pub -O - | apt-key add - || return 1
+        wget $_WGET_ARGS -q https://repo.saltstack.com/apt/ubuntu/ubuntu$DISTRO_MAJOR_VERSION/$STABLE_REV/SALTSTACK-GPG-KEY.pub -O - | apt-key add - || return 1
```
jfindlay: @The-Loeki, if you don't mind, this should be `https://repo.saltstack.com/apt/ubuntu/$DISTRO_VERSION/$CPU_ARCH/$STABLE_REV/SALTSTACK-GPG-KEY.pub`.

bootstrap-salt.sh
The-Loeki: I guess :) tnx for noticing :D

The-Loeki:
@jfindlay  all fixed methinks ;)

vutny:
Hey @The-Loeki
How about to add the pegging versions support for Debian too? ;-)

Related repos URL: http://repo.saltstack.com/apt/debian/8/amd64/

The-Loeki:
@vutny once this one is merged I'll have a looksy, but considering the Ubuntu installs already support it, I think the Debians will or should easily be able to do it as well.

vutny:
@The-Loeki One more thing :-)
Please update the usage heredoc with supported platforms, like:

```
  Installation types:
    - stable (default)
    - stable [version] (supported on Ubuntu and RHEL-family distributions)
```

By the way, repo.saltstack.com doesn't contain packages for fedora...

bootstrap-salt.sh
```diff
@@ -1210,7 +1210,7 @@ __ubuntu_codename_translation
 if ([ "${DISTRO_NAME_L}" != "ubuntu" ] && [ "$ITYPE" = "daily" ]); then
     echoerror "${DISTRO_NAME} does not have daily packages support"
     exit 1
-elif ([ "${DISTRO_NAME_L}" != "ubuntu" ] && [ "$ITYPE" = "stable" ] && [ "$STABLE_REV" != "latest" ]); then
+elif ([ "$(echo "${DISTRO_NAME_L}" | egrep '(ubuntu|centos|red_hat|scientific|oracle|amazon|fedora)')" = "" ] && [ "$ITYPE" = "stable" ] && [ "$STABLE_REV" != "latest" ]); then
```
vutny, line 1213: I guess we should remove 'fedora' from here.

The-Loeki:
@vutny The way I see it the script needs a number of additional modifications (like the Debian repo's already mentioned somewhere here, removing older and unsupported cruft etc.) anyway to bring it up to speed with the new (and still evolving) situation, and I'll be happy to continue to help once all my PR's here are merged; there's only so many PR's/commits I like to keep in flight, especially on a single script.

For now I just keep updating this one with the issues we find.

jfindlay:
Yes, let's get this change merged and then we can look at parallel updates for other platforms and configurations.  @The-Loeki has been handling the extensive feedback like a boss. :-)

PR #692, opened 2015-11-18
Process -s (default sleep for service restarts) in bootstrap-salt.sh

hipikat:
An -s option for the 'default sleep' time in bootstrap-salt.sh is documented in the usage instructions, but not implemented. This commit lets bootstrap-salt.sh recognise and use the -s option.

Originally at saltstack/salt#28938, where I was advised to come to saltstack/salt-bootstrap instead. :)

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -291,7 +291,7 @@ EOT
 }   # ----------  end of function usage  ----------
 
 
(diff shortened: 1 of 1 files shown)


vutny:
:+1: for this. @jfindlay Could you be so kind to merge it?

jfindlay:
Yes.  Thanks @hipikat for fixing this omission.

PR #699, opened 2015-12-04
Update bootstrap-salt.sh

wolfpackmars2:
Some cleanup for Debian Deps, especially for Debian 8 where some Dep packages may not have been installed


s0undt3ch:
Thanks!

PR #717, opened 2015-12-21
Better EPEL repository detection on RHEL and CentOS

vutny:
Hi,

I had a problem with bootstrapping some CentOS machines were was pre-configured YUM repository which name contains "epel" substring, but that's not main Fedora's EPEL.

In such case script failed with very uninformative error:

```
         *  INFO: Running install_centos_stable_deps()
        Loaded plugins: fastestmirror


        Error getting repository data for epel, repository not found
         * ERROR: Failed to run install_centos_stable_deps()!!!
```

Providing more specific regex for grep fixes it.

P.S. I know this is kind of edge case, but somebody may fall into it as well.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2873,7 +2873,7 @@ __install_epel_repository() {
     fi
 
     # Check if epel repo is already enabled and flag it accordingly
-    yum repolist | grep -i "epel" > /dev/null 2>&1
+    yum repolist | grep -i '^epel/' > /dev/null 2>&1
     if [ $? -eq 0 ]; then
         __EPEL_REPOS_INSTALLED=${BS_TRUE}
         return 0
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2873,7 +2873,7 @@ __install_epel_repository() {
     fi
 
     # Check if epel repo is already enabled and flag it accordingly
-    yum repolist | grep -i '^epel/' > /dev/null 2>&1
+    yum repolist | grep -q "^[!]${_EPEL_REPO}/"
     if [ $? -eq 0 ]; then
         __EPEL_REPOS_INSTALLED=${BS_TRUE}
         return 0


jfindlay:
@vutny, thanks for the pull request.  When I do `yum repolist` I get `!epel/x86_64        Extra Packages for Enterprise Linux 7 - x86_64       8,918`, which doesn't look like it will be matched by your modified regex.

``` console
centos-7-main ~ master # yum repolist | grep -i '^epel/'
centos-7-main ~ master # yum repolist | grep -i 'epel/'
!epel/x86_64          Extra Packages for Enterprise Linux 7 - x86_64      8,918
centos-7-main ~ master #
```

vutny:
Yeah, I got it also. (!) Exclamation mark from `yum repolist` command output means that repository metadata was expired.
By the way, this is the only one place in the script where hard-coded `epel` string is used. Everywhere else we have something similar to this command:

```
yum -y install ${package} --enablerepo=${_EPEL_REPO}
```

And this caused the error message for me before:

```
Error getting repository data for epel, repository not found
```

if the repo with exact name from `$_EPEL_REPO` variable was not configured. Another case, I've noticed that repo name would be `epel-testing` when I choose `testing` installation type on bootstrap.
All those cases are covered now.

The final improvement in my last commit is suppression of `grep` output with 'quiet' option instead of `/dev/null` redirection. It just looks more clear.

Thanks.

jfindlay:
This is excellent, thanks, @vutny!

sp1r:
Hello,

There is an issue with current grep pattern. Exclamation mark is optional in `yum repolist` output, and string without it does not match.

```
# yum repolist | grep '^[!]epel/'
# yum repolist  
...  
epel/7Server/x86_64                                         Extra Packages for RHEL Linux 7Server - x86_64                                     10,580
...
```

vutny:
Ooops! Indeed, this should be 

``` sh
yum repolist | grep -q '^!\?'"${_EPEL_REPO}/"
```

Have I really wrote that? :smile: 
@sp1r Feel free to submit a PR to fix it!

sp1r:
@vutny Thank you. I will be glad to join the contributors ;)

PR #720, opened 2015-12-24
Remove RHEL optional repo check and enable

### LINKED ISSUE #669: Bootstrap script requires python-jinja2 before installing Salt COPR
When using the bootstrap script to install onto a RHEL6 box, the script tries to verify that python-jinja2 is available. Normally, this is in the RHEL Optionals channel, but the Salt COPR repo now has python-jinja2 in it.

Seems it would make more sense to configure the Salt COPR yum repo before attempting to check if python-jinja2 is available to cut back on the amount of configuration needed for the host to actually bootstrap salt.

I've confirmed it works by manually configuring the Salt COPR repo, then attempting the bootstrap. The python-jinja2 is detected in the repo. It then goes onto fail on python-urllib3 missing ( which isn't part of the Salt COPR, EPEL, the RHEL media or the RHEL optional channels, but is only available via the RHEL updates channel.. )

It'd also be nice to see all the RPM requirements needed for a 'bootstrap' in the Salt COPR eventually, if we're trying to get to a true 'bootstrap', rather than a 'do a bunch of other configuration, and then bootstrap'.

vutny:
All necessary packages are installed from Salt repo now. Fixes #669.
Verified on all RHEL systems:
- https://gist.github.com/vutny/d710993c13bca9f53fba
- https://gist.github.com/vutny/d413aeda143742103843
- https://gist.github.com/vutny/54a2d09b2b8f8f8b6fd5

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -3314,95 +3314,12 @@ install_centos_check_services() {
 #
 #   RedHat Install Functions
 #
-__test_rhel_optionals_packages() {
-    __install_epel_repository || return 1
-
-    # Make sure yum-utils is installed
-    yum list installed yum-utils > /dev/null 2>&1 || yum -y install yum-utils --enablerepo=${_EPEL_REPO} || return 1
-
-    if [ "$DISTRO_MAJOR_VERSION" -ge 7 ]; then
-        yum-config-manager --enable \*server-optional-rpms || return 1
-    fi
-
-    if [ "$DISTRO_MAJOR_VERSION" -ge 6 ]; then
-        #python-jinja2 is in repo server-releases-optional in EC2/RHEL6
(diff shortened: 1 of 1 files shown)


jfindlay:
@vutny, thanks for your work on this.  I would like to test this further before merging.

vutny:
@jfindlay Sure, I've tested this on official Red Hat AMIs from AWS Marketplace.

vutny:
@jfindlay Any updates on this? I did many bootstraps on RHEL since then, and everything's OK.
How about to merge my PR? :-) Thanks!

jfindlay:
I see that now we are providing all needed deps on repo.saltstack.com, there is no need for RHEL Optional packages.  Thanks, @vutny.

PR #721, opened 2015-12-25
Remove SaltStack COPR repository configuration for CentOS/RHEL5

vutny:
The function `__install_saltstack_copr_salt_el5_repository` is not referenced anywhere else in the script, and COPR became deprecated for RHEL/CentOS recently.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2924,14 +2924,6 @@ __install_saltstack_copr_zeromq_repository() {
     return 0
 }
 
-__install_saltstack_copr_salt_el5_repository() {
-    if [ ! -s /etc/yum.repos.d/saltstack-salt-el5-epel-5.repo ]; then
-        __fetch_url /etc/yum.repos.d/saltstack-salt-el5-epel-5.repo \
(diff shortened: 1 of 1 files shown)


jfindlay:
@vutny, I would like to wait to remove old repo support until we can be sure most people aren't or at least know they shouldn't be using old packages.  Should we warn the user they're using a deprecated repo for now?  Should we gather up old packages on COPR/PPA/etc. and host them somewhere under https://repo.saltstack.com/archive?

vutny:
@jfindlay My aim was not to remove completely COPR support from bootstrap, but just to make the code generally more readable/search-able. This particular "private" function will _never be executed_ because all references to it were removed some time ago.
Another function with same purpose still exists: `__install_saltstack_copr_salt_repository()`, which called only from `install_fedora_deps()` function.

Please consider to merge my PR just for the sake of code clarity. Thank you!

jfindlay:
@vutny, thanks for your patience and explanation.  I will review this again.

jfindlay:
It seems that this is part of an incomplete COPR-based support for RHEL in bootstrap.  EL5 needs to be treated separately from EL6 and EL7 because of incompatible changes in the way repos are handled between EL5 and EL6 (compare `__install_saltstack_copr_salt_repository` which is used only for fedora support).  As it is an internal function that is not used in the script, however, it seems that it should be safe to remove.

PR #723, opened 2015-12-31
fix for FreeBSD 11 CURRENT install functions

serge-p:
## issue: 

when following FreeBSD current: 
discovered some versions mismatch issues, 
when attempting to install salt from saltstack repo,
also repository is not being set properly by the script in /etc/pkg/FreeBSD.conf 
## changes: 
1.  proposing to perform initial check, before changing default FreeBSD repository) 
2.  workaround for current version would be installing salt from original FreeeBSD ports 
   and then fixing config paths by creating a symlink  /etc/salt -> /usr/local/etc/salt (default dir from ports)

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4073,15 +4073,20 @@ __configure_freebsd_pkg_details() {
     mkdir -p /etc/pkg/
 
     ## Use new JSON-like format for pkg repo configs
-    conf_file=/usr/local/etc/pkg/repos/freebsd.conf
-    {
-        echo "FreeBSD:{"
-        echo "    url: \"${PKGCONFURL}\","
-        echo "    mirror_type: \"SRV\","
-        echo "    enabled: true"
-        echo "}"
-    } > $conf_file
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -4073,15 +4073,20 @@ __configure_freebsd_pkg_details() {
...
-        echo "FreeBSD:{"
-        echo "    url: \"${PKGCONFURL}\","
-        echo "    mirror_type: \"SRV\","
-        echo "    enabled: true"
-        echo "}"
-    } > $conf_file
-    copyfile $conf_file /etc/pkg/FreeBSD.conf
+    ## check if /etc/pkg/FreeBSD.conf is already in place 
+    if [ ! -f /etc/pkg/FreeBSD.conf ]; then 
+      conf_file=/usr/local/etc/pkg/repos/freebsd.conf
```
jfindlay: @serge-p, would it be better to name this file and repo `saltstack.conf` and `SaltStack`?  Also, what happens when `/etc/pkg/FreeBSD.conf` does exist?

jfindlay:
Thanks, @serge-p.

PR #726, opened 2016-01-07
Sometimes bootstrap doesn't install zmq

justinta:
When installing from git, bootstrap seems to skip over the zmq installation.
This checks if zmq has been installed, if not, it installs it.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2516,7 +2516,7 @@ install_debian_8_git_deps() {
         __apt_get_install_noinput git || return 1
     fi
 
-    if [ "$(python -c 'import zmq; print(zmq.pyzmq_version())')" = "" ]; then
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -2516,6 +2516,10 @@ install_debian_8_git_deps() {
         __apt_get_install_noinput git || return 1
     fi
 
+    if [ "$(python -c 'import zmq; print zmq.pyzmq_version()')" = "" ]; then
```
s0undt3ch: Wont this show a traceback in case its missing?

s0undt3ch: How about `try: import zmq; exit(0); except ImportError: exit(1);` and look for the exit code?

justinta: Oh I like that. Good idea. Thanks!

PR #731, opened 2016-01-13
Don't add zypp repo if it already exists

### LINKED ISSUE #730: zypp repo added twice if repo name does not match devel_languages_python
The zypp repo devel_languages_python is added a second time if the repo name does not match "devel_languages_python".

furlongm:
closes #730


jfindlay:
Thanks, @furlongm.

PR #741, opened 2016-01-25
Use POSIX-Compliant Command-Exists Test

kojiromike:
Suggest the `command` command instead of `which` to determine if the `lsb_release` command exists.
`which` is a non-standard command and this script and its documentation should work across as many POSIX like systems as possible.
`command` is more portable than `which`.

To be consistent with the existing suggestion, the updated suggestion will only act if `lsb_release` exists and is a simple command, not an alias or shell function.

References:
http://stackoverflow.com/a/677212/418413
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -392,7 +392,7 @@ Please run the following commands and report their output when creating a ticket
 .. code:: console
 
   sudo find /etc/ -name '*-release' -print -exec cat {} \;
-  which lsb_release && lsb_release -a
+  command lsb_release -a
 
 
 Testing in Vagrant
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -107,6 +107,14 @@ echodebug() {
 }
 
 #---  FUNCTION  -------------------------------------------------------------------------------------------------------
+#          NAME:  __check_command_exists
+#   DESCRIPTION:  Check if a command exists.
+#----------------------------------------------------------------------------------------------------------------------
+__check_command_exists() {
+    command -v "$1" > /dev/null 2>&1
+}
+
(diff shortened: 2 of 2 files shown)


vutny:
@kojiromike This is a nice thing to have! :+1:
Especially when bootstrapping some stripped down VM or container images without `which` installed.
I could only suggest to rename your new function to `__check_command_exists`. It's a kind of convention in bootstrap script to highlight internal "private" functions.

I'm looking forward for this to be merged.

kojiromike:
@vutny Happy to. To help me understand, why are some functions "private", but others, like `echoerror`, not?

vutny:
@kojiromike Of course there is no such concept as true private functions in POSIX shell or Bash.
If I understand it correctly, in salt bootstrap script it's just a convention to divide functions as "public" which intended to run once in most cases, traceable in debug mode and do the real job to install, configure and start Salt on target system.
On the other hand, there are "private" fictions which are tiny reusable code pieces for common tasks executed from "public" functions. Their names begin with double underscore: `__*() {`.

For historical reasons, I think, some functions such as

```
echoerror() {
echoinfo() {
echowarn() {
echodebug() {
check_pip_allowed() {
usage() {
copyfile() {
movefile() {
```

considered to be "private" but don't have underscores prefix.
I guess this is the task for future refactoring :-)

Anyhow, I would really like to see your PR merged, but that's the authority of @SaltStack team.

jfindlay:
This is very well done, thanks, @kojiromike, and thanks @vutny for the excellent review.

PR #744, opened 2016-01-28
Fix for enabling universe repo on Ubuntu

The-Loeki:
If universe must be enabled, old-releases will always be enabled, leading to errors


vutny:
@The-Loeki :+1:

PR #750, opened 2016-02-05
Enable shallow cloning for version branches by default, not only tags

vutny:
Now running

```
sudo sh bootstrap-salt.sh git 2015.8
```

will also be faster with shallow clone if supported by git version installed.


jfindlay:
@vutny, nice!

Issue #736, opened 2016-01-20
Install on fedora 23 creates invalid salt repo file


marun:
After successfully bootstrapping on a Fedora 23 host, the following repo file was created at `/etc/yum.repos.d/saltstack-salt-fedora-23.repo`:

```
[saltstack-salt]
name=Copr repo for salt owned by saltstack
baseurl=https://copr-be.cloud.fedoraproject.org/results/saltstack/salt/fedora-$releasever-$basearch/
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://copr-be.cloud.fedoraproject.org/results/saltstack/salt/pubkey.gpg
enabled=1
```

A subsequent 'dnf update' resulted in the following warning message:

```
Failed to synchronize cache for repo 'saltstack-salt' from 'https://copr-be.cloud.fedoraproject.org/results/saltstack/salt/fedora-23-x86_64/': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried, disabling.
```

It appears that salt is distributed by Fedora >= 23 and the extra repo is no longer created or required.

vutny:
@marun Looks like this issue has been resolved in the `develop` branch. Please could you close it?

PR #751, opened 2016-02-05
do not install copr repo on fedora 22+

toanju:
- there is no copr repo for fedora 23
- copr repo for fedora 22 is outdated; hence using the packages of fedora
#736


jfindlay:
Also related: https://github.com/saltstack/salt/issues/28142.

PR #755, opened 2016-02-08
Reformat and correct usage instructions

vutny:
Small fixes. Make fit in 80 character width terminal.


jfindlay:
Excellent, thanks, @vutny.

PR #761, opened 2016-02-17
Ability to change cache dir

clarkperkins:
When launching minions into AWS, often the root partition is not very big - only 8GB usually.  I've recently run into an issue with using `file.managed` to manage large files.  Since the files are cached, the root partition fills up and breaks things.

This PR allows users to change the cache dir to be somewhere with more space.


s0undt3ch:
Go Go Jenkins!

jfindlay:
@clarkperkins, this is looking good.  Would you mind rebasing?  Thanks.

clarkperkins:
@jfindlay done!

jfindlay:
Thanks, @clarkperkins.

PR #769, opened 2016-03-03
Allow archive versions

clarkperkins:
It would be nice to allow users to install a specific patch version of salt via the `stable` install instead of the `git` install.  `Stable` installs are significantly faster when using salt-cloud.  This change would allow this to happen.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -238,7 +238,7 @@ __usage() {
 
   Installation types:
     - stable              (install latest stable release, this is default)
-    - stable [version]    (currently only supported on: Ubuntu, CentOS, RedHat)
+    - stable [version]    (currently only supported on: Ubuntu, CentOS, RedHat, Oracle, Scientific)
     - daily               (Ubuntu specific: configure SaltStack Daily PPA)
     - testing             (RHEL-family specific: configure EPEL testing repo)
     - git [branch_or_tag] (install from 'develop' by default)
@@ -1237,7 +1237,7 @@ __ubuntu_codename_translation
 if ([ "${DISTRO_NAME_L}" != "ubuntu" ] && [ "$ITYPE" = "daily" ]); then
     echoerror "${DISTRO_NAME} does not have daily packages support"
     exit 1
-elif ([ "$(echo "${DISTRO_NAME_L}" | egrep '(ubuntu|centos|red_hat)')" = "" ] && [ "$ITYPE" = "stable" ] && [ "$STABLE_REV" != "latest" ]); then
+elif ([ "$(echo "${DISTRO_NAME_L}" | egrep '(ubuntu|centos|red_hat|oracle|scientific)')" = "" ] && [ "$ITYPE" = "stable" ] && [ "$STABLE_REV" != "latest" ]); then
     echoerror "${DISTRO_NAME} does not have major version pegged packages support"
     exit 1
 fi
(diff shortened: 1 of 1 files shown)


clarkperkins:
So regarding this change, this may not be the best way to implement it, but it would be VERY valuable to be able to do `stable` installs with pegged patch versions rather than just minor versions... Otherwise our automated system ends up installing newer minion versions when they are released, and our master doesn't get upgraded as often

timoguin:
This would be great!

For Ubuntu, I am currently doing this in a custom script like so:

```
$ apt-get install -y -qq \
     -o Dpkg::Options::="--force-confdef" \
     -o Dpkg::Options::="--force-confold" \
     salt-minion=2015.8.7+ds-1 \
     salt-master=2015.8.7+ds-1 \
     salt-api=2015.8.7+ds-1 \
     salt-cloud=2015.8.7+ds-1
$ apt-mark hold salt-minion salt-master salt-api salt-cloud
```

jfindlay:
@clarkperkins, thanks for working on this.  There is existing support for `stable <branch>` on ubuntu and CentOS and I would rather keep this syntax for installing exact releases from https://repo.saltstack.com/, `stable <release>`.  Installing the latest packages on 2015.8 would be `stable 2015.8` and installing 2015.8.8 would be `stable 2015.8.8`  What do you think?

bootstrap-salt.sh
```diff
@@ -3032,10 +3032,10 @@ __install_saltstack_rhel_repository() {
         gpg_key="SALTSTACK-GPG-KEY.pub"
     fi
 
-    repo_file="/etc/yum.repos.d/salt-${repo_rev}.repo"
+    repo_file="/etc/yum.repos.d/saltstack.repo"
```
jfindlay, line 3110: I like this change.  We should probably change the repo name on ubuntu/debian and others to `saltstack` since it is likely that you'd want only one saltstack repo configured at a time.

bootstrap-salt.sh
```diff
@@ -1225,7 +1225,7 @@ __ubuntu_codename_translation
 if ([ "${DISTRO_NAME_L}" != "ubuntu" ] && [ "$ITYPE" = "daily" ]); then
     echoerror "${DISTRO_NAME} does not have daily packages support"
     exit 1
-elif ([ "$(echo "${DISTRO_NAME_L}" | egrep '(ubuntu|centos)')" = "" ] && [ "$ITYPE" = "stable" ] && [ "$STABLE_REV" != "latest" ]); then
+elif ([ "$(echo "${DISTRO_NAME_L}" | egrep '(ubuntu|centos|red_hat)')" = "" ] && [ "$ITYPE" = "stable" ] && [ "$STABLE_REV" != "latest" ]); then
```
jfindlay: This should also be updated on the help output on [L239](https://github.com/clarkperkins/salt-bootstrap/blob/feature/allow-archive-versions/bootstrap-salt.sh#L239).  It looks like Oracle and Scientific are also supported by default as well.

clarkperkins:
@jfindlay I agree, I definitely like the syntax better without the `archive/` - I was just trying to get a quick fix working for our environment and `archive/<version>` made it easier to implement at the time :)

jfindlay:
@clarkperkins, I merged your changes in #797, so I'm going to close this.  Thanks for your work!

PR #771, opened 2016-03-10
Fix sleep time option to recognize a numeric argument

vutny:
Silly bug, possibly introduced by wrong merge.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -18,7 +18,7 @@
 #======================================================================================================================
 set -o nounset                              # Treat unset variables as an error
 
-__ScriptVersion="2016.03.09"
+__ScriptVersion="2016.03.10"
 __ScriptName="bootstrap-salt.sh"
 
 #======================================================================================================================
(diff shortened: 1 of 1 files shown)


s0undt3ch:
Oh, you been bumping the date. That's usually only right before merging to stable, but that's alright...

s0undt3ch:
Could you take care of [this one](https://jenkins-testing.saltstack.com/job/bootstrap/job/pr/job/766/job/lint/2/violations/file/bootstrap-salt.sh/) too since you're at it?

Thanks!

vutny:
@s0undt3ch I've already fixed it in 1b9ea4f. Jenkins triggered job for wrong PR No 766 again, but the first attempt was correct.

vutny:
Regarding the version bump, I thought that `develop` branch went far away from `stable` currently, and there is no way to determine the exact commit from bug reports...
Especially when people use `curl -L https://bootstrap.saltstack.com/develop ...`.

s0undt3ch:
The only way to determine is the -dev in the end. If the version of stable and develop match, develop is more recent because of that suffix...

vutny:
@s0undt3ch Acknowledged, thanks. Will add `-dev` suffix in further PRs.

s0undt3ch:
No no. -dev is added automatically :)

vutny:
@s0undt3ch Strange... I can't find where such thing happen... Maybe you meant this?

``` sh
 504 echowarn "Running the unstable version of ${__ScriptName}"
```

This line is commented in `stable`.

s0undt3ch:
Hmm, yes, that's what I was referring to. Sorry.

PR #773, opened 2016-03-14
Updated Salt Repo Domain from fedoraproject.org to fedorainfracloud.org

ninjada:
Updated Salt repo domain from https://copr.fedoraproject.org/* to https://copr.fedorainfracloud.org/* as the original is now a redirect -- this breaks the bootstrap as a html redirect file is downloaded rather than the Salt repository file.


vutny:
Hi @ninjada 
Thanks for your PR, but it was already fixed in f8715cb9.
Could you please close this?

PR #779, opened 2016-03-22
Add config_freebsd_salt func so freebsd puts cfgs in the right place

### LINKED ISSUE #777: FreeBSD 9/10 install does not use copied configs
When bootstrapping FreeBSD 10.2 the script will move the copied minion config (specified with `-c /tmp`) to `/usr/local/etc/salt/minion` as expected but then it looks like the `install_freebsd_9_stable_post` function just [copies over the sample config](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L4342) rather than checking/using `_TEMP_CONFIG_DIR`.

ryanwalder:
FreeBSD uses /usr/local/etc/salt as config dir not /etc/salt, this adds
a wrapper function to set the right config dir for config_salt to use.

Fixes #777

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4088,6 +4088,11 @@ install_arch_check_services() {
 #
 #   FreeBSD Install Functions
 #
+config_freebsd_salt() {
+    _SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/usr/local/etc/salt}
+    config_salt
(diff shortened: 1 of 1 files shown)


vutny:
Good job @ryanwalder :+1:

s0undt3ch:
Go Go Jenkins!

Issue #724, opened 2016-01-06
Remove EPEL repository for RedHat/Centos installs


sjmh:
With the inclusion of all dependent RPMs into repo.saltstack.com, EPEL is no longer necessary.  In a new install, the only thing that is grabbed from EPEL is msgpack and ordereddict ( this is on RHEL 6.6, with only @Core )

```
================================================================================
 Package                          Arch   Version       Repository          Size
================================================================================
Installing:
 PyYAML                           x86_64 3.11-1.el6    repo-saltstack-el6 152 k
 python-crypto                    x86_64 2.6.1-2.el6   repo-saltstack-el6 501 k
 python-jinja2                    x86_64 2.2.1-2.el6   repo-saltstack-el6 465 k
 python-msgpack                   x86_64 0.4.6-1.el6   epel                69 k
 python-requests                  noarch 1.1.0-3.el6   repo-saltstack-el6  70 k
 python-zmq                       x86_64 14.5.0-2.el6  repo-saltstack-el6 479 k
 yum-utils                        noarch 1.1.30-30.el6 repo-saltstack-el6 110 k
Installing for dependencies:
 libyaml                          x86_64 0.1.3-4.el6   repo-saltstack-el6  51 k
 python-babel                     noarch 0.9.4-5.1.el6 repo-saltstack-el6 1.4 M
 python-backports                 x86_64 1.0-5.el6     repo-saltstack-el6 4.9 k
 python-backports-ssl_match_hostname
                                  noarch 3.4.0.2-2.el6 repo-saltstack-el6  11 k
 python-chardet                   noarch 2.2.1-1.el6   repo-saltstack-el6 229 k
 python-ordereddict               noarch 1.1-2.el6     epel               7.6 k
 python-six                       noarch 1.9.0-2.el6   repo-saltstack-el6  28 k
 python-urllib3                   noarch 1.10.2-1.el6  repo-saltstack-el6 100 k
 zeromq                           x86_64 4.0.5-4.el6   repo-saltstack-el6 568 k

Transaction Summary
================================================================================
Install      16 Package(s)
```

If I modify the bootstrap:

```
--- salt-bootstrap.sh.old       2016-01-06 10:31:16.827012016 -0800
+++ salt-bootstrap.sh   2016-01-06 10:36:54.245814917 -0800
@@ -2980,7 +2980,7 @@
 }

 install_centos_stable_deps() {
-    __install_epel_repository || return 1
+    # __install_epel_repository || return 1
[...]

PR #783, opened 2016-03-26
Lack of HTTPS for RPM packages.

archcloudlabs:
Fixing insecure package download.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2350,7 +2350,7 @@ install_debian_6_deps() {
     fi
 
     # shellcheck disable=SC2086
-    wget $_WGET_ARGS -q http://debian.saltstack.com/debian-salt-team-joehealy.gpg.key -O - | apt-key add - || return 1
+    wget $_WGET_ARGS -q https://debian.saltstack.com/debian-salt-team-joehealy.gpg.key -O - | apt-key add - || return 1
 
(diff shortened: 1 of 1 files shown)


vutny:
@jaredestroud Thanks for your PR. Although, I'd not consider those URLs to the `epel-release` package so much secure, because it redirects to plain-HTTP link of nearby mirror.

Actually, it's still questionable topic about the need of EPEL repo configuration... SaltStack official repo, which the bootstrap script uses by default, already contains all Salt dependencies.
This issue: https://github.com/saltstack/salt-bootstrap/issues/724

bootstrap-salt.sh
```diff
@@ -2350,7 +2350,7 @@ install_debian_6_deps() {
     fi
 
     # shellcheck disable=SC2086
-    wget $_WGET_ARGS -q http://debian.saltstack.com/debian-salt-team-joehealy.gpg.key -O - | apt-key add - || return 1
+    wget $_WGET_ARGS -q https://debian.saltstack.com/debian-salt-team-joehealy.gpg.key -O - | apt-key add - || return 1
```
vutny, line 2353: Unfortunately, debian.saltstack.com is not accessible by https.

jfindlay:
@vutny, we can remove those deps in a later pull request.  Also, at least on CentOS and RHEL, you should be able to install the `epel-release` package from the default repos.

Issue #588, opened 2015-04-26
Multiple MiTM security issues


gdm85:
There are multiple places in the bootstrap-salt.sh script where a text file is downloaded via HTTP and its integrity is never checked with a hash or GPG signature. This is an open door for [MITM attacks](https://en.wikipedia.org/wiki/Man-in-the-middle_attack) and puts the salt-bootstrap script no less than at the same level of any `curl http://get-some-cool-stuff.sh | bash` one-liner advices that are [so popular](https://github.com/saltstack/salt-bootstrap#installing-via-an-insecure-one-liner) on the interwebs these days, with the added cleartext insecurity.

I gathered following examples at current master commit ab3dac2c275308e06739450b7ab10bf1f75a97b4
# Examples

https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L2034 (PGP public key)
https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L2676 (repo file)
https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L2684 (repo file)
https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L4086 (repo file)
https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L4279 (repo file)

There are possibly others that can come up with an extensive analysis.
# Hardening tips

Instead of moving these files to a HTTPS source, I suggest to include them inline in the script using heredoc syntax with cat.

vutny:
There is also an issue with downloading bootstrap script itself in the "official" way: #691

vutny:
#691 was fixed.
This is no longer a case, was removed in develop:
https://github.com/saltstack/salt-bootstrap/blob/ab3dac2c275308e06739450b7ab10bf1f75a97b4/bootstrap-salt.sh#L2684
This is also has been fixed:
https://github.com/saltstack/salt-bootstrap/blob/ab3dac2c275308e06739450b7ab10bf1f75a97b4/bootstrap-salt.sh#L2676

Dealing with Debian and Suse becomes more complicated....
Repositories layouts change quite often these days, so I think using heredocs is not an option, at least for now.

gdm85:
@vutny yes, sorry for not including the commit hash in OP links..

Shall I review the current version of the scripts? Or are there other issues open to track the remaining problems?

[...]

Issue #784, opened 2016-03-26
Lack of HTTPS support for RPM download


archcloudlabs:
[Line 3050](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh) shows files being insecurely downloaded from download.fedoraproject.org over HTTP.  I've issued a pull request [here](https://github.com/saltstack/salt-bootstrap/pull/783) addressing the issue.

vutny:
I've discovered that `epel-release` packages are already uploaded to the _extras_ repos in recent releases of CentOS:
http://mirror.centos.org/centos/5/extras/x86_64/RPMS/epel-release-5-4.noarch.rpm
http://mirror.centos.org/centos/6/extras/x86_64/Packages/epel-release-6-8.noarch.rpm
http://mirror.centos.org/centos/7/extras/x86_64/Packages/epel-release-7-5.noarch.rpm

Hooray! Finally we are able to install it with regular `yum --enablerepo=extras -y install epel-release` command. No need to download anything, at least in CentOS. RHEL/Oracle/etc are still questionable.

---

Issues #588 and #724 seem related.

vutny:
I guess this is already resolved by #798. Ping @s0undt3ch

jfindlay:
@vutny, we should probably change bootstrap to install the `epel-release` packages.

PR #788, opened 2016-03-30
Fixed use of HTTP over HTTPS for anonscm.debian.org

gdm85:
Verify downloaded GPG public over HTTP
Added FIXMEs for remaining unsecure downloads over HTTP

Addresses (partially) #588

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -610,6 +610,28 @@ __fetch_url() {
                     ftp -o "$1" "$2" >/dev/null 2>&1            # OpenBSD
 }
 
+#---  FUNCTION  -------------------------------------------------------------------------------------------------------
+#         NAME:  __fetch_verify
+#  DESCRIPTION:  Retrieves a URL, verifies its content and writes it to standard output
+#----------------------------------------------------------------------------------------------------------------------
+__fetch_verify() {
+	local tmpf url sum size
+	url="$1"
+	sum="$2"
+	size="$3"
+
+	tmpf=$(mktemp) && \
+	__fetch_url "$tmpf" "$url" && \
+	test $(stat --format=%s "$tmpf") -eq "$size" && \
+	test $(md5sum "$tmpf" | awk '{ print $1 }') = "$sum" && \
+	cat "$tmpf" && \
+	rm -f "$tmpf"
+	if [ $? -eq 0 ]; then
+		return 0
+	fi
+	echo "Failed verification of $url"
+	return 1
+}
 
 #---  FUNCTION  -------------------------------------------------------------------------------------------------------
 #          NAME:  __gather_hardware_info
@@ -2352,7 +2374,7 @@ install_debian_6_deps() {
     fi
 
     # shellcheck disable=SC2086
-    wget $_WGET_ARGS -q http://debian.saltstack.com/debian-salt-team-joehealy.gpg.key -O - | apt-key add - || return 1
+    __fetch_verify http://debian.saltstack.com/debian-salt-team-joehealy.gpg.key 267d1f152d0cc94b23eb4c6993ba3d67 3100 | apt-key add - || return 1
 
     if [ "$_PIP_ALLOWED" -eq $BS_TRUE ]; then
         echowarn "PyZMQ will be installed from PyPI in order to compile it against ZMQ3"
@@ -2474,7 +2496,7 @@ install_debian_7_deps() {
     fi
 
     # shellcheck disable=SC2086
-    wget $_WGET_ARGS -q http://debian.saltstack.com/debian-salt-team-joehealy.gpg.key -O - | apt-key add - || return 1
+    __fetch_verify http://debian.saltstack.com/debian-salt-team-joehealy.gpg.key 267d1f152d0cc94b23eb4c6993ba3d67 3100 | apt-key add - || return 1
 
     apt-get update || return 1
     __apt_get_install_noinput -t wheezy-backports libzmq3 libzmq3-dev python-zmq python-apt || return 1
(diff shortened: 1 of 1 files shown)


vutny:
@gdm85 Your changes are generally correct and acceptable, but I have some concerns how to resolve those security issues properly. I will put my comments in-line little bit later.

Also there some other issues and PRs which address the same things. I will also try to put the reference to #588 where possible.

bootstrap-salt.sh
```diff
@@ -2352,7 +2374,7 @@ install_debian_6_deps() {
     fi
 
     # shellcheck disable=SC2086
-    wget $_WGET_ARGS -q http://debian.saltstack.com/debian-salt-team-joehealy.gpg.key -O - | apt-key add - || return 1
+    __fetch_verify http://debian.saltstack.com/debian-salt-team-joehealy.gpg.key 267d1f152d0cc94b23eb4c6993ba3d67 3100 | apt-key add - || return 1
```
vutny, line 2377: This line belongs to `install_debian_6_deps()` function. Debian 6 (old old stable) reached [end-of-life](https://wiki.debian.org/LTS) recently. So the entire function should be dropped along with adding Squeeze to outdated releases.

bootstrap-salt.sh
```diff
@@ -2474,7 +2496,7 @@ install_debian_7_deps() {
...
```
vutny, line 2499: As for Debian 7, http://debian.saltstack.com/ should be considered a deprecated repo. See saltstack/salt#31816.
SaltStack team is finally providing up-to-date 2015.8 builds for Wheezy from their Corporate Repository: https://repo.saltstack.com/apt/debian/7/amd64/latest/.

So I guess the entire function needs to be rewritten.

bootstrap-salt.sh
gdm85, line 2377: I understand, in that case I politely leave it to somebody who can do this better.

vutny, line 2377: It's very nice that you raised those security issues, which point us to more general things to be fixed. Thank you.

bootstrap-salt.sh
```diff
@@ -2783,7 +2805,7 @@ install_debian_git_post() {
             if [ -f "${_SALT_GIT_CHECKOUT_DIR}/debian/salt-$fname.init" ]; then
                 __copyfile "${_SALT_GIT_CHECKOUT_DIR}/debian/salt-$fname.init" "/etc/init.d/salt-$fname"
             else
-                __fetch_url "/etc/init.d/salt-$fname" "http://anonscm.debian.org/cgit/pkg-salt/salt.git/plain/debian/salt-${fname}.init"
+                __fetch_url "/etc/init.d/salt-$fname" "https://anonscm.debian.org/cgit/pkg-salt/salt.git/plain/debian/salt-${fname}.init"
```
vutny, line 2808: This is definitely should be written with https:// from the start.
But in general, I don't think that downloading those init files from elsewhere is a good idea.

It's `install_debian_git_post()` function, so System V style initscripts just need to be included in the main Git repo.
As for now, the initscripts which required for Git installation on Debian 7 are located in [salt-pack](https://github.com/saltstack/salt-pack) repository.

bootstrap-salt.sh
```diff
@@ -4810,6 +4836,7 @@ install_opensuse_stable_deps() {
     fi
 
     # Is the repository already known
+    # FIXME: cleartext download over unsecure protocol (HTTP)
```
vutny, line 4839: As far as I know `rpm` in SUSE verifies GPG signatures the same way as in RHEL or others.
The problem here is that download.opensuse.org server isn't accessible by HTTPS.

I think here we need to obtain and verify GPG key first (using hash), and after that configure the repo "manually" using local key file.
Just need to wait until #791 will be merged.

bootstrap-salt.sh
gdm85, line 2377: You're welcome! In cases like these I hate to give a review which seems pedant/bureaucratic, but in reality if you want to be able to say "my deployment is safe against MiTM" each and every piece must satisfy that requirement (that's where I come from).

bootstrap-salt.sh
gdm85, line 2808: Sounds very good to me. One should try to not introduce too many dependencies as a general principle, because each of them can eventually turn out as a [SPOF](https://en.wikipedia.org/wiki/Single_point_of_failure).

bootstrap-salt.sh
gdm85, line 4839: Oh thanks for telling me about that `rpm` feature, I have not used it in a while and I was assuming it wouldn't do the same as Debian/RHEL regarding verifying GPG signatures.

Yes in that case securing the GPG public key will be enough.

bootstrap-salt.sh
```diff
@@ -3057,11 +3079,14 @@ __install_epel_repository() {
         rpm -q curl > /dev/null 2>&1 || yum -y install curl
         # rpm from CentOS/RHEL release 5 does not support HTTP downloads
         # properly, especially 3XX code redirects
+        # FIXME: cleartext download over unsecure protocol (HTTP)
         __fetch_url /tmp/epel-release.rpm "http://download.fedoraproject.org/pub/epel/5/${EPEL_ARCH}/epel-release-5-4.noarch.rpm" || return 1
```
vutny, line 3083: This is not just insecure, but rather ugly and unnecessary these days. I know because I wrote this particular line as quick workaround for CentOS 5 rpm issue :-)
See discussions here: #784 #724

bootstrap-salt.sh
jfindlay, line 2808: It would be much simpler to source from [`pkg`](https://github.com/saltstack/salt/tree/develop/pkg/) rather than salt-pack when installing from git, but the [salt-pack](https://github.com/saltstack/salt-pack/blob/master/file_roots/pkg/salt/2015_8_8/debian8/spec/salt_debian.tar.xz) service files seem to be different from the vanilla systemd files in salt/pkg.

vutny, line 2808: @jfindlay In this particular case the bootstrap fetches plain SystemV-style initscripts for Debian 7. Unfortunately, they aren't present in current develop branch of `saltstack/salt`repo. There are even no `pkg/debian` directory.

That's why I suggest to include the init files to the `saltstack/salt` as well.

jfindlay:
#799 has been merged, so I am closing this.  Thanks @gdm85, @vutny.

PR #794, opened 2016-04-08
Remove the Saltstack repo's alias.

cro:
Apparently when you use non-interactive mode and need to auto-accept GPG keys, you cannot include a repo alias if the alias doesn't match what's in the repo's metadata file.


s0undt3ch:
:+1:

PR #799, opened 2016-04-12
Merge #788

jfindlay:
Merge #788.

Ping @gdm85, @vutny.


s0undt3ch:
:+1:

s0undt3ch:
Sorry, this needs a rebase now.

gdm85:
Thanks for merging this; although I had checked the usage of `local` for dash, I understand why  it's not portable. Good to know.

s0undt3ch:
@gdm85 the script needs to run in all shells and the only way to ensure that is to be fully POSIX compliant and unfortunately, `local` is not.

PR #666, opened 2015-10-02
Virtualenv support for Ubuntu

l2ol33rt:
Proposing two new flags:

``` bash
  -V :  Install salt into virtualenv. If combined with -a all python pkgs will be installed into a virtualenv
  -a :  Install all python pkgs via pip using salts zeromq.txt in requirements but all requires -V to keep python system packages unchanged.
```

I figured it was worth having two different flags since you could opt to have just salt installed into a virtualenv `(-V)` which will also leverage `--system-site-packages`.   Or install salt with all python packages via pip into a virtualenv `(-a)`.

It currently only works for Ubuntu and added the necessary checks for it to error out in the case that those flags try to run on something other than Ubuntu. Also it only works when using the git installs.

My goal here was to have a neat virtualenv that includes only salt and its requirements. Its been very useful for development since I can install separate virtualenvs on the same host and modify /etc/default/salt-master to switch between them.

I broke out the pip installation to its own function `__install_pip_deps` so that other distros could eventually leverage this. 

Here's output from the 2 different runs:
## Running with `-V` and `-a`:

Tested on `Ubuntu 14.04`

``` bash
~# cat /etc/os-release
NAME="Ubuntu"
VERSION="14.04.2 LTS, Trusty Tahr"
ID=ubuntu
```

Installing salt and all pip pkgs into a virtualenv

``` bash
~# ./bootstrap-salt.sh -M -A 127.0.0.1 -a -V /usr/lib/virtualenv/salt git 2015.8
```

Activate virtualenv (just includes pip packages for salt)

``` bash
~# source /usr/lib/virtualenv/salt/bin/activate
(salt)~# pip freeze
backports.ssl-match-hostname==3.4.0.2
certifi==2015.9.6.2
futures==3.0.3
Jinja2==2.8
MarkupSafe==0.23
msgpack-python==0.4.6
pycrypto==2.6.1
PyYAML==3.11
pyzmq==14.7.0
requests==2.7.0
salt===2015.8.0-107-g1a31b19
tornado==4.2.1
wheel==0.24.0
```

Accept the keys and run salt

``` bash
(salt)~# salt-key -A -y
The following keys are going to be accepted:
Unaccepted Keys:
salt-master
Key for minion salt-master accepted.

(salt)~# salt ‘*’ test.ping
salt-master:
    True
```
## With just -V flag

Tested on `Ubuntu 14.04`

`````` bash
~# cat /etc/os-release
NAME="Ubuntu"
[...]

PR #803, opened 2016-04-13
Merge #666

jfindlay:
Merge #666.


jfindlay:
Lint errors are independently verified as fixed.  I'm not sure why the shellcheck run by jenkins is not picking up the changes.

PR #809, opened 2016-04-18
Update epel-release version number

RuriRyan:
The new epel version breaks the vagrant salt provisioner, because it uses the bootstrap script.


clarkperkins:
Ditto on this one - the `epel-release-7-5.noarch` rpm was removed from the mirror that AWS instances pull from, so the bootstrap script cannot run on AWS instances at this time.

PR #810, opened 2016-04-18
Update epel-release version number

RuriRyan:
The new epel version breaks the vagrant salt provisioner, because it uses the bootstrap script.


jfindlay:
Thanks, @RuriRyan.

PR #816, opened 2016-04-18
Merge develop into stable for new release

jfindlay:
### What does this PR do?

Merge develop into stable for new release.  Replaces #813.


s0undt3ch:
Should we pull in my latest merges? Or hold those off until next release?

PR #832, opened 2016-04-28
Don't sleep 11 seconds if the user allowed overwrite of config files

poelzi:
### What does this PR do?

Removes the 11 second penalty if the user explicitly gave the force overwrite (-F) flag.
### Previous Behavior

Always waits 11 seconds.
### New Behavior

Only waits 11 seconds if user did not specify -F
### Tests written?

No

-C implies a overwrite and warns about this with a 11 second sleep.
If the user explicitly allowed overwrite, there is no reason for this 11 second penelty.


rallytime:
Thanks @poelzi!

Issue #833, opened 2016-04-29
SL 6.x bootstrap and yum repo install issues


kmkramer:
on SL 6.x $releasever returns both the major and minor strings for the release, whereas SL 7.2 returns just the major version. this borks salt from being installed.
this will return yum VARS: python -c 'import yum, pprint; yb = yum.YumBase();  \ pprint.pprint(yb.conf.yumvar, width=1)'

``` yaml
sl6.5 host
'releasever': '6.5',

salt grains
osmajorrelease:
        6
    osrelease:
        6.5
    osrelease_info:
        - 6
        - 5

sl7.2 host

'releasever': '7'

salt grains
osmajorrelease:
        7
    osrelease:
        7.2
    osrelease_info:
        - 7
        - 2
```

basically SL6 and SL7 differ in how they report the DISTRO\* vars

kmkramer:
https://repo.saltstack.com/yum/redhat/6.5/x86_64/latest/repodata/repomd.xml: [Errno 14] PYCURL ERROR 22 - "The requested URL returned error: 404 Not Found"

jfindlay:
@kmkramer, thanks for reporting.  Is SL Scientific Linux?  @dmurphy18, I wonder if we should add `6.5` as a symlink in the package repositories to `6` for situations like this.

kmkramer:
correct SL=Scientific Linux

dmurphy18:
@jfindlay @kmkramer Added a symlink for 6.5 -> 6 as requested, should be available with the next point release which is very soon.

dmurphy18:
@kmkramer Upon reflection what about the other versions of Scientific Linux, 6.0 - 6.6. and 6.7 ?

Adding symlinks for all minion versions of SL 6.x is going to look a little messy.

kmkramer:
We only use 6.5 and will migrate to 7.x at some point. Whatever you
officially support probably needs to be addressed.

dmurphy18:
@kmkramer it will be available with the next point release.  Made symbolic links for all SL 6.x releases and got out ahead of the 6.8 release that is probably soon since Redhat just released 6.8

dmurphy18:
@kmkramer Have made the changes and the next update of staging to live for repo.saltstack.com will show the symbolic links.  Hence I would like to close this issue if that is acceptable

kmkramer:
yes thanks

PR #835, opened 2016-04-29
fix  saltstack repo creation in SL5/6

kmkramer:
### What does this PR do?

creation of saltstack repo used $releasever. $releasever acts differently in sl5/sl6 than sl7 does
#833
### Previous Behavior

https://repo.saltstack.com/yum/redhat/6.5/x86_64/latest/repodata/repomd.xml: [Errno 14] PYCURL ERROR 22 - "The requested URL returned error: 404 Not Found"
### New Behavior

yum repolist  returns as expected repo reachable
### Tests written?

No


rallytime:
Hi @kmkramer - Thank you for submitting this pull request. I see that the issue you're referencing and wanting to fix with this PR seems to be resolved with the Scientific Linux mirroring that you've been discussing with @dmurphy18 in #833. Is this fix still needed? 

If this fix is still needed, I have some concerns here. It seems like this PR is touching much more code than it likely needs to, as it is changing some support for Debian/Ubuntu packages and is also removing some functionality such as the `-r` option. We do not want to remove that support. I'd like this fix to be more isolated to fixing the issue at hand of SL5/6 support.

If this fix is no longer needed because of package mirroring, then we can simply close this out.

kmkramer:
It is probably how I did the pull request. The lines I changed were 3313-3325 specifically for the SL repo creation bits.

rallytime:
Ah ok. So is this fix still needed? If so, can you redo/rebase this so that only the changes you made are present?

kmkramer:
I believe so. I'm working on how to fix my fork.

kmkramer:
gonna start over and resubmit at this point

rallytime:
@kmkramer Sounds good. Thank you very much.

PR #840, opened 2016-05-06
Removed libzmq4 and forking-deamon-patch for Opensuse13

justinta:
### What does this PR do?

Libzmq4 does not exists on OpenSUSE13, so I changed it back to libzmq3.

The foking-daemon patch is incorrect on 2015.8.8.2 and would cause bootstrap to fail. I've removed section where this patch is applied.

these changes allow opensuse 13 to properly bootstrap. 

OpenSUSE13 not bootstrapping. 
### Tests written?

No

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5150,7 +5150,7 @@ install_opensuse_stable_deps() {
     if [ "$DISTRO_MAJOR_VERSION" -lt 13 ]; then
         __PACKAGES="${__PACKAGES} libzmq3"
     elif [ "$DISTRO_MAJOR_VERSION" -eq 13 ]; then
-        __PACKAGES="${__PACKAGES} libzmq4"
+        __PACKAGES="${__PACKAGES} libzmq3"
     elif [ "$DISTRO_MAJOR_VERSION" -gt 13 ]; then
         __PACKAGES="${__PACKAGES} libzmq5"
     fi
@@ -5184,13 +5184,6 @@ install_opensuse_git_deps() {
     __zypper_install patch || return 1
 
     __git_clone_and_checkout || return 1
-
-    if [ -f "${_SALT_GIT_CHECKOUT_DIR}/pkg/suse/use-forking-daemon.patch" ]; then
-        # shellcheck disable=SC2164
-        cd "${_SALT_GIT_CHECKOUT_DIR}"
-        echowarn "Applying patch to systemd service unit file"
-        patch -p1 < pkg/suse/use-forking-daemon.patch || return 1
-    fi
 
     if [ -f "${_SALT_GIT_CHECKOUT_DIR}/requirements/base.txt" ]; then
         # We're on the develop branch, install whichever tornado is on the requirements file


justinta:
@isbm, @aboe76 Can I get your eyes on this as well? I'm not sure why we needed the forking-daemon patch to being with. If you could provide your input that would be very helpful!

cachedout:
Woah, hang on. We're doing what now?!

libzmq3 has known issues and is _not_ a recommended lib for use with Salt. We need to circle back around here and see if there's another way to do this.

isbm:
OpenSUSE 13.1 doesn't exist either. :wink:

justinta:
@cachedout The 13.x repo doesn't have libzmq4 available. Just libzmq3. @isbm is there another repo that would have libzmq4 we can use besides http://download.opensuse.org/distribution/13.2/repo/oss/ ?

thatch45:
@jtand how are we installing salt? We should be using the Suse systemmanagement:saltstack repo:
https://build.opensuse.org/project/show/systemsmanagement:saltstack
That has zeromq4 for opensuse 13 in it
and the Suse packages are the awesome sauce

rallytime:
@thatch45 When I followed the instructions for installing with repo.saltstack.com, it installed libzmq 3, not 4.

thatch45:
ok, after looking this over I remember what is going on
we sync the syse repo to our servers because customers require to download our software from our servers, but we just copy the suse package for both consistency and quality, and we are just a little behind on the sync :)

@UtahDave @dmurphy18 who is in charge of syncing the suse packages? We should not let them get out of data like this :)

cachedout:
Ahhhhh...it all makes sense now. Whew! Glad we caught this!

thatch45:
right :)

isbm:
@jtand Well, we _still_ have 13.2 repos. And 13.1 is long time dead and not supported. I would suggest focusing on Leap 42, since it is a rolling distro from now on. If you still need 13.1, you should try building the package yourself, since no one is using 13.1 and it _meant_ to be very dead. It is always a bad idea to put openSUSE on static machines that cannot be upgraded at least within two years.

justinta:
We did some digging and found that the libzmq name is wrong in the opensuse repo, it says libzmq3, but it actually installs 4.0.4. So this looks to be ok. We're going to leave it as is, unless Suse updates the package name. @isbm @thatch45 @cachedout

thatch45:
Right, for the record, the package name in the opensuse updates repo is libzmq4 ad in the release repo it is libzmq3. Since we have a higher assurance that the release repo is there, and we need to pick one name it makes more sense to pick the release repo name of libzmq3

isbm:
I would strongly suggest stop supporting 13.x and kill it. While we still have 13.2 repos online, this will go away pretty soon. Please ignore every bugs on them, refer to update for Leap and never invest into these anymore.
1. **Leap 42** is where stable things are happening. Test upcoming release there and let us together make sure released version works reliably.
2. **Tumbleweed** where latest possible bleeding edge things are happening (segfault? no problem, this is normal, this is Tumbleweed) — test your `develop` there.

thatch45:
@isbm you are the best upstream man of all time. Basically this PR is fine, and we can say no to future opensuse 13 stuff per upstream.

thatch45:
BTW Leap 42 and Tumbleweed are AWESOME!

aboe76:
I have kept a low profile on the suse part because of the uptake on sles and the new suse manager 3 stuff, I'm not affiliated with suse so @isbm is better suited than me to answer these questions.

I'm very happy that the opensuse/suse support in saltstack is taking off like this..!!!

thatch45:
yes, the guys over at Suse have been fantastic, best upstream to work with EVER. It has made me dramatically reconsider Suse as an OS. So much so that our internal servers were switched over to Suse Leap away from CentOS last year and it has been a buttery smooth experience.

aboe76:
@thatch45 , nice !!

PR #847, opened 2016-05-11
Fix debian-7 setup.py install issue in bootstrap-salt.

supertom:
### What does this PR do?

Makes consistent the command to setup.py as the first argument, followed by the optional flags.

Allows installation of salt from git on Debian 7. Currently broken as the arguments are not in the proper order.
### Previous Behavior

Error generated that the option was unknown.
### Tests written?

No. I've tested this change on Debian 7 on GCE. I've also syntax checked the file with bash.

Around commit c8889ce the install argument to setup.py was moved to the
middle or end of the command string.  This breaks the install on
debian-7.

Revised all setup.py commands to have install as the first argument.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2391,7 +2391,7 @@ install_ubuntu_git() {
     fi
 
     if [ -f "${_SALT_GIT_CHECKOUT_DIR}/salt/syspaths.py" ]; then
-        python setup.py --salt-config-dir="$_SALT_ETC_DIR" --salt-cache-dir="${_SALT_CACHE_DIR}" install --install-layout=deb || return 1
+        python setup.py install --salt-config-dir="$_SALT_ETC_DIR" --salt-cache-dir="${_SALT_CACHE_DIR}" --install-layout=deb || return 1
     else
         python setup.py install --install-layout=deb || return 1
     fi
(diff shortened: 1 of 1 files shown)


supertom:
It seems that this only affects older salt versions (< 2015.5).  I haven't investigated setup.py, but it seems to not be an issue for current versions.

As a workaround, I grabbed the bootstrap-sh file from here, which worked:
https://raw.githubusercontent.com/saltstack/salt-bootstrap/v2014.07.29/bootstrap-salt.sh

For consistency, I would recommend this patch as I suspect the reordering was unintentional.  Further optimizations would include:
- Perhaps a function so more code is reusable (and later, testable)
- (not-code related) Web server redirects to allow users a short URL to other versions of bootstrap-salt.sh (develop works, but not v2014.7.29, for example)

vutny:
Hi @supertom 

The options syntax for the `setup.py` was changed and global parameters such as `--salt-config-dir` and `--salt-cache-dir` have been moved under `install` subcommand. So, the "pre-2015.5" version syntax became deprecated these days and bootstrap script was patched accordingly.

Generally speaking, I believe that community supports only last two stable versions of Salt which are 2015.8 and 2016.3 at the moment. And current `stable` branch of `bootstap-salt.sh` is capable of installing Salt >= 2015.5.X on the majority of platforms. It's too hard to cover all cases with all past Salt versions in the single script... 

The best advise here: if you interested in bootstrapping historical Salt code, it's really better to pick `bootstrap-salt.sh` revision close to the Salt release you'd like to install.

rallytime:
I am going to agree with @vutny on this one. Since versions older than 2015.5 are no longer actively supported in Salt, and it is easy to get a tagged version of the bootstrap script that will work with older versions, I am going to close this change. We can continue the discussion if needed here.

PR #854, opened 2016-05-24
Use HTTPS to clone instead of Git protocol

davidjb:
### What does this PR do?

Switches cloning Salt from git:// to using https://.

Security as git:// is an insecure, cleartext protocol and compatibility as some firewalls may block git://'s custom port but will allow HTTPS traffic.  According to GitHub support, performance is the same or better over HTTPS with updates in recent times.

FYI: See https://gist.github.com/grawity/4392747 for a summary of protocols.  GitHub is sadly light on details about which remote to use, but https://help.github.com/articles/which-remote-url-should-i-use/ recommends HTTPS and if you contact support they have actively discouraged git:// for a long time.
### Previous Behavior

Use git:// to clone
### New Behavior

Use https:// to clone
### Tests written?

Yes (adjusted existing tests)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,5 @@
 Version 2016.05.11:
+	* Use HTTPS instead of Git protocol when cloning. (davidjb)
     * Only overwrite the minion config file if '-C' is passed. Otherwise, preserve it. (rallytime) #848
 
 Version 2016.05.10:
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -190,7 +190,7 @@ __check_config_dir() {
 #----------------------------------------------------------------------------------------------------------------------
 _KEEP_TEMP_FILES=${BS_KEEP_TEMP_FILES:-$BS_FALSE}
 _TEMP_CONFIG_DIR="null"
-_SALTSTACK_REPO_URL="git://github.com/saltstack/salt.git"
+_SALTSTACK_REPO_URL="https://github.com/saltstack/salt.git"
 _SALT_REPO_URL=${_SALTSTACK_REPO_URL}
 _DOWNSTREAM_PKG_REPO=$BS_FALSE
 _TEMP_KEYS_DIR="null"
--- a/tests/bootstrap/test_install.py
+++ b/tests/bootstrap/test_install.py
@@ -605,7 +605,7 @@ class InstallationTestCase(BootstrapTestCase):
             0,
             self.run_script(
                 script=None,
-                args=('git', 'clone', 'git://github.com/saltstack/salt.git'),
+                args=('git', 'clone', 'https://github.com/saltstack/salt.git'),
                 cwd='/tmp/git',
                 timeout=15 * 60,
                 stream_stds=True


vutny:
Thanks, @davidjb. But we have a special option `-G` exactly for this:

```
$ ./bootstrap-salt.sh -h

  Usage :  bootstrap-salt.sh [options] <install-type> [install-type-args]

...

    -G  Instead of cloning from git://github.com/saltstack/salt.git, clone from
        https://github.com/saltstack/salt.git
        (Usually necessary on systems which have the regular git protocol port
        blocked, where https usually is not)
```

davidjb:
Interesting, didn't see that. In that case, I'd suggest that a secure connection should be the default, much as the way GitHub advertises SSH and HTTPS but still allows Git:// if you explicitly use it. Thoughts? As best I'm aware there are only downsides to git:// (security, privacy, firewalls and performance).

vutny:
Hmm. In my environment sometimes (very rarely) I got timeout errors from port 443 on github.com when doing lots of cloning/pulling from GitHub in parallel. But `git://` protocol works every time.

It's only my case though and I'm aware that git protocol is insecure. It should mostly be used against local self-hosted Salt repo mirrors I think.

Anyhow, switching from `git://` to `https://` by default should be done properly, inverting the behavior of `-G` option which brings more changes than just replace the string.

@jfindlay @s0undt3ch Any thoughts about breaking option syntax compatibility in favor of tightening security?

ChangeLog
```diff
@@ -1,4 +1,5 @@
 Version 2016.05.11:
+	* Use HTTPS instead of Git protocol when cloning. (davidjb)
```
vutny, line 2: Just a note:  changelog usually updated according to the commit log right before next bootstrap stable version will be released.

rallytime:
@s0undt3ch Just a bump here to see if you'd had a chance to consider conversation around these changes.

s0undt3ch:
I guess we should favor security. I'm cool with switching to HTTPS and inverting the flag.

rallytime:
@davidjb Are you in a position to make the suggested changes above from @vutny? We would like to get this in, but this will need more changes to make the switch properly to invert the `-G` flag.

davidjb:
@rallytime Since this is a backwards incompatible change I'm unsure how you as a project want to manage this.  Of particular concern is that if switching to using `https://``and inverting the`-G`flag so it uses`git://`, this means anyone's current usage of`-G`will go from secure --> insecure (and broken in networks that block`git://`).  Perhaps deprecating`-G`, using`https://`by default and having a separate`--protocol [git|https|git+ssh|etc]` argument might make things clearer.

My time's unfortunately limited at this moment so that's all I can suggest.  How are other backwards-incompatible issues handled?

vutny:
@davidjb Good point about deprecation `-G` option, I like that.
I think `git://` protocol is only useful for local mirrors on self-hosted git server in a secure perimeter of user's network. There is no point to enable it with GitHub.

Otherwise, if someone will need to use `git://` or any other proto explicitly, we have a `-g` option to his/her service, e.g.:

``` sh
sh bootstrap-salt.sh -g 'git://github.com/saltstack/salt.git' git
```

I will try to come up with another PR in the near future. Thanks for the tip!

rallytime:
Thank you for your work and discussion here, @davidjb! This PR has been supersceeded by #945, which contains the removal of the `-G` option and some documentation changes. As such, I'll close this PR in favor of #945.

PR #857, opened 2016-05-24
Keep original name of salt executable executable

themalkolm:
It is a good idea to not rename and keep the original executable name. Renaming to salt.exe we loose version and arch information. It helps during debugging as simply listing temp directory you can ensure that proper version and arch is installed.

### DIFF
--- a/bootstrap-salt.ps1
+++ b/bootstrap-salt.ps1
@@ -217,10 +217,11 @@ If (!$version) {
 #===============================================================================
 # Download minion setup file
 #===============================================================================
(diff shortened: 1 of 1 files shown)


rallytime:
@themalkolm This pull request will need to be rebased, as it is conflicting and we won't be able to merge it as it stands.

@twangboy Once this is rebased, how do these changes look to you?

twangboy:
@rallytime Looks good to me.

themalkolm:
@rallytime done

rallytime:
Thanks!

Issue #856, opened 2016-05-24
What is the purpose of changing default master/minion names?


themalkolm:
This scripts gives another default values for minion id and master address. Is there really a good reason to do so? In salt configuration we see that default master address is `salt` and minion id is its hostname. This bootstrap script silently changes them to `master` and `salt-minion`. Why?

```
  [Parameter(Mandatory=$false,ValueFromPipeline=$true)]
  [string]$minion = "salt-minion",

  [Parameter(Mandatory=$false,ValueFromPipeline=$true)]
  [string]$master = "master"
```

twangboy:
@themalkolm  I agree. This should probably only send parameters to the installer if they are passed. That way installer defaults will be used... which would be hostname for the minion id and salt for the master id. I'll take a look at it.

themalkolm:
Thanks!

twangboy:
@themalkolm I submitted a PR to address this issue. I also fixed another issue that was causing the config layed down by the installer to be overwritten by a config copied from a Vagrant environment.

themalkolm:
@twangboy is there some kind of automatic tests for Windows bootstrap script? Do you test manually on your PC?

twangboy:
Yeah, I don't think there are any tests. Especially for the powershell script. I just test on my box.

jfindlay:
I am closing this since it appears to have been fixed.  @themalkolm, running continuous tests on windows is one of our short term goals, so when that is ready, I imagine that we will be running windows bootstraps very frequently.

PR #858, opened 2016-05-24
Use installer defaults

twangboy:
### What does this PR do?

Addresses the issue of default values for bootstrap differing from those in the installer

Also copies the minion config from TMP before the installer runs, otherwise the copied config would overwrite any changes made by installer

https://github.com/saltstack/salt-bootstrap/issues/856
### Previous Behavior

salt-bootstrap defaulted to:
- minion id: salt-minion
- master: master
### New Behavior

Allows the installer to set the defaults which are:
- minion id: <host name>
- master: salt
### Tests written?

NA


themalkolm:
Oh, was service starting anyway? That explains a bug we have!

PR #860, opened 2016-05-26
Drop support for Debian 6 old-old-stable "squeeze": reached end-of-life

vutny:
### What does this PR do?

Subj. https://www.debian.org/News/2016/20160212
### Previous Behavior

Script tries to bootstrap modern Salt version on outdated distro.
### New Behavior

Display "EOL" error message and exit when Debian 6 would be detected.
### Tests written?

No

Ping @s0undt3ch

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -206,7 +206,7 @@ Supported Operating Systems
 Debian and derivatives
 ~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


vutny:
@s0undt3ch Fixed lint error

s0undt3ch:
:+1:

PR #865, opened 2016-05-30
Add option to make setup.py install quiet

nasenbaer13:
### What does this PR do?

setup.py install is very verbose by default. This makes bootstrapping via slow connections cumbersome, since all the output has to be sent to the origin. The -q switch adds -q to setup.py install commands when installing from git.

n/a
### Previous Behavior

still present / -q switch is only optional
### New Behavior

-q switch turns of verbose installation
### Tests written?

No


rallytime:
@nasenbaer13 I like this idea. However, it looks like this PR needs to be rebased.

nasenbaer13:
@rallytime - rebased!

rallytime:
Thanks @nasenbaer13! This will be a nice option.

jfindlay:
:+11:

PR #876, opened 2016-06-17
CONFIG_PROTECT_MASK is not realiable. Leave CONFIG_PROTECT empty

alinefr:
### For Gentoo install

At the moment CONFIG_PROTECT_MASK does not work when the file is missing. 

salt-bootstrap.sh tries to emerge salt from ACCEPT_KEYWORDS="~amd64", so it needs to write /etc/portage/package.accept_keywords. As the file is missing instead of writing a new file, portage writes ._cfg000\* file as if package.accept_keywords is a protected file. 

Instead of playing around with CONFIG_PROTECT_MASK, we should just disable CONFIG_PROTECT altogether.  As a script we should not expect any user interaction, so setting CONFIG_PROTECT="-*" temporarily should be safe.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5654,7 +5654,7 @@ __gentoo_config_protection() {
     # this point, manually merge the changes using etc-update/dispatch-conf/
     # cfg-update and then restart the bootstrapping script, so instead we allow
     # at this point to modify certain config files directly
-    export CONFIG_PROTECT_MASK="${CONFIG_PROTECT_MASK:-} /etc/portage/package.keywords /etc/portage/package.unmask /etc/portage/package.use /etc/portage/package.license"
+    export CONFIG_PROTECT="-*"
 }
 
 __gentoo_pre_dep() {


rallytime:
Thanks for submitting this change @alinefr - @s0undt3ch Can you look at this one when you get a chance?

s0undt3ch:
My Gentoo days are long gone. Is this really a safe assumption @alinefr?

If so, say something and we'll merge this...

s0undt3ch:
@eliasp care to chimme in?

s0undt3ch:
Sorry, @eliasp could you comment on this change please?

eliasp:
Setting the `CONFIG_PROTECT_MASK` variable to `-*` is too dangerous. It would allow Portage to overwrite any existing configuration (not only in `/etc/portage`) and as `app-admin/salt` pulls in quite a few dependencies and might also trigger rebuilds of further non-dependency packages, this unnecessarily endangers the system.

Instead I'd rather see `CONFIG_PROTECT_MASK` to be properly defined to either contain the full list of involved files for keywording `app-admin/salt`.

So all that should be needed is to add `/etc/portage/package.accept_keywords` to `CONFIG_PROTECT_MASK`.

rallytime:
Thanks @eliasp for the suggestion.

@alinefr If you'd like to make the suggested change from @eliasp above, we'd be happy to get this in.

rallytime:
@alinefr Just a bump here to see if you'd had a chance to look at the suggestion above.

rallytime:
Since we haven't heard back in a while here, I have added #947 to update the CONFIG_PROTECT_MASK with the `accept_keyworks` portage, instead of accepting `-*`, as discussed above. Therefore, I'll close this PR. Thanks to @alinefr for bringing this to our attention.

PR #879, opened 2016-06-20
added python-psutil to fedora deps

robnagler:
### What does this PR do?

Added Fedora dependency python-psutil.

Prints stack trace if psutil not available: saltstack/salt#33942
### Tests written?

No

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -3090,7 +3090,7 @@ install_fedora_deps() {
         __install_saltstack_copr_salt_repository || return 1
     fi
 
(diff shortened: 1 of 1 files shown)


rallytime:
Hi @robnagler - I don't think we want to add this dependency to the bootstrap script. Installing this package should remain optional in bootstrap. We are working on a change in Salt to not stack trace when psutil cannot be imported right now (see https://github.com/saltstack/salt/issues/34043).

rallytime:
@robnagler Thank you again for submitting this. The error caused by saltstack/salt#33942 has now been resolved by https://github.com/saltstack/salt/pull/34182 and is available at the HEAD of all current release branches, which makes this change unnecessary.

Issue #883, opened 2016-06-23
SLES 11 and CentOS 5 syntax error


jfindlay:
### Description of Issue/Question

SLES 11 and CentOS 5 syntax error
### Setup

create a SLES 11 or CentOS 5 VM and log into root
### Steps to Reproduce Issue

``` console
# curl https://raw.githubusercontent.com/saltstack/salt-bootstrap/e471bb27f4e617d999528f150a6c04577a4d048c/bootstrap-salt.sh > bootstrap-salt.sh ; chmod +x bootstrap-salt.sh ; ./bootstrap-salt.sh -D -P git v2016.3.0'
...
./bootstrap-salt.sh: line 629: *: unbound variable
```
### Versions and Systems

https://raw.githubusercontent.com/saltstack/salt-bootstrap/e471bb27f4e617d999528f150a6c04577a4d048c/bootstrap-salt.sh

PR #885, opened 2016-06-24
save invocation command and arguments into variables

jfindlay:
### What does this PR do?

So that the invocation line is correctly printed.  In addition, on older
shells, such as those on CentOS 5 and SLES 11, a `shift` statement
results in `${*}` being undefined.

Fixes #869.
### Tests written?

No


rallytime:
Fixes #883

PR #893, opened 2016-06-28
RFC: Add tests for bootstrap-salt.ps1

themalkolm:
**DO NOT MERGE AS EVERYTHING IS CONFIGURED TO WORK ON MY REPO**

We heavily rely on this script and would like to somehow ensure it won't be broken in future releases. I also would like to fix some minor issues but very reluctant unless there are tests. So here you are, first attempt to make tests using https://www.appveyor.com/

I'm open to comments, suggestions and ideas how it must be structured in a better way. For now it heavily relies on running in AppVeyor and therefore no cleanup logic as AppVeyor will spin up a new worker for every build.
### What does this PR do?

Adds AppVeyor configuration enabling automatic smoke tests for `bootstrap-salt.ps1`. It does not test for all possible cases. Itmostly works as smoke tests to ensure that at least something works. While it sounds strange, the power is in the fact that it runs **automatically** and has in-built support for matrix builds.

It add smoke tests for `bootstrap-salt.ps1` and a nice status badge.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,6 +2,8 @@
 Bootstrapping Salt
 ==================
 
+|windows_build|
+
 Before `Salt`_ can be used for provisioning on the desired machine, the binaries need to be
 installed. Since `Salt`_ supports many different distributions and versions of operating systems,
 the `Salt`_ installation process is handled by this shell script ``bootstrap-salt.sh``.  This
@@ -535,3 +537,7 @@ Salt is ready and working in the Docker container with Minion authenticated on M
 
 
 .. vim: fenc=utf-8 spell spl=en cc=100 tw=99 fo=want sts=2 sw=2 et
+
+.. |windows_build|  image:: https://ci.appveyor.com/api/projects/status/github/themalkolm/salt-bootstrap?branch=develop&svg=true
+    :target: https://ci.appveyor.com/project/themalkolm/salt-bootstrap
+    :alt: Build status of the develop branch on Windows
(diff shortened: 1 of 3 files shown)


themalkolm:
You can check how badge looks like here https://github.com/themalkolm/salt-bootstrap

themalkolm:
You can check how everything looks here https://ci.appveyor.com/project/themalkolm/salt-bootstrap

rallytime:
What do you think about this @s0undt3ch?

s0undt3ch:
Some tests is better than no tests...
I'll setup appveyor for this repo in a bit.

s0undt3ch:
Woops, missed the no merge part.

Want to fix the links to your own repo as a new PR?

themalkolm:
@s0undt3ch sure, but could you provide me a link to appveyor project you created? Can't find. You also need to configure it to build stable, develop & pull requests I believe.

Then we need to agree (would be nice to do it explicitly) that badge should show current status of the develop branch. Having this badge rises the next question if we should show travis badge as well and enable travis pull requests too.

s0undt3ch:
https://ci.appveyor.com/project/saltstack-public/salt-bootstrap

As for badges, I'd like to leave them out for now because enabling Travis for the Linux side will give a false sense of a working script because it only tests Ubuntu.

There are plans to cover all distributions the script supports but its not an easy task and we don't currently have the free time to work on this. But its not forgotten and the SaltStack QA team does run a huge batch of manual tests before releasing the script.

themalkolm:
@s0undt3ch well, not like we test all Windows versions here too ...

PR #899, opened 2016-07-04
Set repo_arch for raspberry pi installs.

mrichar1:
### What does this PR do?

Set `$repo_arch` to `armhf` for armv7l chipsets so that raspbian builds can find the packages built for that arch.


rallytime:
Thank you @mrichar1

PR #919, opened 2016-07-19
Install `salt-cloud` package with `-L` option in `stable` mode

vutny:
### What does this PR do?

It allows to install `salt-cloud` package (from SaltStack Corp repo in most cases) using `-L` option, not only Apache Libcloud Python module.
Currently supported on RHEL-family distributions, support for other platforms will be added soon.
### Previous Behavior

By giving the `-L` option to the `bootstrap-salt.sh` script, it only installed `apache-libcloud`  from `pip` without actually installing `salt-cloud`.
### New Behavior
- `sh bootstrap-salt.sh -L stable` installation mode:
  
  install `salt-cloud` package with all needed dependencies using package manager from configured SaltStack repository. `pip` not required.
- `sh bootstrap-salt.sh -L git` installation mode:
  
  1) install `python-libcloud` package using package manager from any available repository (SaltStack, EPEL, local mirror). `pip` not required.
  2) copy relevant config files from Salt checkout directory

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -291,6 +291,7 @@ __usage() {
         the master.
     -s  Sleep time used when waiting for daemons to start, restart and when
         checking for the services running. Default: ${__DEFAULT_SLEEP}
+    -L  Also install salt-cloud and required python-libcloud package
     -M  Also install salt-master
     -S  Also install salt-syndic
     -N  Do not install salt-minion
@@ -317,8 +318,6 @@ __usage() {
         \${BS_SALT_ETC_DIR}/minion.d/99-master-address.conf
     -i  Pass the salt-minion id. This will be stored under
         \${BS_SALT_ETC_DIR}/minion_id
-    -L  Install the Apache Libcloud package if possible
-        (required for salt-cloud)
     -p  Extra-package to install while installing Salt dependencies. One package
         per -p flag. You're responsible for providing the proper package name.
     -H  Use the specified HTTP proxy for all download URLs (including https://).
@@ -1376,7 +1375,7 @@ if [ "$_INSTALL_SYNDIC" -eq $BS_TRUE ]; then
 fi
 
 if [ "$_INSTALL_CLOUD" -eq $BS_TRUE ] && [ "$_CONFIG_ONLY" -eq $BS_FALSE ]; then
-    echoinfo "Installing Apache-Libcloud required for salt-cloud"
+    echoinfo "Installing salt-cloud and required python-libcloud package"
 fi
 
 if [ $_START_DAEMONS -eq $BS_FALSE ]; then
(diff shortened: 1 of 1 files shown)


vutny:
Added support for Debian. In the meanwhile, fixed `git` installation mode: `python-yaml` package was missed in dependencies.

vutny:
Added support for Ubuntu. Also fixed virtualenv installation.

vutny:
Added support for Fedora.

vutny:
Added support for Amazon. Mostly was done in the commit fot RHEL.

bootstrap-salt.sh
```diff
@@ -6026,6 +6030,23 @@ config_salt() {
         fi
     fi
 
+    if [ "$_INSTALL_CLOUD" -eq $BS_TRUE ]; then
```
rallytime, line 6059: Hrm...I am not sure if the bootstrap script should be responsible for something like this. The rest of your changes look great, but when I got to this block, it seems like it is outside of the scope of "install the salt-cloud package along with libcloud when I pass `-L`"

I'd like @s0undt3ch to weigh in on this matter before we merge this.

rallytime:
@vutny Thank you very much for this change. I think overall this looks great and the cleanup you've done in here is great. 

I had one thing that needs to be addressed before we can merge this in. Once we hear from @s0undt3ch, and a decision is made about that last block, we can get this in.

vutny:
Hi @rallytime ! Thanks for your review.

I've updated `config_salt()` function this way because I would like to have some config files examples in place after `git` bootstrap. Actually, the `salt-cloud` package also installs some basic `/etc/salt.cloud*` file structure.

The main point here is to have something you could start from just right now instead of searching other repository or docs to copy-and-paste example configuration. I thought it would be helpful.

But, if your concern is that it's not a Salt Bootstrap job, I can just create necessary directories for `salt-cloud` and put only `/etc/salt/cloud` file from `conf` directory in Salt repo clone.

vutny:
By the way, added support for SUSE. Also fixed some dealing with `systemd` on RHEL during `git` installation.

rallytime:
@vutny Ah, OK. I can see how that would be useful. I think if we're going to do that then, I'd like to gate that part of the functionality behind something like passing `-F` and/or `-C`/`-c`. I don't want to run the risk of overwriting people's cloud files if they run the bootstrap script with `-L`, especially since this is a change in behavior where that would previously only install apache-libcloud.

vutny:
@rallytime We are safe here. I've used `__copyfile` function which honours `$_FORCE_OVERWRITE` variable which, in turn, controlled by `-F` and `-C` command line options. Nothing will be overwritten without an explicit request by user.

rallytime:
Ah, you're right. I didn't see that before. This is a big change, so thanks for explaining this part.

s0undt3ch:
:+1:

Thanks @vutny and @rallytime.

Issue #853, opened 2016-05-23
Cannot install salt based off of a commit hash when specifying a different repo


Ch3LL:
### Description of Issue/Question

It seems I cannot install salt with the bootstrap when using a commit hash from a different repo.
### Steps to Reproduce Issue
1. `sh bootstrap-salt.sh -g https://github.com/jfindlay/salt.git -M -P git cf81afab477ce03177a46a5f5f3eb89df92a2576`.
2. And you will see this error:

```
 *  WARN: The git revision being installed does not match a Salt version tag. Shallow cloning disabled
Cloning into 'salt'...
Checking out files: 100% (3788/3788), done.
 *  INFO: Adding SaltStack's Salt repository as a remote
From git://github.com/saltstack/salt
 - [tag update]      v2015.5.1  -> v2015.5.1
 * [new tag]         v2015.8    -> v2015.8
 * [new tag]         v2015.8.0rc1 -> v2015.8.0rc1
 * [new tag]         v2015.8.0rc3 -> v2015.8.0rc3
 * [new tag]         v2015.8.0rc4 -> v2015.8.0rc4
 * [new tag]         v2015.8.0rc5 -> v2015.8.0rc5
 * [new tag]         v2015.8.10 -> v2015.8.10
 * [new tag]         v2015.8.5  -> v2015.8.5
 * [new tag]         v2015.8.6  -> v2015.8.6
 * [new tag]         v2015.8.7  -> v2015.8.7
 * [new tag]         v2016.3    -> v2016.3
 * [new tag]         v2016.3.0rc0 -> v2016.3.0rc0
 * [new tag]         v2016.3.0rc1 -> v2016.3.0rc1
 * [new tag]         v2016.3.0rc2 -> v2016.3.0rc2
error: pathspec 'origin/cf81afab477ce03177a46a5f5f3eb89df92a2576' did not match any file(s) known to git.
 * ERROR: Failed to run install_centos_git_deps()!!!
```
### Versions and Systems

```
[root@ip-10-150-35-101 ~]# sh bootstrap-salt.sh -v
bootstrap-salt.sh -- Version 2016.05.11
```

vutny:
Hi @Ch3LL 

I have fixed this in PR #922. Welcome to test.

PR #922, opened 2016-07-22
Fix git install mode when using commit hash with `-g` option

vutny:

Fixes issue #853
### Previous Behavior

Command `git checkout origin/<commit_hash>` is invalid and it fails.
### New Behavior

Check if the specified branch or tag is actually exists in the `origin` remote repo and only then do checkout using `git checkout origin/<branch_name>` syntax.


rallytime:
Thanks @vutny!

PR #925, opened 2016-08-03
missing python dependency for salt-cloud added

toanju:
python-netaddr is not installed in case -L is used


toanju:
Just as a reference, python-netaddr used in openstack and nova cloud modules.

rallytime:
Hi @toanju - Thank you for submitting this change. It looks like this needs to be rebased before we can get this in, however, as there are merge conflicts.

toanju:
@rallytime done

rallytime:
Great! Thanks!

PR #928, opened 2016-08-12
Disable the '--enablerepo=${_EPEL_REPO}' part of yum install cmds when passing -R/-r

### LINKED ISSUE #924: Remove EPEL references in install_centos_stable_deps when using -r/R

Problems arise when using the `-R` option installing CentOS stable packages when you hit the lines containing references to the EPEL repo, which hasn't been configured. We need to remove those references when `-R` is passed as a script arg.

Also need to audit the rest of the script for these references, not just the `install_centos_stable_deps` function.
### Versions and Systems

`2016.07.07`

rallytime:
If we've passed in `-R` or `-r`, we should not be using the `--enablerepo=` when running a `yum install`, as `-R/-r` is supposed to disable external repo configuration. 

Fixes #924

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1599,13 +1599,19 @@ __apt_get_upgrade_noinput() {
 #   DESCRIPTION:  (DRY) apt-get install with noinput options
 #----------------------------------------------------------------------------------------------------------------------
 __yum_install_noinput() {
+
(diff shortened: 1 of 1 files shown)


rallytime:
@s0undt3ch Can you review this when you get a moment, please?

s0undt3ch:
:+1:

PR #931, opened 2016-08-14
Configure `salt-api` service, but do not attempt to start it on all OS

vutny:
### What does this PR do?

Makes sure that dealing with `salt-api` service is the same on all distros:
- inscripts/rc configs/systemd units should be put in place, so user could start it manually with OS specific tools
- since the service should be opt-in and not necessarily started on boot, there is no need to check the status and run-level.


rallytime:
This clean up will be very helpful. Thanks @vutny!

Issue #939, opened 2016-08-18
salt-cloud deploy fails in package update (without helpful error output)


Reiner030:
As already in https://github.com/saltstack/salt/issues/35561#issuecomment-240819086 mentioned:

```
root@salt-euc1-01:~# lsb_release -d
Description:    Debian GNU/Linux 8.5 (jessie)
root@salt-euc1-01:~# salt-call --version
salt-call 2016.3.2 (Boron)
root@salt-euc1-01:~# salt-cloud --version
salt-cloud 2016.3.2 (Boron)
```

```
...
Fetched 655 kB in 5s (110 kB/s)
W: Failed to fetch http://cloudfront.debian.net/debian/dists/jessie-backports/main/source/Sources  Hash Sum mismatch

W: Failed to fetch http://cloudfront.debian.net/debian/dists/jessie-backports/main/binary-amd64/PackagesIndex  Hash Sum mismatch

E: Some index files failed to download. They have been ignored, or old ones used instead.
 * ERROR: Failed to run install_debian_8_deps()!!!
[ERROR   ] Failed to deploy 'deploy-euc1-01'. Error: Command 'ssh -t -t -oStrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null -oControlPath=none -oPasswordAuthentication=no -oChallengeResponseAuthentication=no -oPubkeyAuthentication=yes -oIdentitiesOnly=yes -oKbdInteractiveAuthentication=no -i /srv/pillar/salt-cloud/setup-key.pem -p 22 admin@172.31.9.103 'sudo /tmp/.saltcloud-391d19a3-e884-4d20-aac3-2c4d634822c7/deploy.sh -c '"'"'/tmp/.saltcloud-391d19a3-e884-4d20-aac3-2c4d634822c7'"'"' -U -p curl'' failed. Exit code: 1
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/cloud/__init__.py", line 2213, in run_map
    profile, local_master=local_master
  File "/usr/lib/python2.7/dist-packages/salt/cloud/__init__.py", line 1281, in create
    output = self.clouds[func](vm_)
  File "/usr/lib/python2.7/dist-packages/salt/cloud/clouds/ec2.py", line 2599, in create
    for key, value in six.iteritems(salt.utils.cloud.bootstrap(vm_, __opts__)):
[...]

PR #940, opened 2016-08-19
[SECURITY] Properly handle GnuPG keys for APT repositories

vutny:
### What does this PR do?

It allows to handle GPG keys for configured APT repositories on Debian/Ubuntu in _secure_ or _INsecure_ way depending on user request.

First, `apt-key adv` command is the preferred method for obtaining public keys by different URLs without any additional downloading tools such as `wget`; it also honours `-I` and `-l` parameters from the Salt Bootstrap.

Second, having pubkey fingerprints hardcoded in the script is unreliable because of:
- We have only main Debian archive key verification, but Security and Backports repositories haven't been covered. Also, derivative distros such as Raspbian has different keys, so adding the other distro key is meaningless and mostly unwanted. 
- Once the key would be expired or revoked, it will require immediate patching of the Salt Bootstrap script; otherwise the attempt of installing wrong key could fail or leave the system in an inappropriate state.

This PR mostly address those issues. See the details below in behavior sections.

Should indirectly resolve issue #939 by skipping the check of `apt-get update` command status.
### Previous Behavior
- Bootstrap script attempts to upgrade `debian-archive-keyring` each time. Although, it will fail if archive signature would be changed.
- Install hard-coded pub key if the fingerprint wasn't found.
### New Behavior
- Bootstrap script updates archive keys only if the user granted the permission to upgrade the whole system with `-U` option. The signature of `debian-archive-keyring` package will not be verified if user allowed insecure downloading with `-I` option. This enables the possibility to bootstrap Salt during distro PKI renewal period on older systems or outdated VM/container images.
- System upgrade will be done prior to any bootstrapping operations (other repo configurations and packages installations). This should go smoothly, because `debian-archive-keyring` would be updated before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 /.project
-*.swp
+*.sw?
 .vagrant
 
 # Pycharm
 .idea
-
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -218,9 +218,10 @@ _EPEL_REPO=${BS_EPEL_REPO:-epel}
 _EPEL_REPOS_INSTALLED=$BS_FALSE
 _UPGRADE_SYS=${BS_UPGRADE_SYS:-$BS_FALSE}
 _INSECURE_DL=${BS_INSECURE_DL:-$BS_FALSE}
-_WGET_ARGS=${BS_WGET_ARGS:-}
 _CURL_ARGS=${BS_CURL_ARGS:-}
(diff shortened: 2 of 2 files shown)


rallytime:
This looks good to me, but I'd like @s0undt3ch to take a look before merging this in since it's such a large and important change.

rallytime:
I spot with @s0undt3ch offline about this one and he says this is ready to go. Thanks again @vutny!

Issue #944, opened 2016-08-25
Problem with proxy, unable to look up github.com


LuboVarga:
### Description of Issue/Question

I am using installation (I want latest develop thing in my centos 7 vagrant env, to test https://github.com/saltstack/salt/issues/35651#issuecomment-242183356) using this:

```
curl -o bootstrap_salt.sh -L https://bootstrap.saltstack.com
sudo sh bootstrap_salt.sh -H 'http://proxy1.aaa.sk' git develop
```
### Setup

I have setup in my env http and https proxy. Most applications are ok with this and will work (curl, wget, firefox).

```
[root@vagrant-backend-prometheus ~]# env | grep -i proxy
NO_PROXY=localhost,127.0.0.1,dev1.aaa.sk
http_proxy=http://proxy1.aaa.sk:8080
ftp_proxy=http://proxy1.aaa.sk:8080
FTP_PROXY=
https_proxy=http://proxy1.aaa.sk:8080
HTTPS_PROXY=http://proxy1.aaa.sk:8080
no_proxy=127.0.0.1, localhost, *.aaa.sk
HTTP_PROXY=http://proxy1.aaa.sk:8080
```
### Steps to Reproduce Issue

Just run bootstrap script. I have also provided proxy parameter to it.

```
[root@vagrant-backend-prometheus ~]# sudo sh bootstrap_salt.sh -D -H 'http://proxy1.aaa.sk' git develop 
 *  INFO: Running version: 2016.08.16
 *  INFO: Executed by: sh
 *  INFO: Command line: 'bootstrap_salt.sh -D -H http://proxy1.aaa.sk git develop'

 *  INFO: System Information:
 *  INFO:   CPU:          GenuineIntel
 *  INFO:   CPU Arch:     x86_64
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   3.10.0-327.28.2.el7.x86_64
 *  INFO:   Distribution: CentOS 7.2

 * DEBUG: Binaries will be searched using the following $PATH: /sbin:/bin:/usr/sbin:/usr/bin
 *  INFO: Using http proxy http://proxy1.aaa.sk
 *  INFO: Installing minion
 * DEBUG: install_centos_7_git_deps not found....
 * DEBUG: install_centos_7_2_git_deps not found....
 * DEBUG: install_centos_7_deps not found....
 * DEBUG: install_centos_7_2_deps not found....
 *  INFO: Found function install_centos_git_deps
[...]

PR #945, opened 2016-08-28
Deprecate `-G` option: use `https` scheme for GitHub by default

vutny:
### What does this PR do?

It makes Salt Bootstrap to clone from SaltStack GitHub repo using https scheme by default.
Also, the `-G` option is going to be deprecated. REAME and usage info updated accordingly.

It supersedes PR #854 and completely resolves issue #944.
### Previous Behavior

The script utilized insecure git protocol to fetch Salt sources from GitHub.
### New Behavior

The script will clone Salt repo by https.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -66,7 +66,7 @@ If you want to install a specific release version (based on the Git tags):
 .. code:: console
 
   curl -o bootstrap_salt.sh -L https://bootstrap.saltstack.com
-  sudo sh bootstrap_salt.sh git v2015.8.8
+  sudo sh bootstrap_salt.sh git v2016.3.2
 
 To install a specific branch from a Git fork:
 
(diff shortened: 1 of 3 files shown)


rallytime:
@vutny This looks good and thanks for putting this together. Perhaps it would make sense to document how people can use the bootstrap script like they were using it with `-G`, but with different options. For example, how to use the `-G` option for testing different tags/commits on our individual forks of the salt repo. Just in case people are wondering where that option went.

rallytime:
Oh! Bah. I was confusing the `-G` and `-g` options before. Never mind my comment above.

The-Loeki:
In the future I'd really like `deprecated` to mean just that; it works, with a message that it's likely to break in the near future. 

These commits 'just' remove the feature and therefore borking our wrapper scripts :(

vutny:
@The-Loeki Your concern is totally valid. Sorry if that caused you troubles.
I agree that maybe we need to put the parsing of `-G` option back, but just display a warning message like:

```
No need to specify this option anymore, now it's default behavior.
```

rallytime:
@The-Loeki I had that thought when reviewing this PR, but I forgot to make an issue for it. I created #949 to track the properly deprecate `-G`, which should block the release of the next stable version. We don't want to break people's usage in the stable release!

The-Loeki:
Well np pplz, it's not a dev version for nothing ;)

Issue #953, opened 2016-09-12
Fails on Arch with -j or -J because of python3 default


fizmat:
### Description of Issue/Question

When running on Arch with -j or -J there is an error, and another error when reporting the first one.

What actually happens:
all libraries are installed on python2 (correctly) by salt-bootstrap
python is a symlink to python3 on Arch.
__overwriteconfig() uses python -c "~code~"
python3 does not have yaml installed, there's an error
when reporting the error there is again python -c "... print ~something~ ..."
print syntax is different in python3 so there's another error, but the original one is lost
I have a patch, will create a pull request soon.
### Setup

A new vagrant box with terrywang/archlinux
### Steps to Reproduce Issue

sudo sh bootstrap-salt.sh -j '{"master": "example.com"}' -D

```
......
 *  INFO: Running config_salt()
  File "<string>", line 1
    import os; print os.stat('/tmp/salt-config-7qiJA6VC').st_size;
                      ^
SyntaxError: invalid syntax
bootstrap_salt.sh: line 1921: [: : integer expression expected
  File "<string>", line 1
    tmp_file=open('/tmp/salt-config-7qiJA6VC'); print tmp_file.read(); tmp_file.close()
                                                             ^
SyntaxError: invalid syntax
 * DEBUG: 
 * ERROR: Python error encountered. This is likely due to passing in a malformed JSON string. Please use -D to see stacktrace.
 * ERROR: Failed to run config_salt()!!!
 * DEBUG: Removing the logging pipe /tmp/bootstrap-salt.logpipe
 * DEBUG: Killing logging pipe tee's with pid(s): 18473
```

rallytime:
Thanks for finding and fixing this issue @fizmat! I'll close this since I just merged in your fix.

PR #954, opened 2016-09-12
Fix issues with replacing configs when python3 is default in os.

fizmat:
### What does this PR do?
1. __overwriteconfig does not use python when reporting errors anymore (less stuff to break)
2. __overwriteconfig checks if python can import yaml before using it to convert configs. If it can't it assumes python is python3 and uses python2 instead

https://github.com/saltstack/salt-bootstrap/issues/953
### New Behavior

`bootstrap-salt.sh -j '{"master": "example.com"}'` works on Arch now

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1898,8 +1898,15 @@ __overwriteconfig() {
         tempfile="/tmp/salt-config-$$"
     fi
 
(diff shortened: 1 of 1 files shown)


rallytime:
Thanks for fixing this @fizmat. Looks great!

Issue #952, opened 2016-09-12
Bootstrap puts keys in wrong place on FreeBSD when -c is passed


justinta:
### Description of Issue/Question

When bootrapping FreeBSD and passing `-c /tmp/somedir/` for the keys, the keys get placed in `/etc/salt/pki` instead of `/usr/local/etc/salt/pki`. This causes issues when bootstrapping via salt-cloud. 

Replaces https://github.com/saltstack/salt/issues/36197
### Setup

Create some keys in a tmp dir to pass in with -c
### Steps to Reproduce Issue

`sh install_salt.sh -c /tmp/testdir git v2016.3.1`
### Versions and Systems

No salt installed on vm
install_salt.sh -- Version 2016.08.16

PR #955, opened 2016-09-12
Allow correct salt paths to be used when -c is passed

### LINKED ISSUE #36197: FreeBSD fails to start daemons with salt-cloud
When provisioning a freebsd vm with salt-cloud and bootstrap-salt.sh the preseeded key gets put into /etc/salt, when it should be put into /usr/local/etc/salt/. When using just bootstrap and starting the minion service by hand key's are populated appropriately.

### Setup
`salt-cloud -p <cloud-provider>-freebsd10 test-freebsd10`

If you're using rackspace it looks like, `salt-cloud -p rs-freebsd10 test-freebsd10`

### Steps to Reproduce Issue
Wait for the salt-cloud provision to finish and fail.

On the new freebsd minion check both /etc/salt/pki/ and /usr/local/etc/salt/pki/. Both directories will have the key placed in them. 

### Versions Report
```yaml
Salt Version:
           Salt: 2016.3.0

Dependency Versions:
           cffi: 1.7.0
       cherrypy: 3.8.0
       dateutil: 2.4.2
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: 1.5.0
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: 1.4.4
       M2Crypto: 0.24.0
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.7
   mysql-python: 1.2.5
      pycparser: 2.14
       pycrypto: 2.6.1
         pygit2: Not Installed
         Python: 2.7.12 (default, Jun 28 2016, 08:31:05)
   python-gnupg: 0.3.7
         PyYAML: 3.12
          PyZMQ: 15.2.0
           RAET: 0.6.5
          smmap: 0.9.0
        timelib: 0.2.4
        Tornado: 4.4.1
            ZMQ: 4.1.5

System Versions:
           dist:
        machine: x86_64
        release: 4.1.3
         system: Linux
        version: Not Installed
```

justinta:
### What does this PR do?

Allows keys to be placed in the correct directory when `-c` is passed to bootstrap. 

@rallytime 

Fixes https://github.com/saltstack/salt/issues/36197

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4401,10 +4401,6 @@ install_arch_check_services() {
 #
 #   FreeBSD Install Functions
 #
-config_freebsd_salt() {
-    _SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/usr/local/etc/salt}
-    config_salt
-}
 
 __freebsd_get_packagesite() {
     if [ "$CPU_ARCH_L" = "amd64" ]; then
@@ -4571,7 +4567,7 @@ install_freebsd_git_deps() {
     # Let's trigger config_salt()
     if [ "$_TEMP_CONFIG_DIR" = "null" ]; then
         _TEMP_CONFIG_DIR="${_SALT_GIT_CHECKOUT_DIR}/conf/"
-        CONFIG_SALT_FUNC="config_freebsd_salt"
+        CONFIG_SALT_FUNC="config_salt"
 
     fi
 
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4563,15 +4563,16 @@ install_freebsd_git_deps() {
     fi
     echodebug "Finished patching"
 
+    # Set _SALT_ETC_DIR to ports default
+    _SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/usr/local/etc/salt}
+    # We also need to redefine the PKI directory
+    _PKI_DIR=${_SALT_ETC_DIR}/pki
+
     # Let's trigger config_salt()
     if [ "$_TEMP_CONFIG_DIR" = "null" ]; then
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -4563,15 +4563,16 @@ install_freebsd_git_deps() {
...
+    # Set _SALT_ETC_DIR to ports default
+    _SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/usr/local/etc/salt}
+    # We also need to redefine the PKI directory
+    _PKI_DIR=${_SALT_ETC_DIR}/pki
+
     # Let's trigger config_salt()
     if [ "$_TEMP_CONFIG_DIR" = "null" ]; then
         _TEMP_CONFIG_DIR="${_SALT_GIT_CHECKOUT_DIR}/conf/"
-        CONFIG_SALT_FUNC="config_salt"
+        CONFIG_SALT_FUNC="config_freebsd_salt"
```
rallytime: Do we need to switch this here? Since all that does is call the same line that is set above: `_SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/usr/local/etc/salt}` and then call config_salt?

I also don't see anywhere else that we're calling that function, so we might want to just remove it. What do you think? I might be missing something.

Here's the line I referenced above: https://github.com/jtand/salt-bootstrap/blob/e6e07c63b7a6ffa7bfc1520073121753137ce850/bootstrap-salt.sh#L4405

rallytime:
Fixes #952

bootstrap-salt.sh
justinta: That's a good point. I just changed it there since there is a specific config_freebsd_salt. I'll have to run a few more tests to make sure that will work like I think though. But, I'm assuming it should be fine.

rallytime:
This looks great. Thanks!

Issue #949, opened 2016-09-08
Properly deprecate -G option, don't just remove it.


rallytime:
### Description of Issue/Question

The `-G` option was removed in #945. However, we should still allow the -G option to work by providing a way for it to still work as-is with a warning message that it will be removed in the near future. 

We should make sure this gets fixed before we release a new stable version of the bootstrap script. #945 is not in the current stable release, and we don't want to remove that functionality from underneath people without proper warning and time for people to update their script usage, wrappers, and salt-cloud configs.

ping @vutny and @The-Loeki

rallytime:
Fixed via #958, which displays a warning that the option will be removed in a future release.

Since the original `-G` option only switched the cloning repo url from `git://github.com/saltstack/salt.git` to `https://github.com/saltstack/salt.git`, and the `https` version is now the default, we don't need to do anything other than warn.

PR #958, opened 2016-09-14
Display warning message about deprecating the `-G` option

vutny:

Resolves issue #949


rallytime:
Thanks @vutny!

PR #963, opened 2016-09-23
Fix detection of EPEL repository with not-expired metadata

sp1r:
Hello,

Currently only EPEL repository with expired metadata (repo name is prepended with "!") will be discovered during installation process. This tiny PR corrects regexp to match repositories with not-expired metadata.


vutny (approved):
Nice! Thanks.

rallytime:
Thank you @sp1r!

PR #967, opened 2016-10-04
adds supports for Cumulus Linux

plumbis:
### What does this PR do?

This adds support for Cumulus Linux as a Debian derivative.

Without this PR the boostrap script will not recognize Cumulus Linux as a candidate OS and will fail. After applying this fix the bootstrap will treat Cumulus Linux as the relevant Debian derivative.

this change adds Cumulus Linux 2.x and 3.x as proper Debian derivatives. Cumulus Linux maintains a separate repo that does not contain all of the salt dependencies. The user will still be required to point their system to a global Debian repo for this to execute successfully.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -947,6 +947,8 @@ __gather_linux_system_info() {
             return
         elif [ "${DISTRO_NAME}" = "Raspbian" ]; then
            DISTRO_NAME="Debian"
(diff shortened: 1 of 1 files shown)


rallytime:
Thanks for adding this @plumbis!

sergeyhush:
As of Cumulus Linux 3.2 release, Debian repo will not be needed for Cumulus Linux 3.X

PR #970, opened 2016-10-14
FreeBSD: set _SALT_ETC_DIR directory to match ports

eradman:
### What does this PR do?

Unbreak salt-cloud provisioning for FreeBSD-stable
### Previous Behavior

Minion was started on new host but did not read the config written by salt-bootstrap
### New Behavior

FreeBSD packages use `/usr/local/etc/salt`. Modify the Salt `etc/`
location early so that all config is written to the correct location.

Tested using the following DigitalOcean images:

```
freebsd-11-0-x64
freebsd-10-3-x64
freebsd-10-1-x64
```

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4497,6 +4497,8 @@ __configure_freebsd_pkg_details() {
 }
 
 install_freebsd_9_stable_deps() {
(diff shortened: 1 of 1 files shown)


rallytime:
Thanks @eradman

Issue #985, opened 2016-11-10
Python YAML module may not be present when creating YAML from JSON.


brianholland99:
### Description
When not obtaining Salt from "git" and either "-J" or "-j" arguments are passed, an "import error" occurs if the appropriate Python YAML module was not already present before running the bootstrap.

### Setup using Vagrant
Vagrantfile

```ruby
Vagrant.configure("2") do |config|
  config.vm.define "centos7" do |centos7| 
    centos7.vm.box = "centos/7"
    centos7.vm.hostname = "centos7"
  end
  config.vm.define "ubuntu16" do |ubuntu16|
    ubuntu16.vm.box = "ubuntu/xenial64"
    ubuntu16.vm.hostname = "ubuntu16"
  end
end
```
### Steps to Reproduce Issue

The steps below should reproduce the issue. Replacing "ubuntu16" with "centos7" should show the same thing on CentOS.

Commands:

```sh
vagrant up
vagrant ssh ubuntu16
sudo -i
cd /tmp
curl -o bootstrap_salt.sh -L https://bootstrap.saltstack.com
chmod 0755 bootstrap_salt.sh
/tmp/bootstrap_salt.sh -D -J '{"syndic_master":"192.168.1.124"}' -F -c /tmp -M -S stable
```

The develop branch of salt-bootstrap does the same when replacing the URL above with `https://bootstrap.saltstack.com/develop`.

Debug info:
```
 *  INFO: Running config_salt()
 * DEBUG: Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named yaml
 * ERROR: Python error encountered. This is likely due to passing in a malformed JSON string. Please use -D to see stacktrace.
 * ERROR: Failed to run config_salt()!!!
```

### Versions and Systems
Verified on Centos7.2 and Ubuntu16.04.  I also examined the code on the salt-bootstrap "develop" branch when checking the order of calls.

The error occurs prior to Salt getting installed, so the Salt version is not involved. This was tried on both released and develop version of bootstrap_salt.  Below should be the develop version.
```
ubuntu@ubuntu16:/tmp$ /tmp/bootstrap_salt.sh -v
/tmp/bootstrap_salt.sh -- Version 2016.10.25
```

### Detailed info
Using "git" for obtaining Salt has a side effect of installing the Python YAML library, so that was omitted from the "I Found a Bug" instructions. The various install_XXXX_git_deps() will cause YAML packages to be installed.

[...]

Issue #959, opened 2016-09-14
multiple keys are generated despite invocation with "-i minion1"


oz123:
I am invoking the bootstrap script as the following:

`sudo sh bootstrap_salt.sh -i ${minion_id} stable 2015.8`

The ${minion_id} is replaced by `terraform apply`.

When the EC2 is created it is automatically joined the cluster. However, when I list the keys on the master
I see that the minion has a key which was generated by script with the correct name, but also with the hostname given by amazon:

```
ubuntu@ip-172-16-80-80:~$ sudo salt-key -L
Accepted Keys:
ip-172-16-80-80.ec2.internal  # unwanted key ...
ip-172-16-91-202.ec2.internal    # unwanted key ...
master1 
minion1
Denied Keys:
Unaccepted Keys:
Rejected Keys:
```

```
$ sudo salt --versions-report
Salt Version:
           Salt: 2015.8.12

Dependency Versions:
         Jinja2: 2.7.2
       M2Crypto: Not Installed
           Mako: 0.9.1
         PyYAML: 3.10
          PyZMQ: 14.0.1
         Python: 2.7.6 (default, Jun 22 2015, 17:58:13)
           RAET: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.5
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 1.5
          gitdb: 0.5.4
      gitpython: 0.3.2 RC1
          ioflo: Not Installed
        libgit2: Not Installed
        libnacl: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: 1.2.3
      pycparser: Not Installed
       pycrypto: 2.6.1
         pygit2: Not Installed
   python-gnupg: Not Installed
          smmap: 0.8.2
        timelib: Not Installed

System Versions:
           dist: Ubuntu 14.04 trusty
        machine: x86_64
        release: 3.13.0-91-generic
         system: Ubuntu 14.04 trusty

$ bash bootstrap_salt.sh -v
bootstrap_salt.sh -- Version 2016.08.16

```

cloud/VM provider: AWS EC2 (default Ubuntu 14.04.01 version).

oz123:
So in case somebody else is interested, I made a small script based on https://github.com/saltstack/salt/issues/13141 to remove the dead keys:

```
#!/bin/bash

# salt bootstrap start with 2 keys for each minion which causes
# a havoc ...

[...]

Issue #957, opened 2016-09-14
/etc/salt/minion is not updated


pavlod:
### Description of Issue/Question

Bootstrap script deny to overwrite /etc/salt/minion because of that default config is used, and second salt key generated. As far as salt-master don't know about new key, minion became unmanaged. 
### Setup

```
cat cloud.profiles.d/qastage.conf 
qastage: 
  provider: qastage
  image: ami-...
  size: t2.micro
  ssh_username:
    - centos
    - ec2-user
    - ubuntu
  volumes:
    - { size: 200, device: /dev/xvdd }
  tag: {'Environment': 'qastage', 'Role': 'appserver'}
  del_root_vol_on_destroy: True
  del_all_vol_on_destroy:  True
  script_args: -D
  script: /etc/salt/cloud.deploy.d/bootstrap-salt.sh
  ssh_interface: private_ips
  grains:
    environment: qastage
  sync_after_install: grains

```
### Steps to Reproduce Issue

```
*  INFO: Running config_salt()
 * DEBUG: Moving provided grains file from /tmp/.saltcloud-7dd28065-21e7-4ad7-b78c-dcc03cc7cff6/grains to /etc/salt/grains
 * DEBUG: Moving /tmp/.saltcloud-7dd28065-21e7-4ad7-b78c-dcc03cc7cff6/grains to /etc/salt/grains
 * DEBUG: The passed destination(/etc/salt) is a directory
 * DEBUG: Full destination path is now: /etc/salt/minion
>> * DEBUG: Not overriding /etc/salt/minion with /tmp/.saltcloud-7dd28065-21e7-4ad7-b78c-dcc03cc7cff6/minion**

```

Also tried -F and -K and option, same result:

```
 *  INFO: Running config_salt()
 * DEBUG: Moving provided grains file from /tmp/.saltcloud-939e7b79-5f34-4c1d-bad2-ef1b8808e554/grains to /etc/salt/grains
 * DEBUG: Since BS_KEEP_TEMP_FILES=1 we're copying files instead of moving them
 * DEBUG: Copying /tmp/.saltcloud-939e7b79-5f34-4c1d-bad2-ef1b8808e554/grains to /etc/salt/grains
 * DEBUG: Since BS_KEEP_TEMP_FILES=1 we're copying files instead of moving them
 * DEBUG: The passed destination (/etc/salt) is a directory
 * DEBUG: Full destination path is now: /etc/salt/minion
[...]

PR #987, opened 2016-11-18
Configure Salt components after the dependencies installation

vutny:
### What does this PR do?
It performs configuration of Salt components right after all the dependencies installed/compiled, but before Salt packages.

It fixes issues #985, #959 and #957


rallytime:
This looks great @vutny! Thank you again for your work.

I'll wait for @brianholland99 to test this out. Once he gets a chance to try it, I'll get this merged in.

PR #988, opened 2016-11-24
Fix overwriting Minion config from temporarily directory

vutny:
### What does this PR do?
It fixes the Bootstrap behavior when specified temporarily configuration directory (`-c`) with Minion config there.

Fixes another case for the issue #957 explained below.

### Previous Behavior
If a Minion config file is already present on target system (i.e. `/etc/salt/minion`), it wouldn't be overwritten from temp dir even with `-F` (force overwrite option), but only with `-C` (config-only mode).

### New Behavior
A Minion config file will be overwritten from temp dir if specified either `-F` or `-C` option.


s0undt3ch (approved):

s0undt3ch:
Excellent as always!

PR #993, opened 2016-12-01
Adjust code examples to use the actual bootstrap-salt.sh file name

rallytime:
Matches https://github.com/saltstack/salt/pull/38011 for salt bootstrap tutorial docs.


vutny (approved):
How could we miss that?! :+1: :smile:

PR #994, opened 2016-12-02
Let's detect newly released Salt 2016.11 stable release

vutny:
### What does this PR do?
This PR makes sure that explicit call to bootstrap Salt release 2016.11 by invoking the script like:
```sh
sudo sh bootstrap-salt.sh stable 2016.11
```
will work on Debian and Ubuntu.

This is just a quick solution, the code need to be refactored to detect future and older releases properly and in more generic way, assuming that user knows what does he/she requesting.


rallytime:
Heh. Good call! I forgot to update this before the release. Thanks @vutny!

PR #995, opened 2016-12-10
Drop unnecessary workarounds (creating directories) for early releases

vutny:
### What does this PR do?
It removes no longer needed workarounds for bootstrapping early Salt releases from git, such as creating log and cache directories. Modern Salt versions and installation scripts just take care of it.
I've tested these changes on few randomly picked versions starting from `v2015.5.0` and everything works fine: directories are being created and Salt daemons start normally.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -6150,9 +6150,6 @@ config_salt() {
         exit 0
     fi
 
-    # Create default logs directory if not exists
-    mkdir -p /var/log/salt
-
     return 0
 }
 #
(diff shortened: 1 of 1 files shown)


rallytime:
@vutny Sorry It has taken me a while to get to this, but what about git installations? Just want to make sure we're not accidentally removing something useful.

bootstrap-salt.sh
```diff
@@ -6150,9 +6150,6 @@ config_salt() {
         exit 0
     fi
 
-    # Create default logs directory if not exists
-    mkdir -p /var/log/salt
-
```
vutny, line 6155: Actually these lines were added by me, but it looks like I had some other issue with directories and attempted to workaround it here. As for now I can't recall any details.

Nevertheless, I've tested the script without making logs dir, and it works just fine. All needed directories and log files being successfully created during packages installation and compilation from `git` sources.

bootstrap-salt.sh
```diff
@@ -6464,15 +6461,6 @@ if [ "$_CONFIG_ONLY" -eq $BS_FALSE ]; then
...
 
-# Ensure that the cachedir exists
-# (Workaround for https://github.com/saltstack/salt/issues/6502)
-if [ "$_INSTALL_MINION" -eq $BS_TRUE ]; then
-    if [ ! -d "${_SALT_CACHE_DIR}/minion/proc" ]; then
-        echodebug "Creating salt's cachedir"
-        mkdir -p "${_SALT_CACHE_DIR}/minion/proc"
-    fi
-fi
-
```
vutny, line 6475: That issue seems to be resolved long-long time ago...

vutny (commented):
@rallytime Yeah, I have verified both packages and git installations on popular distros and it appeared that we don't need these workarounds anymore.

bootstrap-salt.sh
rallytime: Ok, sounds good. Thank you for confirming.

PR #997, opened 2016-12-14
fix systemctl path

dmitrievav:
### What does this PR do?
fix systemctl path

not given

### Previous Behavior
It could not check if systemd unit is enabled

### New Behavior
Now it can


vutny (approved):
Good catch! Thanks.

rallytime:
Thanks @dmitrievav!

Issue #976, opened 2016-10-24
Ubuntu 16.10 Unsupported


forty8bits:
### Description of Issue/Question

Ubuntu 16.10 was released earlier this month; salt-bootstrap (understandably) fails to run successfully, but support should likely be added.
### Steps to Reproduce Issue

Simply run `curl -L https://bootstrap.saltstack.com | sudo sh` on a fresh install of any variant of Ubuntu 16.10 (Ubuntu Mate in my case).
### Versions and Systems

Output of `sudo find /etc/ -name \*-release -print -exec cat {} \;`:

```
/etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.10
DISTRIB_CODENAME=yakkety
DISTRIB_DESCRIPTION="Ubuntu 16.10"
/etc/os-release
NAME="Ubuntu"
VERSION="16.10 (Yakkety Yak)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.10"
VERSION_ID="16.10"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="http://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=yakkety
UBUNTU_CODENAME=yakkety
/etc/update-manager/meta-release
# default location for the meta-release file

[METARELEASE]
URI = http://changelogs.ubuntu.com/meta-release
URI_LTS = http://changelogs.ubuntu.com/meta-release-lts
URI_UNSTABLE_POSTFIX = -development
URI_PROPOSED_POSTFIX = -proposed
```

Output of `command lsb_release -a`:

```
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.10
Release:        16.10
Codename:       yakkety
```

rallytime:
Ah, good reminder @forty8bits! Thanks!

rallytime:
@forty8bits The installation for 16.10 has been fixed in #998. There is some clean up to do, as mentioned in the PR, but downloading the latest develop version of the bootstrap script should help you now. :)

forty8bits:
[...]

PR #998, opened 2016-12-17
Handle renaming of gnupg-curl to gnupg1-curl for Ubuntu 16.10

marco-m:
### What does this PR do?

Handle renaming of gnupg-curl to gnupg1-curl for Ubuntu 16.10.

This allows to install salt on Ubuntu 16.10, tested with `bootstrap-salt.sh -M git v2016.11.1`

Fixes at least partially #976

### Previous Behavior

Installing salt failed due to `gnupg-curl` non existing.

### Notes

I still see many errors when the script performs `apt-get install` operations, but the install is successful.

Feel free to edit this PR at will.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1459,10 +1459,9 @@ if ([ "${DISTRO_NAME_L}" != "ubuntu" ] && [ $_PIP_ALL -eq $BS_TRUE ]);then
     exit 1
 fi
 
-# Hack proof of concept. Feel free to refactor properly.
-# Starting from Ubuntu 16.01, gnupg-curl has been renamed to gnupg1-curl.
+# Starting from Ubuntu 16.10, gnupg-curl has been renamed to gnupg1-curl.
 GNUPG_CURL="gnupg-curl"
-if ([ "${DISTRO_NAME_L}" = "ubuntu" ] && [ ${DISTRO_VERSION} = "16.10" ]); then
+if ([ "${DISTRO_NAME_L}" = "ubuntu" ] && [ "${DISTRO_VERSION}" = "16.10" ]); then
     GNUPG_CURL="gnupg1-curl"
 fi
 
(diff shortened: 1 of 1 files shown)


marco-m:
Jenkins fails with 

```
Groovy script failed:

groovy.lang.MissingPropertyException: No such property: msg for class: com.saltstack.jenkins.projects.Bootstrap
```

I don't know what it means :-)

vutny (changes requested):
This is good enough. Thanks @marco-m !

Here is the correct link to the Jenkins build against your PR:
 https://jenkins-testing.saltstack.com/job/bootstrap/job/pr/job/998/job/lint/violations/file/bootstrap-salt.sh/

Got lintian error though. Could you please enclose `${DISTRO_VERSION}` variable in double quotes?

marco-m:
I addressed the lintian error regarding missing double quotes.



Note that, although this PR allows to succesfully install salt master and minion on Ubuntu 16.10, there are still bugs in bootstrap-salt.sh:

Running either `bootstrap-salt.sh` or `bootstrap-salt.sh git v2016.11.1`

will create the wrong APT entry (points to xenial instead than yakkety):

```
cd /etc/apt
find . -type f | xargs grep xenial
sources.list.d/saltstack.list:deb https://repo.saltstack.com/apt/ubuntu/16.10/amd64/latest xenial main
```
I have done these tests on a pristine destroy + up vagrant image.

rallytime:
@marco-m Good find! This same package name issue will likely hit Debian 9 as well (though I have not tried that yet).

rallytime:
@marco-m I made a follow up PR to this with #1001. It makes it so the `xenial` repo doesn't get added but adds a `yakkety` repo instead. That link will currently 404 unfortunately, but I think that is better than have the `xenial` link. If we end up making an official repo link for 16.10, then the bootstrap script will be ready to go.

rallytime:
See issue #976 for more discussion

marco-m:
@rallytime Thanks for the quick reaction. yes I agree, a 404 is better than a wrong link.

PR #1000, opened 2016-12-21
Update supported versions for Fedora: 23, 24, and 25.

rallytime:
The bootstrap script should match the supported versions of SaltStack
more closely, and we don't support versions that are considered EOL by
the OS distributors themselves.

This change updates the bootstrap script to support Fedora 23, 24, and 25.

Since "dnf" replaced "yum" as the package provider in Fedora 23, we no
longer need to set the $FEDORA_PACKAGE_MANAGER variable based on the
version of Fedora. This can simply be "dnf" now for Fedora installs.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1744,7 +1744,7 @@ __check_end_of_life_versions() {
 
         fedora)
             # Fedora lower than 18 are no longer supported
-            if [ "$DISTRO_MAJOR_VERSION" -lt 18 ]; then
+            if [ "$DISTRO_MAJOR_VERSION" -lt 23 ]; then
                 echoerror "End of life distributions are not supported."
                 echoerror "Please consider upgrading to the next stable. See:"
                 echoerror "    https://fedoraproject.org/wiki/Releases"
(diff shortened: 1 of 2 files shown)


bootstrap-salt.sh
```diff
@@ -3308,16 +3292,15 @@ install_fedora_stable_post() {
 }
 
 install_fedora_git_deps() {
-    __fedora_get_package_manager
 
     if [ "$_INSECURE_DL" -eq $BS_FALSE ] && [ "${_SALT_REPO_URL%%://*}" = "https" ]; then
-        $FEDORA_PACKAGE_MANAGER ca-certificates || return 1
+        dnf ca-certificates || return 1
```
vutny: I think the above line should be
```bash
dnf install -y ca-certificates || return 1
```
Right?

It seems that the `install` parameter was accidentally removed in the previous version.

bootstrap-salt.sh
```diff
@@ -3308,16 +3292,15 @@ install_fedora_stable_post() {
...
     fi
 
     install_fedora_deps || return 1
 
     if ! __check_command_exists git; then
-        $FEDORA_PACKAGE_MANAGER install -y git || return 1
+       dnf install -y git || return 1
```
vutny: Wrong indentation here, need 4 spaces instead of 3 :smiley:

vutny (changes requested):
:+1: @rallytime

I have found just a couple of small issues, please take a look at the in-line comments.

bootstrap-salt.sh
rallytime: Ah, yeah. I think you're right. I'm not sure why the `install -y` is missing...

bootstrap-salt.sh
rallytime: Oops! Thanks!

rallytime:
@vutny I fixed those issues that you raised. Thanks for the review!

PR #1001, opened 2016-12-21
Add "yakkety" to ubuntu_codename_translation function

rallytime:
Follow up to #998 and #976

This change will make it so the xenial repo.saltstack.com is not added in `/etc/apt/sources.list.d/saltstack.list` file, but is instead added as yakkety. The 16.10 link to repo.saltstack.com currently will throw 404 errors, but the install will succeed by installing 2016.3.1 from Ubuntu's repo. Git installs will work.

It's possible that a 16.10 repo will be added in the future. When that happens, the bootstrap script will be ready.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1272,7 +1272,11 @@ __ubuntu_codename_translation() {
             fi
             ;;
         "16")
-            DISTRO_CODENAME="xenial"
+            if [ "$_april" ]; then
+                DISTRO_CODENAME="xenial"
+            else
+                DISTRO_CODENAME="yakkety"
+            fi
             ;;
         *)
             DISTRO_CODENAME="trusty"


vutny (commented):
@rallytime As far as I understand, this PR is just a fix for the `16.10` repo support in the future.

And for now we would have Salt installed from SaltStack's "xenial" repo on "yakkety" systems when PR #1002 will be merged, right?

rallytime:
@vutny This PR sets up repo support for 16.10 in the future, yes. It also cleans up the existence of `deb https://repo.saltstack.com/apt/ubuntu/16.10/amd64/latest xenial main` in the `/etc/apt/sources.list.d/` directory, which leaves apt in a broken state.

Salt is not installed from repo.saltstack.com on yakkety right now, regardless of the existence of the yakkety or xenial repo source entry. If you just run `sh bootstrap-salt.sh`, the salt package that gets installed is 2016.3.1, which comes from Ubuntu repos. To get a more recent version, you'll have to use git tags.

My idea here is that a 404 is better than a wrong apt entry for now. The entry will start working if we decide to add a repo file for 16.10 in the future, which is currently being investigated.

vutny (approved):
Thanks for the explanation @rallytime.
These are good and definitely needed changes.

My concern was how that would be compatible with changes in PR #1002, because an another workaround for having Salt installed on 16.10 has been added there... occasionally.

Now I think I understand the impact and everything should go smooth.
I will share the details in the #1002 review.

Issue #972, opened 2016-10-17
Ubuntu 14.10 is not supported


tjyang:
### Description of Issue/Question

I have a few ubuntu 14.10 and they can't be bootstrapped by ./bootstrap_salt.sh
### Setup

(Please provide relevant configs (Be sure to remove sensitive info).)
### Steps to Reproduce Issue

<snipped>
     W: Failed to fetch https://repo.saltstack.com/apt/ubuntu/14.10/amd64/latest/dists/trusty/main/binary-amd64/Packages  HttpError404

```
              W: Failed to fetch https://repo.saltstack.com/apt/ubuntu/14.10/amd64/latest/dists/trusty/main/binary-i386/Packages  HttpError404
```

<snipped>
### Versions and Systems

./bootstrap_salt.sh -v
./bootstrap_salt.sh -- Version 2016.08.16

rallytime:
Hmm...Yeah we don't package 14.10 as it is not part of our [supported distros](http://saltstack.com/wp-content/uploads/2016/08/SaltStack-Supported-Operating-Systems.pdf), but we do list `14.x` on the [README](https://github.com/saltstack/salt-bootstrap/blob/stable/README.rst#ubuntu-and-derivatives) of this repo. 

Thanks for filing this. I wonder if we could update the bootstrap script to try to install/run the 14.04 packages on a 14.10 machine, like we're doing for some of the SLES 11 installations...

rallytime:
@tjyang I discussed this with @thatch45 and we have decided that we won't be fixing this. 14.10 is EOL from Ubuntu and has been for over a year. Our policy is to only support LTS releases from Ubuntu, and 14.10 does not fall under that policy. I have updated the bootstrap script in #1002 to only support 12.04, 14.04, and 16.04 (with best-effort support for 16.10) as discussed in #976.

With those changes, if you run the script on a non-supported ubuntu release, you'll get an error message saying it is not supported and it will exit. I have also updated the README to make this more clear.

rallytime:
The changes mentioned above are available in the current `develop` branch and will be in the next stable release of the bootstrap script. Since we won't be fixing this issue, I am going to close it. Thanks again!

PR #1002, opened 2016-12-21
Updated supported versions for Ubuntu: 12.04, 14.04, 16.04, and 16.10 (best effort)

rallytime:
The bootstrap script should match the supported versions of SaltStack
more closely, and we don't support versions that are considered EOL by
the OS distributors themselves.

This change updates the bootstrap script to support Ubuntu 12.04, 14.04,
16.04, and 16.10 (which the .10 release is a "best-effort" support since
it is a non-LTS OS release from Ubuntu).

@vutny I'd appreciate a review on this one as well if you have a moment.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -253,7 +253,7 @@ Ubuntu and derivatives
 - Linaro 12.04
 - Linux Mint 13/14/16/17/18
 - Trisquel GNU/Linux 6 (based on Ubuntu 12.04)
-- Ubuntu 10.x/11.x/12.x/13.x/14.x/15.x/16.04
+- Ubuntu 12.04/14.04/16.04
 
 
 Other Linux distro
--- a/README.rst
+++ b/README.rst
@@ -257,17 +257,15 @@ Ubuntu and derivatives
 
 Ubuntu Best Effort Support: Non-LTS releases 
 
-This script provides best-effort support for current, non-LTS 
-Ubuntu releases. If package repositories are not provided on 
-[repo.saltstack.com](http://repo.saltstack.com/#ubuntu) for 
-the non-LTS release, the bootstrap script will attempt to 
-install the packages for the most closely related LTS Ubuntu
-release instead. 
+This script provides best-effort support for current, non-LTS Ubuntu releases. If package 
+repositories are not provided on `SaltStack's Ubuntu repository`_ for the non-LTS release, the
+bootstrap script will attempt to install the packages for the most closely related LTS Ubuntu
+release instead.
 
-For example, when installing Salt on Ubuntu 16.10, the bootstrap
-script will setup the repository for Ubuntu 16.04 from 
-[repo.saltstack.com](http://repo.saltstack.com/#ubuntu) and 
-install the 16.04 packages.
+For example, when installing Salt on Ubuntu 16.10, the bootstrap script will setup the repository 
+for Ubuntu 16.04 from `SaltStack's Ubuntu repository`_ and install the 16.04 packages.
+
+.. _`SaltStack's Ubuntu repository`: http://repo.saltstack.com/#ubuntu
 
 
 Other Linux distro
--- a/README.rst
+++ b/README.rst
@@ -216,8 +216,8 @@ Debian and derivatives
 .. note::
 
   Installation of Salt packages on Debian 8 based distribution from repo.saltstack.com repository
-  is currently supported for ``amd64`` (``x86-64``) and ``armhf`` architechtures ONLY. Use ``git``
-  bootstrap mode as described above to install Salt on other architechtures, such as ``i386`` or
+  is currently supported for ``amd64`` (``x86-64``) and ``armhf`` architectures ONLY. Use ``git``
+  bootstrap mode as described above to install Salt on other architectures, such as ``i386`` or
   ``armel``. You also may need to disable repository configuration and allow ``pip`` installations
   by providing ``-r`` and ``-P`` options to the bootstrap script, i.e.:
 
@@ -251,7 +251,7 @@ Ubuntu and derivatives
 
 - Elementary OS 0.2 (based on Ubuntu 12.04)
 - Linaro 12.04
-- Linux Mint 13/14/16/17/18
+- Linux Mint 13/17/18
 - Trisquel GNU/Linux 6 (based on Ubuntu 12.04)
 - Ubuntu 12.04/14.04/16.04
 
--- a/README.rst
+++ b/README.rst
@@ -255,6 +255,20 @@ Ubuntu and derivatives
 - Trisquel GNU/Linux 6 (based on Ubuntu 12.04)
 - Ubuntu 12.04/14.04/16.04
 
+Ubuntu Best Effort Support: Non-LTS releases 
+
+This script provides best-effort support for current, non-LTS 
+Ubuntu releases. If package repositories are not provided on 
+[repo.saltstack.com](http://repo.saltstack.com/#ubuntu) for 
+the non-LTS release, the bootstrap script will attempt to 
+install the packages for the most closely related LTS Ubuntu
+release instead. 
+
+For example, when installing Salt on Ubuntu 16.10, the bootstrap
+script will setup the repository for Ubuntu 16.04 from 
+[repo.saltstack.com](http://repo.saltstack.com/#ubuntu) and 
+install the 16.04 packages.
+
 
 Other Linux distro
 ~~~~~~~~~~~~~~~~~~
--- a/README.rst
+++ b/README.rst
@@ -255,7 +255,8 @@ Ubuntu and derivatives
 - Trisquel GNU/Linux 6 (based on Ubuntu 12.04)
 - Ubuntu 12.04/14.04/16.04
 
-Ubuntu Best Effort Support: Non-LTS releases 
+Ubuntu Best Effort Support: Non-LTS Releases 
+********************************************
 
 This script provides best-effort support for current, non-LTS Ubuntu releases. If package 
 repositories are not provided on `SaltStack's Ubuntu repository`_ for the non-LTS release, the
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2294,12 +2294,7 @@ __enable_universe_repository() {
 
     echodebug "Enabling the universe repository"
 
-    # Ubuntu versions higher than 12.04 do not live in the old repositories
-    if [ "$DISTRO_MAJOR_VERSION" -gt 12 ]; then
-        add-apt-repository -y "deb http://archive.ubuntu.com/ubuntu $(lsb_release -sc) universe" || return 1
-    fi
-
-    add-apt-repository -y "deb http://old-releases.ubuntu.com/ubuntu $(lsb_release -sc) universe" || return 1
+    add-apt-repository -y "deb http://archive.ubuntu.com/ubuntu $(lsb_release -sc) universe" || return 1
 
     return 0
 }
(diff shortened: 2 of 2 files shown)


rallytime:
Refs #972

bootstrap-salt.sh
```diff
@@ -1201,12 +1201,6 @@ __ubuntu_derivatives_translation() {
     # Mappings
     trisquel_6_ubuntu_base="12.04"
     linuxmint_13_ubuntu_base="12.04"
-    linuxmint_14_ubuntu_base="12.10"
-    #linuxmint_15_ubuntu_base="13.04"
-    # Bug preventing add-apt-repository from working on Mint 15:
-    # https://bugs.launchpad.net/linuxmint/+bug/1198751
-
-    linuxmint_16_ubuntu_base="13.10"
     linuxmint_17_ubuntu_base="14.04"
```
vutny, line 1204: Since we are dropping some of Linux Mint releases support, shouldn't we also reflect this in the `README` file?

bootstrap-salt.sh
```diff
@@ -2303,13 +2295,8 @@ __enable_universe_repository() {
     echodebug "Enabling the universe repository"
 
     # Ubuntu versions higher than 12.04 do not live in the old repositories
-    if [ "$DISTRO_MAJOR_VERSION" -gt 12 ] || ([ "$DISTRO_MAJOR_VERSION" -eq 12 ] && [ "$DISTRO_MINOR_VERSION" -gt 04 ]); then
+    if [ "$DISTRO_MAJOR_VERSION" -gt 12 ]; then
         add-apt-repository -y "deb http://archive.ubuntu.com/ubuntu $(lsb_release -sc) universe" || return 1
```
vutny: And since this PR drops support for Ubuntu major releases before the `12.04`,  the above `if` statement is totally unnecessary.
The `precise` release is still in the main archive: http://archive.ubuntu.com/ubuntu/dists/precise/universe/.

bootstrap-salt.sh
```diff
@@ -2303,13 +2295,8 @@ __enable_universe_repository() {
...
-    elif [ "$DISTRO_MAJOR_VERSION" -lt 11 ] && [ "$DISTRO_MINOR_VERSION" -lt 10 ]; then
-        # Below Ubuntu 11.10, the -y flag to add-apt-repository is not supported
-        add-apt-repository "deb http://old-releases.ubuntu.com/ubuntu $(lsb_release -sc) universe" || return 1
-    else
-        add-apt-repository -y "deb http://old-releases.ubuntu.com/ubuntu $(lsb_release -sc) universe" || return 1
     fi
 
     add-apt-repository -y "deb http://old-releases.ubuntu.com/ubuntu $(lsb_release -sc) universe" || return 1
```
vutny: That line need to be removed, there are no LTS releases hosted on http://old-releases.ubuntu.com/ubuntu

bootstrap-salt.sh
```diff
@@ -2385,7 +2364,7 @@ install_ubuntu_deps() {
 }
 
 install_ubuntu_stable_deps() {
-    if ([ "${_SLEEP}" -eq "${__DEFAULT_SLEEP}" ] && [ "$DISTRO_MAJOR_VERSION" -lt 15 ]); then
+    if ([ "${_SLEEP}" -eq "${__DEFAULT_SLEEP}" ] && [ "$DISTRO_MAJOR_VERSION" -lt 16 ]); then
```
vutny: I guess the parentheses could be safely omitted here... :smiley:

bootstrap-salt.sh
```diff
@@ -2476,24 +2450,12 @@ install_ubuntu_stable_deps() {
 install_ubuntu_daily_deps() {
     install_ubuntu_stable_deps || return 1
 
-    if [ "$DISTRO_MAJOR_VERSION" -ge 12 ]; then
-        # Above Ubuntu 11.10 add-apt-repository is in a different package
-        __apt_get_install_noinput software-properties-common || return 1
-    else
-        __apt_get_install_noinput python-software-properties || return 1
-    fi
```
vutny, line 2484: This code here is for a reason, just the expression and comment are little bit wrong.
On Ubuntu 12.04 you need to install `python-software-properties` to get `add-apt-repository` binary:
http://packages.ubuntu.com/precise/all/python-software-properties/filelist

It is not in the `software-properties-common` pkg though: http://packages.ubuntu.com/precise/all/software-properties-common/filelist

The `if` statement should be:
```sh
if [ "$DISTRO_MAJOR_VERSION" -gt 12 ]; then
    ...
```

bootstrap-salt.sh
```diff
@@ -2599,7 +2561,7 @@ install_ubuntu_git() {
 install_ubuntu_stable_post() {
     # Workaround for latest LTS packages on latest ubuntu. Normally packages on
     # debian-based systems will automatically start the corresponding daemons
-    if [ "$DISTRO_MAJOR_VERSION" -lt 15 ]; then
+    if [ "$DISTRO_MAJOR_VERSION" -lt 16 ]; then
         return 0
     fi
```
vutny: Looks like this whole `if` statement is really a legacy of the previous versions. We do not trying to start Salt daemons here in the post install function for Ubuntu, but only enabling them to start on boot. Generally it is done automatically for all packages regardless of distro version, but the later code just double checks that.

I think the above statement could be safely removed.

vutny (changes requested):
This is really awesome @rallytime !

I think we just need to handle few things more carefully :smiley:

bootstrap-salt.sh
rallytime: Yeah I looked around for any documented support for Linux Mint and didn't see any, but I should probably add some.

EDIT: Just kidding. There is Linux Mint support listed in the README. IDK why I didn't find it before. I'll adjust that.

bootstrap-salt.sh
rallytime: Ok, can do. I was wondering about why that was there, but didn't want to remove it if there was a reason.

bootstrap-salt.sh
rallytime: Great! That will be much cleaner! :D

bootstrap-salt.sh
rallytime: Good call.

bootstrap-salt.sh
rallytime: Ah, so this logic wasn't quite right to begin with. Interesting...I'll fix that.

rallytime:
@vutny Thanks for taking such a careful look here! I have made all of the changes you suggested. Please take a look and make sure I didn't miss anything.

bootstrap-salt.sh
```diff
@@ -2430,10 +2404,10 @@ install_ubuntu_stable_deps() {
         # Versions starting with 2015.5.6, 2015.8.1 and 2016.3.0 are hosted at repo.saltstack.com
         if [ "$(echo "$STABLE_REV" | egrep '^(2015\.5|2015\.8|2016\.3|2016\.11|latest|archive\/)')" != "" ]; then
             # Workaround for latest non-LTS ubuntu
-            if [ "$DISTRO_MAJOR_VERSION" -eq 15 ]; then
+            if [ "$DISTRO_MAJOR_VERSION" -eq 16 ] && [ "$DISTRO_MINOR_VERSION" -eq 10 ]; then
                 echowarn "Non-LTS Ubuntu detected, but stable packages requested. Trying packages from latest LTS release. You may experience problems."
-                UBUNTU_VERSION=14.04
-                UBUNTU_CODENAME=trusty
+                UBUNTU_VERSION=16.04
+                UBUNTU_CODENAME=xenial
```
vutny, line 2410: @rallytime This is what I've mentioned in comments for PR #1001 .

The above and later code explicitly says:
**If** we are on 16.10 "yakkety" **then** go configure the SaltStack's repo for 16.04 "xenial", which is:
```
https://repo.saltstack.com/apt/ubuntu/16.04/amd64/latest/SALTSTACK-GPG-KEY.pub
```

This is the "best effort" approach introduced before for Ubuntu non-LTS releases and now it simply works for 16.10 too!
I have tested it and got latest Salt packages successfully installed (and running) from SaltStack's repo on "yakkety", while all other dependencies were fetched from Ubuntu's main and universe repositories.

I think this is really nice, now we have full Ubuntu 16.10 support! That's worth adding to the README :smiley: 

Anyhow, the changes in #1001 still make sense because for now the packages from "xenial" repo are compatible and installable on the "yakkety" system, but that could be changed any moment.

bootstrap-salt.sh
```diff
@@ -2430,10 +2404,10 @@ install_ubuntu_stable_deps() {
         # Versions starting with 2015.5.6, 2015.8.1 and 2016.3.0 are hosted at repo.saltstack.com
         if [ "$(echo "$STABLE_REV" | egrep '^(2015\.5|2015\.8|2016\.3|2016\.11|latest|archive\/)')" != "" ]; then
             # Workaround for latest non-LTS ubuntu
-            if [ "$DISTRO_MAJOR_VERSION" -eq 15 ]; then
+            if [ "$DISTRO_MAJOR_VERSION" -eq 16 ] && [ "$DISTRO_MINOR_VERSION" -eq 10 ]; then
```
vutny: I think we can just use this here:
```bash
if [ "$DISTRO_VERSION" = "16.10" ]; then
```

vutny (changes requested):
@rallytime Again, this PR is absolutely awesome. While dropping lots of legacy stuff, it also brings 16.10 support and fixes some long standing issues. Well done!

I have only one minor suggestion... :smiley: Please look at the in-line comments. Thanks!

bootstrap-salt.sh
rallytime: Oh, haha. Yes. Let's make that simpler.

bootstrap-salt.sh
rallytime: Good call! I'll add a note about that in the README. Thanks!

rallytime:
@vutny I made some more updates per your suggestions. Please take another look and let me know what you think.

README.rst
```diff
@@ -255,6 +255,20 @@ Ubuntu and derivatives
...
+[repo.saltstack.com](http://repo.saltstack.com/#ubuntu) for 
+the non-LTS release, the bootstrap script will attempt to 
+install the packages for the most closely related LTS Ubuntu
+release instead. 
+
+For example, when installing Salt on Ubuntu 16.10, the bootstrap
+script will setup the repository for Ubuntu 16.04 from 
+[repo.saltstack.com](http://repo.saltstack.com/#ubuntu) and 
+install the 16.04 packages.
+
```
vutny, line 270: Maybe it would be better to format links using RsT syntax? Like this:
```
text text `SaltStack corporate repository for Ubuntu`_ some other text

.. _`SaltStack corporate repository for Ubuntu`: http://repo.saltstack.com/#ubuntu
```

This will probably look more clear when the file will be rendered on GitHub.
Also, the line width of 100 chars was used before.

vutny (changes requested):
@rallytime, I know I'm annoying, sorry about that :wink: 
Please take a look at the comment I made.

README.rst
rallytime: OK. I just have been doing too much context switching. I'll fix it.

rallytime:
@vutny No worries. It's always better to get this stuff right on the first try! I appreciate the careful review. Is there anything else you see here that should be fixed before I merge this in?

README.rst
```diff
@@ -251,9 +251,21 @@ Ubuntu and derivatives
 
 - Elementary OS 0.2 (based on Ubuntu 12.04)
 - Linaro 12.04
-- Linux Mint 13/14/16/17/18
+- Linux Mint 13/17/18
 - Trisquel GNU/Linux 6 (based on Ubuntu 12.04)
-- Ubuntu 10.x/11.x/12.x/13.x/14.x/15.x/16.04
+- Ubuntu 12.04/14.04/16.04
+
+Ubuntu Best Effort Support: Non-LTS releases
```
vutny: Okay, this looks like another sub-section to me, so it can be marked accordingly:
```
Ubuntu Best Effort Support: Non-LTS releases
********************************************
```

Maybe we could add some table of contents later.

vutny (changes requested):
@rallytime I promise this is a last one request :smile: 
The changes look perfect to me. After the final formatting cleanup will be made this PR could be merged.
Thank you very much!

rallytime:
Alright, I have made the final formatting change, so I will merge this in. Yay!

I created Issue #1008 to track the addition of TOC for the README file. It's good idea.

PR #1010, opened 2017-01-08
Alpine Linux support #1009

### LINKED ISSUE #1009: Alpine Linux Support

Alpine Linux Support is missing

ek9:
### What does this PR do?

Alpine Linux support

Fixes #1009 

### Previous Behavior

Some minor changes that had to be done. They won't cause problems, but I mention them here:

- `sort --ignore-case` changed to `sort -f`. `--ignore-case` is alias to `-f`, but `sort` in Alpine Linux ONLY supports `-f` parameter.
- instead of `sort --unique` we now use `| uniq` as `sort` in Alpine Linux does NOT support `--unique` argument. `uniq` is part of `coreutils` and is generally available on all OSes.

### New Behavior

- Added Alpine Linux support (tested on 3.5.0)
- `sort -f` is used instead of `sort --ignore-case` for Alpine Linux Support.
- `|uniq` is used instead of `sort --unique` for Alpine Linux Support.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4247,7 +4247,7 @@ install_alpine_linux_stable() {
         __PACKAGES="${__PACKAGES} salt-syndic"
     fi
 
-    apk -U add ${__PACKAGES} || return 1
+    apk -U add "${__PACKAGES}" || return 1
     return 0
 }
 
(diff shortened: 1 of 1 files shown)


rallytime:
Thanks @ek9! Can you take a look at the lint failures and clean those up? Then we should be able to get this in.

https://jenkins-testing.saltstack.com/job/bootstrap/job/pr/job/1010/job/lint/1/violations/file/bootstrap-salt.sh/

ek9:
@rallytime thanks for the link, I fixed the issues. I could not figure out how to access the details (when I click "Details" links near CI fails on the PR it showed me an error page so I did not know what was wrong).

Should be good to merge now.

rallytime:
@ek9 Ah, yeah that link is the `Violations` option on the left sidebar. Thanks for grabbing those!

PR #1015, opened 2017-01-13
Update bootstrap-salt.sh

caelor:
This is a typo fix on line 4302 of boostrap-salt.sh (applies to Alpine Linux) - missing leading / on the redirect to /dev/null.

It squashes a shell error, but makes no functional difference to the install.


rallytime:
Good catch. Thanks @caelor!

PR #1017, opened 2017-01-18
Adding 2016.11 to stable version

ashokrajar:
### What does this PR do?
Make bootstrap to recognise 2016.11 as a stable version.

None

### Previous Behavior
bootstrap script doesn't recognise 2016.11 as a stable version.

### New Behavior
bootstrap script now recognise 2016.11 as a stable version.


vutny (approved):
Awesome! I thought we have that covered, but I haven't tested that explicitly.

FYI @rallytime

rallytime:
Ah, we missed those! Thanks for this @ashokrajar!

PR #1022, opened 2017-01-27
AWS Linux Native Support

bkruger99:
### What does this PR do?
Add in logic to use the new native AWS repo rather than default to RHEL6.
Also added version pinning support and to use the right repos accordingly
Fix packages for AWS so they install the correct versions
Add ability to do specific version pinning with AWS linux that was previously unavailable for some reason.

** Does use some bashisms in the logic check, but figured this was safe since AWS installs bash as /bin/sh.

N/A 

### Previous Behavior
Install RHEL/CentOS 6 versions, including python 2.6

### New Behavior
Does a version check to see if install version is at least 2016.11 and then uses the appropriate repo to do the install.  If installation is 2016.11 or newer, then installer will use the AWS specific repo, which includes using python 2.7 and no longer installing python 2.6 and dependencies.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4443,8 +4443,8 @@ install_amazon_linux_ami_deps() {
         fi
 
         # This should prob be refactored to use __install_saltstack_rhel_repository()
-        # With args passed in to do the right thing.  Reformated to be more like the 
-        # __install_saltstack_rhel_respository() yum file.
+        # With args passed in to do the right thing.  Reformatted to be more like the
+        # amazon linux yum file.
         if [ ! -s "/etc/yum.repos.d/${__REPO_FILENAME}" ]; then
           cat <<_eof > "/etc/yum.repos.d/${__REPO_FILENAME}"
 [saltstack-repo]
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1428,7 +1428,7 @@ __ubuntu_codename_translation
 if ([ "${DISTRO_NAME_L}" != "ubuntu" ] && [ "$ITYPE" = "daily" ]); then
     echoerror "${DISTRO_NAME} does not have daily packages support"
     exit 1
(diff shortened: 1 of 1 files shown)


bkruger99:
One thing that'll need looking into, is post install yum upgrade:


 python27-futures                                                           noarch             3.0.3-2.amzn1                saltstack-repo                                                            29 k

 python27-futures                                                           noarch             3.0.3-1.3.amzn1              installed


Not sure what the differences are, if anything, perhaps the saltstack created one can be excluded.  Will try to get a look to see if it makes a difference.

dmurphy18:
@bkruger99 The Salt Stack version  python27-futures noarch 3.0.3-2.amzn1 shall be removed in a future release, near future.

dmurphy18:
@bkruger99 The latest release Salt 2016.11.2 for Amazon Linux no long provides python27-futures noarch 3.0.3-2.amzn1 which should resolve your issue.  If this solves the issue for you, please consider closing this issue, otherwise let me know of other problems encountered with Amazon Linux.

bkruger99:
@dmurphy18 cool. Then I'll not look into it and I'll leave my patch here as-is if someone chooses to merge it.

rallytime:
@bkruger99 I think this looks like a good addition to the bootstrap script since we'd like to get the repos in the script set up changed when installing 2016.11 packages. Did you figure out what to do about you question above about the yum post install? Just wondering if this is ready to go or needs a little more work.

bkruger99:
Since the salt packaged "futures" was disappearing in 2016.11.2, I decided to just not bother.  This could be a potential issue if someone tries to install 2016.11.1 from the archive.  It's an rpm dependency issue, I guess one could just set the ignore at worst case.   

It's ready to go otherwise.   Future fix might be to just pull the repo rpm from salt directly and install that. (same with RHEL, centos, etc).   Feel free to give it a test run and let me know if something needs to be fixed or not.  I'm just happy to see support so I can try to not install python2.6 on aws and cause massive confusion. 

-b

rallytime:
@bkruger99 I think that sounds good. There are some outstanding issues in the salt repo that are fixed by this PR, so let's get it in. Thank you for your work here.

Issue #1004, opened 2016-12-24
aarch64 (arm64) not supported in 2016.11.1 distribution for Ubuntu?


vielmetti:
### Description of Issue/Question

I'm attempting to bootstrap `salt` on an ARMv8 (aarch64, arm64) machine running Ubuntu 16.04. It doesn't appear that there are packages built for this system, and the attempt to get it running hasn't worked yet (returns errors).

### Setup

aarch64 (ARMv8) 96-core machine at packet.net, running Ubuntu 16.04.

```
# dmesg | grep -i ubuntu
[    0.000000] Linux version 4.4.0-38-generic (buildd@bos01-arm64-006) (gcc version 5.4.0 20160609 (Ubuntu/Linaro 5.4.0-6ubuntu1~16.04.2) ) #57-Ubuntu SMP Wed Sep 7 10:19:14 UTC 2016 (Ubuntu 4.4.0-38.57-generic 4.4.19)
```

### Steps to Reproduce Issue
(Include debug logs if possible, `bootstrap-salt.sh -D`.)

```gpgkeys: no key data found for https://repo.saltstack.com/apt/ubuntu/16.04/arm64/latest/SALTSTACK-GPG-KEY.pub
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
gpg: keyserver communications error: key not found
gpg: keyserver communications error: bad public key
gpg: WARNING: unable to fetch URI https://repo.saltstack.com/apt/ubuntu/16.04/arm64/latest/SALTSTACK-GPG-KEY.pub: bad public key
 * ERROR: Failed to run install_ubuntu_git_deps()!!!
 * DEBUG: Cleaning up the Salt Temporary Git Repository
 * DEBUG: Removing the logging pipe /tmp/bootstrap-salt.logpipe
 * DEBUG: Killing logging pipe tee's with pid(s): 24471
```

### Versions and Systems
(`salt --versions-report`, `bootstrap-salt.sh -v`, system type and version,
cloud/VM provider as appropriate.)

I was able to get an older version of `salt` running, so what's running on that machine now is 2015.8.8

Cloud provider is packet.net.

dmurphy18:
@vielmetti Sorry for the delay in replying to this issue, but most staff have been away over the Christmas/New Year break.  

[...]

PR #1029, opened 2017-02-13
README: describe architectures support for Salt deps on Linux distros

vutny:
### What does this PR do?
It recommends to use `git` bootstrap mode with some options to get Salt running on "foreign" CPU architectures besides `x86_64`. This is proven to work on Debian and Ubuntu, but also may be applicable for other Linux distros, for which Salt does not build packages at all.

This should resolve issue #1004

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,9 +49,7 @@ well as several ways of obtaining the bootstrap script itself.
 
   These examples below show how to bootstrap Salt directly from GitHub or other Git repository.
   Run the script without any parameters to get latest stable Salt packages for your system from
-  `SaltStack corporate repository`_. See first example in the `Install using wget`_ section.
-
-.. _`SaltStack corporate repository`: https://repo.saltstack.com/
+  `SaltStack's corporate repository`_. See first example in the `Install using wget`_ section.
 
 
 Install using curl
@@ -172,8 +170,8 @@ The following examples illustrate how to install Salt via a one-liner.
 
 .. note::
 
-  Warning! These methods do not involve a verification step and assume that the delivered file
-  is trustworthy.
+  Warning! These methods do not involve a verification step and assume that the delivered file is
+  trustworthy.
 
 Any of the example above which use two-lines can be made to run in a single-line
 configuration with minor modifications.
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -200,6 +198,20 @@ Installing the latest develop branch of Salt:
 Supported Operating Systems
 ---------------------------
 
+Since Salt has written in Python, the packages available from `SaltStack's corporate repository`_
```
rallytime: `has written` should be `is written` here.

rallytime (changes requested):
Thanks @vutny - Just one small comment and then I can get this in.

README.rst
vutny: Oh sure, thanks. Need to improve my grammar :)

vutny:
@rallytime , corrected.

rallytime (approved):

rallytime:
Thanks!

README.rst
```diff
@@ -200,6 +198,20 @@ Installing the latest develop branch of Salt:
 Supported Operating Systems
 ---------------------------
 
+Since Salt is written in Python, the packages available from `SaltStack's corporate repository`_
+are CPU architecture independent and could be installed on any hardware supported by Linux kernel.
+However, SaltStack does package Salt's binary dependencies only for ``x86_64`` (``amd64``) and
+``AArch32`` (``armhf``), which is limited for Debian/Raspbian 8 platforms.
```
alexandruavadanii, line 204: I think this might be misleading. armhf is ARMv7 (see [1]), while AArch32 is the 32-bit mode of operation / instruction set for ARMv8.
So both `AArch64` and `AArch32` are actually ARMv8.
[1] https://en.wikipedia.org/wiki/ARM_architecture

vutny: Hi @alexandruavadanii 
In Debian ports, the `armhf` (which is [ARM hard-float](https://wiki.debian.org/ArmHardFloatPort)) only supports 32-bit Armv7 CPU with at least Thumb-2 and VFP3D16, which is ARMv7-A therein referred to as "AArch32".
I'm not an ARM architecture expert, but that seems correct to me.

More info here: https://www.debian.org/ports/arm/

PR #1045, opened 2017-04-13
Allow -R option to work for Debian/Ubuntu

### LINKED ISSUE #968: RFE: Support rpm/deb package download only operation

For some minions sitting behind firewall that have minion ports opened but not http/https connection to the internet, this download only will all them to have smooth upgrade.
### Setup

This is a RFE to add this new feature. 
### Steps to Reproduce Issue

```
#download all the latest minion rpm/deb package to temp directory
bootstrap_salt.sh --downloadonly --out=/tmp 

# then we can copy the pkg file to minions inside firewall and run local pkg install using package managers.
```
### Versions and Systems

./bootstrap_salt.sh -- Version 2016.08.16

rallytime:
### What does this PR do?
Allows the `-R` option to work with Debian-based distributions.

I put together the initial commits for this a while ago, but never finished up and tested the change. After a quick test and a small change, this is ready to go.

Fixes #968

### Previous Behavior
The `-R` option was restricted to only working on RHEL-based distributions.

### New Behavior
The `-R` option has been expanded to also work with Debian/Ubuntu.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -239,6 +239,7 @@ _CUSTOM_REPO_URL="null"
 _CUSTOM_MASTER_CONFIG="null"
 _CUSTOM_MINION_CONFIG="null"
 _QUIET_GIT_INSTALLATION=$BS_FALSE
+_REPO_URL="repo.saltstack.com"
 
 #---  FUNCTION  -------------------------------------------------------------------------------------------------------
 #         NAME:  __usage
@@ -558,8 +559,6 @@ if [ "$_CUSTOM_REPO_URL" != "null" ]; then
         echowarn "Detected -R option. No other repositories will be configured when -R is used. Setting -r option to True."
         _DISABLE_REPOS=$BS_TRUE
     fi
-else
-    _REPO_URL="repo.saltstack.com"
 fi
 
 # Check for any unparsed arguments. Should be an error.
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -549,10 +549,17 @@ if [ "$ITYPE" != "git" ]; then
...
+if [ "$_CUSTOM_REPO_URL" != "null" ]; then
+    _REPO_URL="$_CUSTOM_REPO_URL"
+
+    # Check for -r since -R is being passed. Set -r with a warning.
+    if [ "$_DISABLE_REPOS" -eq $BS_FALSE ]; then
+        echowarn "Detected -R option. No other repositories will be configured when -R is used. Setting -r option to True."
+        _DISABLE_REPOS=$BS_TRUE
+    fi
+else
+    _REPO_URL="repo.saltstack.com"
```
vutny: I think, maybe it's better to put the default `_REPO_URL` value even earlier, where all default settings already defined?
And then just override it here. Just to keep all similar things in one place.

See line 194: https://github.com/saltstack/salt-bootstrap/pull/1045/files#diff-a8778033f6225c137abf05cf8d0bb113L194

bootstrap-salt.sh
```diff
@@ -2922,16 +2929,16 @@ install_debian_8_deps() {
...
-            echoerror "repo.saltstack.com doesn't have packages for your system architecture: $DPKG_ARCHITECTURE."
+            echoerror "$_REPO_URL doesn't have packages for your system architecture: $DPKG_ARCHITECTURE."
             echoerror "Try git installation mode with pip and disable SaltStack apt repository, for example:"
             echoerror "    sh ${__ScriptName} -r -P git v2016.3.1"
 
             exit 1
         fi
 
-        # Versions starting with 2015.5.6, 2015.8.1 and 2016.3.0 are hosted at repo.saltstack.com
+        # Versions starting with 2015.5.6, 2015.8.1 and 2016.3.0 are hosted at $_REPO_URL
```
vutny: Do we really need to put the variable in the comment here? ;-)

vutny (changes requested):
@rallytime This is cool!

I just have couple of minor suggestions here...

s0undt3ch (approved):
@vutny's suggestions are great.

bootstrap-salt.sh
rallytime: Ah, yes, good idea.

bootstrap-salt.sh
rallytime: Heh, nope. I'll fix that.

rallytime:
Thanks for the suggestions @vutny! I have updated this PR with your suggestions.

bootstrap-salt.sh
```diff
@@ -239,6 +239,7 @@ _CUSTOM_REPO_URL="null"
 _CUSTOM_MASTER_CONFIG="null"
 _CUSTOM_MINION_CONFIG="null"
 _QUIET_GIT_INSTALLATION=$BS_FALSE
+_REPO_URL="repo.saltstack.com"
```
vutny, line 242: :+1:

vutny (approved):

PR #1049, opened 2017-04-18
Add ability to install and use a different python version when installing salt

Ch3LL:
### What does this PR do?
This adds two options to the bootstrap script: -x <pythonversion> and -y. Because we are dropping python2.6 support we need an option to install python2.7 for our testing on jenkins for Centos6. These options will install a different version of python and install salt with that python version. This currently only works for a git install of salt.

The user would need to run the following: `salt-bootstrap.sh -y -x python2.7 -q -M git v2016.11.3`. This command will install python27 and use the python2.7 executable to install salt, instead of the system default (python2.6)

I would love to get any feedback to see if this is the right approach.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -3771,9 +3771,9 @@ install_centos_git_deps() {
 install_centos_git() {
     if [ "$DISTRO_MAJOR_VERSION" -eq 5 ]; then
         _PYEXE=python2.6
-    elif [ ${_PY_EXE:='None'} != 'None' ]; then
+    elif [ "${_PY_EXE}" != "" ]; then
         _PYEXE=${_PY_EXE}
-        echoinfo "Using the following python version: ${_PY-EXE} to install salt"
+        echoinfo "Using the following python version: ${_PY_EXE} to install salt"
     else
         _PYEXE=python2
     fi
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -355,12 +355,18 @@ __usage() {
         no ".bak" file will be created as either of those options will force
         a complete overwrite of the file.
     -q  Quiet salt installation from git (setup.py install -q)
+    -x  Changes the python version used to install a git version of salt. Currently
+        this is considered experimental and has only been tested on Centos 6.
+    -y  Installs a different python version on host. Currently this only works
+        with Centos 6 and is considered experimental. This will install the ius
+        repo on the box. This must be used in conjunction with -x <pythonversion>
+        For example: sh bootstrap.sh -y -x python2.7 git v2016.11.3
 
 EOT
 }   # ----------  end of function __usage  ----------
 
 
-while getopts ':hvnDc:g:Gwk:s:MSNXCPFUKIA:i:Lp:dH:ZbflV:J:j:rR:aq' opt
+while getopts ':hvnDc:g:Gyx:wk:s:MSNXCPFUKIA:i:Lp:dH:ZbflV:J:j:rR:aq' opt
 do
   case "${opt}" in
 
@@ -425,6 +431,8 @@ do
     J )  _CUSTOM_MASTER_CONFIG=$OPTARG                  ;;
     j )  _CUSTOM_MINION_CONFIG=$OPTARG                  ;;
     q )  _QUIET_GIT_INSTALLATION=$BS_TRUE               ;;
+    x )  _PY_EXE="$OPTARG"                              ;;
+    y )  _INSTALL_PY="$BS_TRUE"                         ;;
 
     \?)  echo
          echoerror "Option does not exist : $OPTARG"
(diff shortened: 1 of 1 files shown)


rallytime:
Hi @Ch3LL - can you clean up these lint errors? 

https://jenkins-testing.saltstack.com/job/bootstrap/job/pr/job/1049/job/lint/1/violations/file/bootstrap-salt.sh/

Ch3LL:
I'm looking at them now. I can see the lint checker is running this: `shellcheck -s sh -f checkstyle /var/lib/jenkins/jobs/bootstrap/jobs/pr/jobs/1049/jobs/lint/workspace/bootstrap-salt.sh` but when I run this commadn on my box I don't get all the same warnings. I am guessing there is some sort of file with definitions I need to use? I just want to be able to run this lint checker on my box before I submit the fix to make sure i got everything i needed. Is there a way to run this locally on my box?

rallytime:
We discussed this offline, but for completeness, the `runtests.py` file in the `tests` directory of this repo (run with the `-L` option) will do what you need: `python tests/runtests.py -L`.

Ch3LL:
Okay thanks for your help @rallytime those lint errors should be taken care of now..errrr or not. let me check them

bootstrap-salt.sh
```diff
@@ -1069,6 +1081,38 @@ __gather_linux_system_info() {
...
+    if [ ${_PY_EXE:='None'} = 'None' ]; then
+        echoerror "Must specify -x <pythonversion> with -y to install a specific python version"
+        exit 1
+    fi
+
+    py_pkg_v=$(echo "$_PY_EXE" | sed -r "s/\.//g")
+    __PACKAGES="${py_pkg_v} ${py_pkg_v}-pip ${py_pkg_v}-devel gcc"
+
+    #Install new python version
+    __PYTHON_REPO_URL="https://centos${DISTRO_MAJOR_VERSION}.iuscommunity.org/ius-release.rpm"
```
rallytime: If we're hard-coding this function to use `centos` repositories, then we should probably do some error checking and then warn/exit if we're running on a distribution that isn't supported.

bootstrap-salt.sh
```diff
@@ -1069,6 +1081,38 @@ __gather_linux_system_info() {
...
+
+    if [ $_DISABLE_REPOS -eq $BS_FALSE ]; then
+        echoinfo "Installing IUS repo"
+        yum install -y ${__PYTHON_REPO_URL} || return 1
+    fi
+
+    echoinfo "Installing ${_PY_EXE}"
+    yum install -y ${__PACKAGES} || return 1
+
+    _PIP_PACKAGES="tornado PyYAML msgpack-python jinja2 pycrypto zmq"
```
rallytime: In other functions we check if pip is allowed (this is done by passing `-P` at the CLI) and I think we should probably do this here as well. Check out the `install_fedora_git_deps()` function for a good example of a centos-based example.

rallytime (changes requested):
I had a couple of comments/questions about this, but overall I like this approach. It looks like in addition to those changes, there are still a couple of lint errors:

https://jenkins-testing.saltstack.com/job/bootstrap/job/pr/job/1049/job/lint/6/violations/file/bootstrap-salt.sh/

I think a review from @s0undt3ch would probably be good here as well to see if there is something I am missing.

bootstrap-salt.sh
s0undt3ch: :+1: on the review comment

bootstrap-salt.sh
s0undt3ch: Also a spot on review comment.

s0undt3ch (commented):
Wouldn't it be better if the saltstack distro packages repo had python 2.7 available? This way we would only depend on our own repo. (Stage 2?)

Ch3LL:
@s0undt3ch I agree python27 packages on our repo would be better. @cachedout I can't remember is there plans to provide a python27 packages alongside the nitrogen release?
Even if there are plans I think we merge this with that repo for now until those packages can be created so we can get the cent6 tests running momentarily, and add the new repo here later. 

I will work on the other suggestions right now. Thanks for the review :)

s0undt3ch:
Yeah, no need to block the merge of this PR because of the packages.

cachedout:
Let's bring @dmurphy18 into this to discuss the feasibility of that packaging.

Ch3LL:
Okay this is ready for final review @rallytime I added a _pip_install_pkg function to hanle some more error handling around pip

bootstrap-salt.sh
```diff
@@ -3671,8 +3759,12 @@ install_centos_git_deps() {
         __PACKAGES="${__PACKAGES} python-libcloud"
     fi
 
-    # shellcheck disable=SC2086
-    __yum_install_noinput ${__PACKAGES} || return 1
+    if [ "${_INSTALL_PY}" = "${BS_TRUE}" ]; then
+        __install_python_and_deps || return 1
+    else
+        # shellcheck disable=SC2086
+        __yum_install_noinput ${__PACKAGES} || return 1
```
rallytime, line 3770: Is this supposed to be in the `else` statement here? It seems like we've done a lot of work up to this point to add many `python-*` packages to the `__PACKAGES` list, and now they won't be installed if you're installing a new version of Python. Are we passing that `__PACKAGES` chain to `__install_python_and_deps` function that I am not seeing?

rallytime (changes requested):
Hi @Ch3LL - I just have one more question here with the changes you've made.

bootstrap-salt.sh
Ch3LL: There is not a repo that provides the appropriate python packages for the dependencies. For example the repository added does not provide python27-jinja2,python2-msgpack, etc. I only have access to the python26 packages, So I have had to install all the dependencies using pip that you can see in the `__install_python_and_deps` function. I did notice in those packages that I am missing the check to install libcloud if `-L` is passed. So I have added that in an extra commit for that. (although I haven't been able to test it because the epel repo is down right now)

rallytime: Ok that sounds good. Thanks!

rallytime (approved):

rallytime:
That lint error is coming from a different pull request, so I'll merge this now.

PR #1054, opened 2017-04-21
Fix `git` bootstrap mode for CentOS

vutny:
### What does this PR do?
It fixes installation from GitHub on CentOS caused by missing `epel` repo during early `yum` operations.

### Previous Behavior
```console
 *  INFO: Running install_centos_git_deps()
Loaded plugins: fastestmirror, ovl

Error getting repository data for epel, repository not found
 * ERROR: Failed to run install_centos_git_deps()!!!
```

### New Behavior
```console
 *  INFO: Salt installed!
 * DEBUG: Cleaning up the Salt Temporary Git Repository
```

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -3732,9 +3732,7 @@ install_centos_stable_post() {
 }
 
 install_centos_git_deps() {
-    if [ $_DISABLE_REPOS -eq $BS_FALSE ]; then
-        __install_epel_repository || return 1
-    fi
+    install_centos_stable_deps || return 1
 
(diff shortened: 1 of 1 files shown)


rallytime:
Thanks for grabbing this @vutny. However, there is a conflict here due to the changes in #1049. I don't think this change is needed anymore, though either way will work since both set up the epel repo in the appropriate place.

vutny:
@rallytime I have rebased my changes against latest develop.

I'm confident that my fix is still applicable because having EPEL configured is ultimately not enough for successful bootstrapping across all variants of RHEL distros: some deps are still served from SaltStack's corp repo. Also, the `install_centos_stable_deps()` function holds the logic for system upgrade and extra pkgs installation options, so it is vital to call it from `install_centos_git_deps()` most early.

rallytime:
Yeah that sounds good. Thanks!

PR #1059, opened 2017-04-27
Alpine: fix adding, checking and running Salt Syndic in stable mode

vutny:
### What does this PR do?
It enables starting and checking of `salt-syndic` service during `stable` bootstrapping on Alpine Linux.

Alpine does not provide init script in their `salt-syndic` package.

### Previous Behavior
The Salt Syndic was installed successfully, but was disabled to start after bootstrapping and reboot.  

### New Behavior
The missing init script for Syndic service will be obtained from GitHub, installed and enabled. The `salt-syndic` will be started if not explicitly disabled by the `-X` option.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4311,11 +4311,18 @@ install_alpine_linux_stable_deps() {
     fi
 
     apk update
+
+    # Get latest root CA certs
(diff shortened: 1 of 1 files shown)


rallytime (approved):
👍

PR #1060, opened 2017-04-27
Archlinux must always update

gtmanfred:
### What does this PR do?
Pacman does not manage dependencies, so there is no way to always rely
on -Sy, you must use -Su afterwards, or broken dependencies could
abound.

The only time this is not always the case is with archlinux-keyring.

This removes the ability to configure arch to not update the whole
system, because of its rolling release nature, anytime the repositories
are refreshed, you must do a system upgrade.

Fixes saltstack/zh#1210
Fixes saltstack/zh#1263


rallytime (approved):

rallytime:
Thanks @gtmanfred!

Issue #1050, opened 2017-04-19
Cannot install standalone minion on kde neon


EHJ-52n:
# Description of Issue

## Setup
- *Script start:*
   ```
    *  INFO: Running version: 2017.01.10
    *  INFO: Executed by: sh
    *  INFO: Command line: './bootstrap_salt.sh '
   
    *  INFO: System Information:
    *  INFO:   CPU:          GenuineIntel
    *  INFO:   CPU Arch:     x86_64
    *  INFO:   OS Name:      Linux
    *  INFO:   OS Version:   4.8.0-46-generic
    *  INFO:   Distribution: Debian 16.04
   ```
- `lsb_release -a`:
   ```
   No LSB modules are available.
   Distributor ID: neon
   Description:    KDE neon User Edition 5.9
   Release:        16.04
   Codename:       xenial
   ```

## Steps to Reproduce Issue
1. Install latest kde neon
1. Open terminal
1. Execute the following two commands (Documentation: [Salt Masterless Quickstart](https://docs.saltstack.com/en/latest/topics/tutorials/quickstart.html#salt-masterless-quickstart)):
   ```
   $ curl -L https://bootstrap.saltstack.com -o bootstrap_salt.sh
   $ sudo sh bootstrap_salt.sh
   ```

## Versions and Systems
-  *Salt version*:
   ```
   $ salt --versions-report
   The program 'salt' is currently not installed. You can install it by typing:
   sudo apt install salt-master
   ```
- *Salt bootstrap*:
   ```
   $ sudo sh ./bootstrap_salt.sh -v
   ./bootstrap_salt.sh -- Version 2017.01.10
   ```
- *salt-call*:
   ```
   $ salt-call --version
   salt-call 2015.8.8 (Beryllium)
   ```

## Solution

```
--- bootstrap_salt.sh   2017-04-19 21:07:17.731136826 +0200
+++ bootstrap_salt_fixed.sh     2017-04-19 20:53:02.658094547 +0200
@@ -959,6 +959,8 @@
            DISTRO_NAME="Debian"
         elif [ "${DISTRO_NAME}" = "Cumulus Linux" ]; then
             DISTRO_NAME="Debian"
+        elif [ "${DISTRO_NAME}" = "neon" ]; then
+            DISTRO_NAME="Debian"
         fi
         rv=$(lsb_release -sr)
         [ "${rv}" != "" ] && DISTRO_VERSION=$(__parse_version_string "$rv")
```

[...]

PR #1062, opened 2017-04-28
Properly detect all supported Debian GNU/Linux derivatives

vutny:
### What does this PR do?
It fixes detection mechanism for Debian OS derivatives and properly outputs distribution.  Updated README.

The issue has been raised in #1050.

### Previous Behavior
The script reports it runs on "pure" Debian in case of Cumulus and Raspbian.

### New Behavior
The script reports real distribution name and uses appropriate repo URL for base Debian version.


rallytime (approved):
Thanks for grabbing this @vutny

Issue #778, opened 2016-03-20
Manjaro Linux not supported/cannot install


bissei:
Hi,

Cannot install Salt to Manjaro. Please, add support for this. Thanks!

```
NAME="Manjaro Linux"
ID=manjaro
PRETTY_NAME="Manjaro Linux"
ANSI_COLOR="1;32"
HOME_URL="http://www.manjaro.org/"
SUPPORT_URL="http://www.manjaro.org/"
BUG_REPORT_URL="http://bugs.manjaro.org/"
/etc/manjaro-release
Manjaro Linux
/etc/lsb-release
DISTRIB_ID=ManjaroLinux
DISTRIB_RELEASE=15.12
DISTRIB_CODENAME=Capella
DISTRIB_DESCRIPTION="Manjaro Linux"

LSB Version:    n/a
Distributor ID: ManjaroLinux
Description:    Manjaro Linux
Release:    15.12
Codename:   Capella

 *  INFO: /bin/sh ./install_salt.sh -- Version 2015.11.09

 *  INFO: System Information:
 *  INFO:   CPU:          GenuineIntel
 *  INFO:   CPU Arch:     x86_64
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   4.4.5-1-MANJARO
 *  INFO:   Distribution: ManjaroLinux 15.12

 *  INFO: Installing minion
 *  INFO: Found function daemons_running
 * ERROR: No dependencies installation function found. Exiting...
```

jfindlay:
@bissei, thanks for reporting.

briancappello:
@bissei There's an easy work-around to get this working. You just gotta fool the script into thinking you're running Arch (because, in actuality, you are):

``` bash
sudo sed -i 's/ManjaroLinux/Arch/' /etc/lsb-release
sudo sh install_salt.sh [...]
sudo sed -i 's/Arch/ManjaroLinux/' /etc/lsb-release
```

rallytime:
This should be fixed with #1063. As such, I will close this. Thanks!

PR #1063, opened 2017-05-02
Add Manjaro as Arch derivative

luthes:
### What does this PR do?
Simple fix that allows the bootstrap to run on Manjaro.

Issue #778 

### Previous Behavior
Bootstrap failed to recognize lsb_release -si "ManjaroLinux", exited.

### New Behavior
Bootstrap recognizes "ManjaroLinux" as "Arch Linux", and installs the salt minion.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -972,6 +972,8 @@ __gather_linux_system_info() {
             DISTRO_NAME="Oracle Linux"
         elif [ "${DISTRO_NAME}" = "AmazonAMI" ]; then
             DISTRO_NAME="Amazon Linux AMI"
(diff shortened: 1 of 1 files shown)


vutny (approved):
I think this is fine for now.

But for the future development I believe we should not mask the real distro with its "parent". See how we handle Debian and Ubuntu derivatives. That's not perfect though.
It would be better to avoid hard-coding of numerous fork names just to support them. We need another way to detect base distribution properly.

Anyway, thanks for submitting this.

rallytime (approved):

rallytime:
@vutny Yes, that could certainly be improved. But this is good for now to get it working and the other pieces can be worked out another day. Thank you for pointing that out!

@luthes Thanks for fixing this. :)

Issue #1067, opened 2017-05-05
Scientific Linux 6.9 bootstrap and yum repo install and upgrade issues


rtournoy:
### Description of Issue/Question
yum upgrade fails with https://repo.saltstack.com/yum/redhat/6.9/x86_64/latest/repodata/repomd.xml: [Errno 14] PYCURL ERROR 22 - "The requested URL returned error: 404 Not Found"

Indeed directory 6.9 doesn't exist : https://repo.saltstack.com/yum/redhat/

Unfortunately it is breaking the whole distro update process and I have to disable the salt repository to upgrade SL 6.9 servers.

Please see issue #833 because the same problem was fixed here from SL 6.5 to SL 6.8, until the SL 6.9 was released, now the new directory is HTTP 404.

This is probably only a matter of a symlink here : 
https://repo.saltstack.com/yum/redhat/

Thanks.

### Setup
Scientific Linux 6.9 and Salt: 2016.11.3
SL 6.9 is a REDHAT / CENTOS like distro.

### Steps to Reproduce Issue
- yum -y install https://repo.saltstack.com/yum/redhat/salt-repo-2016.11-1.el7.noarch.rpm
- yum upgrade -y 
- https://repo.saltstack.com/yum/redhat/6.9/x86_64/latest/repodata/repomd.xml: [Errno 14] PYCURL ERROR 22 - "The requested URL returned error: 404 Not Found"

### Versions and Systems
Salt Version:
           Salt: 2016.11.3
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8.1
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.20.2
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
         pygit2: Not Installed
         Python: 2.6.6 (r266:84292, Aug 18 2016, 12:06:14)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 14.5.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.5
 
System Versions:
           dist: redhat 6.9 Carbon
[...]

PR #1068, opened 2017-05-05
Fix configuring SaltStack's repo URL for RHEL variants

vutny:
### What does this PR do?
It removes usage of special `$releasever` variable recognized by `yum` and brings in `${DISTRO_MAJOR_VERSION}` to set the repo URL correctly.

#1067 

### Previous Behavior
There was a workaround only for Cloud Linux 7 which has `$releasever` set to `7.x`.

### New Behavior
The workaround from above set for all RHEL variants to prevent dependency on correct `$releasever` and having symlinks for each minor distro version in the repo.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -3605,13 +3605,6 @@ _eof
         __rpm_import_gpg "${fetch_url}${gpg_key}" || return 1
     fi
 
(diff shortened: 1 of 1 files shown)


rallytime (approved):
Very nice. This will help things be more consistent.

Issue #1065, opened 2017-05-03
CentOS 5 stable install fails


grandmasterv:
### Description of Issue/Question
I have tried to run the bootstrap script on a few CentOS 5 systems (not just the Vagrant box listed below) and it keeps giving the same error in the same place.  I'm just setting up a Vagrant box and want to use their built in Salt provisioning configuration but that relies on the bootstrap script.

I can successfully install a Salt Minion on these same systems using the manual CentOS 5 instructions here:
https://repo.saltstack.com/#rhel

### Setup
Vagrant box:
https://atlas.hashicorp.com/bento/boxes/centos-5.11

Vagrant Salt Provisioning:
https://www.vagrantup.com/docs/provisioning/salt.html

### Steps to Reproduce Issue
```
[vagrant@vince-centos5 salt]$ sudo ./bootstrap-salt.sh -F -D -c /tmp stable
 *  INFO: Running version: 2017.01.10
 *  INFO: Executed by: /bin/sh
 *  INFO: Command line: './bootstrap-salt.sh -F -D -c /tmp stable'

 *  INFO: System Information:
 *  INFO:   CPU:          GenuineIntel
 *  INFO:   CPU Arch:     x86_64
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   2.6.18-398.el5
 *  INFO:   Distribution: CentOS 5.11

 * DEBUG: Binaries will be searched using the following $PATH: /usr/bin:/bin
 *  INFO: Installing minion
 * DEBUG: install_centos_5_stable_deps not found....
 * DEBUG: install_centos_5_11_stable_deps not found....
 * DEBUG: install_centos_5_deps not found....
 * DEBUG: install_centos_5_11_deps not found....
 *  INFO: Found function install_centos_stable_deps
 * DEBUG: DEPS_INSTALL_FUNC=install_centos_stable_deps
 * DEBUG: config_centos_5_stable_salt not found....
 * DEBUG: config_centos_5_11_stable_salt not found....
 * DEBUG: config_centos_5_salt not found....
 * DEBUG: config_centos_5_11_salt not found....
 * DEBUG: config_centos_stable_salt not found....
 * DEBUG: config_centos_salt not found....
 *  INFO: Found function config_salt
 * DEBUG: CONFIG_SALT_FUNC=config_salt
 * DEBUG: preseed_centos_5_stable_master not found....
 * DEBUG: preseed_centos_5_11_stable_master not found....
[...]

PR #1070, opened 2017-05-10
Declare End-of-Life for RHEL 5 and its variants

vutny:
### What does this PR do?
It disables bootstrapping on RHEL/CentOS/CL/OL/SL **5** distros and removes workarounds used to get Salt running on them. The README has been updated as well.

The upstream RHEL release 5 reached EOL on March 31, 2017. The issue has been raised in #1065.

### Previous Behavior
Attempt to install Salt on RHEL 5 fails due to missing EPEL support.

### New Behavior
The error message about EOL policy will be displayed when the script would be executed on RHEL 5.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -235,12 +235,13 @@ Debian and derivatives
 Red Hat family
 ~~~~~~~~~~~~~~
 
-- Amazon Linux
-- CentOS 5/6/7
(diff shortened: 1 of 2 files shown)


rallytime (approved):
Very nice. Thank you for cleaning up this EOL support for RHEL 5.

Issue #1071, opened 2017-05-17
[DEVELOP] Amazon git install fails due to wrong repo.saltstack.com path


Ch3LL:
### Description of Issue/Question
Currently on develop (3750240fcf099ab836c3678e9cc4f9be99e154ad) if you try to use the bootstrap script as follows: `sh /tmp/bootstrap-salt.sh -P git v2016.11.4` it fails because it tries to add the repo.saltstack.com as follows: 

```
https://repo.saltstack.com/yum/redhat/6/x86_64/v2016.11.4/repodata/repomd.xml: [Errno 14] HTTPS Error 404 - Not Found
Trying other mirror.
To address this issue please refer to the below knowledge base article 
```

As you can see it is trying to add `x86_64/v2016.11.4/repodata/repomd.xml` where v2016.11.4 does not exist on the repo. I think we need to fix this by one of two ways, just add that to be latest when using git install or adding the archive link. I think this was caused by this pr: https://github.com/saltstack/salt-bootstrap/pull/1032 

I think another issue with this is it is using the redhat link instead of the amazon link. After salt version 2016.11 i believe it should be using the amazon native packages. 

I will dive more into this tomorrow. 

### Setup
Amazon 2015.03

dmurphy18:
@Ch3LL That url is incorrect for repo.saltstack.com
 For a specific version it should read as follows I believe

https://repo.saltstack.com/yum/redhat/6/x86_64/archive/2016.11.4/repodata/repomd.xml

Ch3LL:
Right. That is the problem. I need to update the bootstrap so it points to the archive path or latest. Previously we were pointing to latest if user was installing a specific version for amazon. Looking at this now. I need to see if we are pointing to latest in other parts of the script.

Ch3LL:
Okay I need to revisit this tomorrow.

I have this: 

```diff --git a/bootstrap-salt.sh b/bootstrap-salt.sh
index 3f7a769..cf796d9 100755
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4461,7 +4461,11 @@ install_amazon_linux_ami_deps() {
     # Shim to figure out if we're using old (rhel) or new (aws) rpms.
     _USEAWS=$BS_FALSE
 
-    repo_rev="$(echo "${GIT_REV:-$STABLE_REV}"  | sed 's|.*\/||g')"
+    if [ "$ITYPE" = "stable" ]; then
[...]

PR #1073, opened 2017-05-22
Allow amazon to work with python2.7 on installs over 2016.11

Ch3LL:
### What does this PR do?

- Fixes issue https://github.com/saltstack/salt-bootstrap/issues/1071
- Also now uses python2.7 when version is above 2016.11, which is the same behavior with the new amazon packages in 2016.11 and above in repo.saltstack.com.
- Changes the default repo to latest if not stable install. I believe this is the approach currently to take since its the same behavior in others ares of the bootstrap script.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4461,7 +4461,11 @@ install_amazon_linux_ami_deps() {
     # Shim to figure out if we're using old (rhel) or new (aws) rpms.
     _USEAWS=$BS_FALSE
 
-    repo_rev="$(echo "${GIT_REV:-$STABLE_REV}"  | sed 's|.*\/||g')"
+    if [ "$ITYPE" = "stable" ]; then
+        repo_rev="$(echo "${STABLE_REV}"  | sed 's|.*\/||g')"
+    else
+        repo_rev="latest"
+    fi
 
     if echo "$repo_rev" | egrep -q '^(latest|2016\.11)$'; then
        _USEAWS=$BS_TRUE
@@ -4538,6 +4542,13 @@ install_amazon_linux_ami_git_deps() {
         yum -y install ca-certificates || return 1
     fi
 
+    PIP_EXE='pip'
+    if [ "$(echo "$GIT_REV" | egrep -o '2[0-9]{3}')" -ge 2016 ]; then
+        [ $(which pip2.7) ] || /usr/bin/easy_install-2.7 pip || echoerror "Could not install pip2.7"
+        PIP_EXE='/usr/local/bin/pip2.7'
+        _PY_EXE='python2.7'
+    fi
+
     install_amazon_linux_ami_deps || return 1
 
     ENABLE_EPEL_CMD=""
@@ -4564,7 +4575,7 @@ install_amazon_linux_ami_git_deps() {
         # We're on the develop branch, install whichever tornado is on the requirements file
         __REQUIRED_TORNADO="$(grep tornado "${_SALT_GIT_CHECKOUT_DIR}/requirements/base.txt")"
         if [ "${__REQUIRED_TORNADO}" != "" ]; then
-            __PACKAGES="${__PACKAGES} python-tornado"
+            __PACKAGES="${__PACKAGES} ${pkg_append}-tornado"
         fi
     fi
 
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -4461,7 +4461,11 @@ install_amazon_linux_ami_deps() {
     # Shim to figure out if we're using old (rhel) or new (aws) rpms.
     _USEAWS=$BS_FALSE
 
-    repo_rev="$(echo "${GIT_REV:-$STABLE_REV}"  | sed 's|.*\/||g')"
+    if [ "$ITYPE" = "stable" ]; then
+        repo_rev="$(echo "${STABLE_REV}"  | sed 's|.*\/||g')"
+    else
+        repo_rev="latest"
+    fi
```
vutny: I'm not sure why we need this `repo_rev` variable at all, since `STABLE_REV` will always contain `latest` during `git` bootstrap mode. https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L533

bootstrap-salt.sh
```diff
@@ -4538,6 +4542,13 @@ install_amazon_linux_ami_git_deps() {
         yum -y install ca-certificates || return 1
     fi
 
+    PIP_EXE='pip'
+    if [ "$(echo "$GIT_REV" | egrep -o '2[0-9]{3}')" -ge 2016 ]; then
```
vutny: Since `GIT_REV` could be any SHA1 commit hash, I think this statement could really make false positives...
Maybe it would be better just to check if Python 2.7 is available and go further with it?

bootstrap-salt.sh
```diff
@@ -4538,6 +4542,13 @@ install_amazon_linux_ami_git_deps() {
...
+        [ $(which pip2.7) ] || /usr/bin/easy_install-2.7 pip || echoerror "Could not install pip2.7"
```
vutny: I guess here the script should die, right?

vutny (changes requested):
I have couple of suggestions to make things little bit clearer.

bootstrap-salt.sh
Ch3LL: I think you are right that repo_rev isn't actually required but I was trying to use the same variable used throughout the code. This repo_rev is also used in other parts of the bootstrap script. You can see with my latest commit I removed the if logic because we don't even really need that since as you pointed out it will be latest, and we can keep the original sed command to remove archive if its present. Are you okay with that approach?

bootstrap-salt.sh
Ch3LL: Thanks for this suggestion adn catching this I like your approach much better. I added in my latest commit.

bootstrap-salt.sh
Ch3LL: Yes thank you I made an assumption that echoerror would return 1 for me without actually looking. Thanks for that catch :) And seriously thanks for all your other suggestions. I really appreciate it :)

bootstrap-salt.sh
vutny: Yeah, that's fine. I just wanted to highlight that it seems like `$repo_rev` variable was used to guess the right path for the repo URL. This actually should honor exact version pinning for stable installations, e.g. `2016.11.4` and the feature doesn't work for Amazon Linux for now. But that's another issue :-)

bootstrap-salt.sh
vutny: I have another small suggestion, will comment below...

bootstrap-salt.sh
```diff
@@ -4538,6 +4538,13 @@ install_amazon_linux_ami_git_deps() {
         yum -y install ca-certificates || return 1
     fi
 
+    PIP_EXE='pip'
+    if [ $(command -v python2.7) ]; then
+        [ $(which pip2.7) ] || /usr/bin/easy_install-2.7 pip || return 1
```
vutny: We have a special function: `__check_command_exists` (https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L118)
Quite useful! :smiley:

Ch3LL: OH so much easier! Thanks I will make that change right now.

Ch3LL: k added

rallytime (approved):

Issue #1074, opened 2017-05-22
SLES12 fails on step install_suse_check_services() when running git install


Ch3LL:
### Description of Issue/Question
When installing a git install on sles12 I am seeing it fail on step install_suse_check_services()

```
13:33:12                        *  INFO: Running install_suse_check_services()
13:33:12                        * ERROR: Failed to run install_suse_check_services()!!!
```

I am simply running `sh -x bootstrap.sh -P`. I am seeing this behavior on develop and the currently live available bootstrap script. 

When running with `sh -x` I see the following in the step `install_suse_12_stable_post`:

```
 *  INFO: Running install_suse_12_stable_post()
+ install_suse_12_stable_post
+ '[' 1 -gt 1 ']'
+ for fname in api master minion syndic
+ '[' api = api ']'
+ '[' 0 -eq 0 ']'
+ continue
+ for fname in api master minion syndic
+ '[' master = api ']'
+ '[' master = master ']'
+ '[' 0 -eq 0 ']'
+ continue
+ for fname in api master minion syndic
+ '[' minion = api ']'
+ '[' minion = master ']'
+ '[' minion = minion ']'
+ '[' 1 -eq 0 ']'
+ '[' minion = syndic ']'
+ '[' -f /bin/systemctl ']'
+ curl -L https://github.com/saltstack/salt/raw/develop/pkg/salt-minion.service -o /usr/lib/systemd/system/salt-minion.service
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
^M  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0^M100   146  100   146    0     0   2952      0 --:--:-- --:--:-- --:--:--  2979
^M100   202  100   202    0     0   2365      0 --:--:-- --:--:-- --:--:--  2365
+ continue
+ for fname in api master minion syndic
+ '[' syndic = api ']'
+ '[' syndic = master ']'
+ '[' syndic = minion ']'
+ '[' syndic = syndic ']'
+ '[' 0 -eq 0 ']'
+ continue
+ return 0
+ '[' 0 -ne 0 ']'
+ '[' install_suse_check_services '!=' null ']'
+ '[' 0 -eq 0 ']'
+ echoinfo 'Running install_suse_check_services()'
+ printf '\033[1;32m *  INFO\033[0m: %s\n' 'Running install_suse_check_services()'
```

[...]

PR #1075, opened 2017-05-22
ensure sles12 enables services with stable installs

Ch3LL:
### What does this PR do?
Fixes https://github.com/saltstack/salt-bootstrap/issues/1074

This continue made sure it did not continue later into the code to enable the services. This is why it was failing on the next step: `install_suse_check_services()` because it did not see the service as being enabled. Tested this and it is working now.


vutny (approved):
:+1:

rallytime (approved):

PR #1080, opened 2017-05-25
Declare End-Of-Life for Ubuntu 12.04 LTS

vutny:
### What does this PR do?
It drops support for Ubuntu LTS 12.04 which reached EOL at April 28, 2017.
Also, installation on derivatives forked from 12.04 release is no longer possible.
The README has been updated accordingly.

### Previous Behavior
Salt installation on Ubuntu 12.04 was possible.

### New Behavior
The script notifies a user that Ubuntu 12.04 reached EOL and no longer supported.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2480,37 +2480,28 @@ __enable_universe_repository() {
 }
 
 install_ubuntu_deps() {
-    if [ "$DISTRO_MAJOR_VERSION" -gt 12 ]; then
-        # Above Ubuntu 12.04 add-apt-repository is in a different package
-        __apt_get_install_noinput software-properties-common || return 1
-    else
-        __apt_get_install_noinput python-software-properties || return 1
-    fi
-
     if [ $_DISABLE_REPOS -eq $BS_FALSE ]; then
-        __enable_universe_repository || return 1
-
-        # Versions starting with 2015.5.6 and 2015.8.1 are hosted at repo.saltstack.com
-        if [ "$(echo "$STABLE_REV" | egrep '^(2015\.5|2015\.8|2016\.3|2016\.11|latest|archive\/)')" = "" ]; then
-            if [ "$DISTRO_MAJOR_VERSION" -lt 14 ]; then
-                echoinfo "Installing Python Requests/Chardet from Chris Lea's PPA repository"
-                add-apt-repository -y "ppa:chris-lea/python-requests" || return 1
-                add-apt-repository -y "ppa:chris-lea/python-chardet" || return 1
-                add-apt-repository -y "ppa:chris-lea/python-urllib3" || return 1
-                add-apt-repository -y "ppa:chris-lea/python-crypto" || return 1
-            fi
-
+        # Install add-apt-repository
+        if ! __check_command_exists add-apt-repository; then
+            __apt_get_install_noinput software-properties-common || return 1
         fi
+
+        __enable_universe_repository || return 1
 
         apt-get update
     fi
 
-    # Minimal systems might not have upstart installed, install it
-    __PACKAGES="upstart"
+    __PACKAGES=''
+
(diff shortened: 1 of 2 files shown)


bootstrap-salt.sh
```diff
@@ -1403,6 +1403,149 @@ __debian_derivatives_translation() {
...
+            if ([ "$DISTRO_MAJOR_VERSION" -eq 11 ] && [ "$SUSE_PATCHLEVEL" -lt 02 ]) || [ "$DISTRO_MAJOR_VERSION" -lt 11 ]; then
+                echoerror "Versions lower than SuSE 11 SP2 are not supported."
+                echoerror "Please consider upgrading to the next stable"
+                exit 1
+            fi
+            ;;
+
+        fedora)
+            # Fedora lower than 18 are no longer supported
+            if [ "$DISTRO_MAJOR_VERSION" -lt 23 ]; then
```
rallytime: I think 23 is EOL too, so let's make that 24 instead.

Also, can you make the comment line above match? (18 should be 24)

bootstrap-salt.sh
```diff
@@ -1403,6 +1403,149 @@ __debian_derivatives_translation() {
...
+                echoerror "End of life distributions are not supported."
+                echoerror "Please consider upgrading to the next stable. See:"
+                echoerror "    https://fedoraproject.org/wiki/Releases"
+                exit 1
+            fi
+            ;;
+
+        centos)
+            # CentOS versions lower than 5 are no longer supported
```
rallytime: Comment should be `lower than 6` for clarity.

bootstrap-salt.sh
```diff
@@ -1403,6 +1403,149 @@ __debian_derivatives_translation() {
...
+            if [ "$DISTRO_MAJOR_VERSION" -lt 6 ]; then
+                echoerror "End of life distributions are not supported."
+                echoerror "Please consider upgrading to the next stable. See:"
+                echoerror "    http://wiki.centos.org/Download"
+                exit 1
+            fi
+            ;;
+
+        red_hat*linux)
+            # Red Hat (Enterprise) Linux versions lower than 5 are no longer supported
```
rallytime: Same thing here, this should read `lower than 6`.

bootstrap-salt.sh
```diff
@@ -1403,6 +1403,149 @@ __debian_derivatives_translation() {
...
+            if [ "$DISTRO_MAJOR_VERSION" -lt 6 ]; then
+                echoerror "End of life distributions are not supported."
+                echoerror "Please consider upgrading to the next stable. See:"
+                echoerror "    https://access.redhat.com/support/policy/updates/errata/"
+                exit 1
+            fi
+            ;;
+
+        oracle*linux)
+            # Oracle Linux versions lower than 5 are no longer supported
```
rallytime: Same thing here. There are a couple more below this comment that should be updated as well, but I won't add a comment for each one.

rallytime (changes requested):
Very nice! Thanks for cleaning this up. I like moving the EOL checks up further in the script. I had some small comments on that function. I know that was largely copy/pasted from below, but we might as well fix it now. :)

bootstrap-salt.sh
rallytime: This change would need an update to the README as well.

vutny: @rallytime I think it would be better if I make a change for Fedora in separate PR just to make things clear and separate concerns.

bootstrap-salt.sh
vutny: Yeah, I'll fix this and later comments to match real comparisons. Thanks for spotting that!

vutny:
@rallytime I've fixed the comments in `__check_end_of_life_versions()` function. I will drop Fedora 23 support with another PR later. Thanks for review.

rallytime (approved):

rallytime:
@vutny sounds good. thanks!

PR #1082, opened 2017-05-26
Fix the -R option on Debian/Ubuntu 

### LINKED ISSUE #1081: The `-R` option is broken on Ubuntu/Debian systems
When attempting to use the `-R` option the bootstrap script fails to use the provided repo and ends up installing from the default distro repos which results in Salt version 0.17.5.

I tested this on the latest bootstrap script downloaded like this:
```
curl -L https://bootstrap.saltstack.com -o install_salt.sh
```
I tested this on Ubuntu 14.04

### Setup
Even using the main `repo.saltstack.com` repo with the option fails

```
./install_salt.sh -D -R repo.saltstack.com -A 10.181.130.1
```
```
root@testbootstrap:~# salt-minion --version
salt-minion 0.17.5
```

rallytime:
### What does this PR do?
When the `-R` option was originally implemented for Debian/Ubuntu, I must not have been testing the right thing because I am not sure how that ever could have worked. This PR reworks some of those `-R` checks and sets up the repositories correctly.

Fixes #1081

### Previous Behavior
The `-R` option wouldn't set up any repo, and would install the default old packages in apt, which is version 0.17. :(

### New Behavior
-R option works well!

```
root@ubuntu:~# sh test.sh -R repo.saltstack.com/staging
<snipped>
root@ubuntu:~# salt-minion --version
salt-minion 2016.11.5 (Carbon)
root@ubuntu:~# cat /etc/apt/sources.list.d/saltstack.list
deb https://repo.saltstack.com/staging/apt/ubuntu/14.04/amd64/latest trusty main
```

I also did a bunch of clean up in this PR that I have been wanting to do for a while. Namely: 
- I moved all of the duplicated warnings about the checking for the correct DPKG_ARCHITECTURE values into a single function
- I created an `install_saltstack_ubuntu_repository` function and moved the saltstack repo set up (and also will work with the `-R` setting with the `_REPO_URL` value) there. This is similar to the way the RHEL distros gate and set up the saltstack repo.
- I created an `install_saltstack_debian_repository` function and moved the saltstack repo set up for BOTH debian 7 and debian 8 into one function. This also works with `-R`, just like the ubuntu repo set up function.
- In these two new functions, I also removed the checks for which "stable" version the user is installing. The packages located on saltstack's repo have been around long enough for now and the old repos are not getting updates any longer.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2527,7 +2527,6 @@ __install_saltstack_ubuntu_repository() {
     apt-get update
 }
 
-
 install_ubuntu_deps() {
     if [ $_DISABLE_REPOS -eq $BS_FALSE ]; then
         # Install add-apt-repository
@@ -2887,6 +2886,27 @@ install_ubuntu_check_services() {
 #
 #   Debian Install Functions
 #
+__install_saltstack_debain_repository() {
+    if [ "$DISTRO_MAJOR_VERSION" -eq 7 ]; then
+        DEBIAN_CODENAME="wheezy"
+    else
+        DEBIAN_CODENAME="jessie"
+    fi
+
+    # amd64 is just a part of repository URI, 32-bit pkgs are hosted under the same location
+    SALTSTACK_DEBIAN_URL="${HTTP_VAL}://${_REPO_URL}/apt/debian/${DISTRO_MAJOR_VERSION}/${__REPO_ARCH}/${STABLE_REV}"
+    echo "deb $SALTSTACK_DEBIAN_URL $DEBIAN_CODENAME main" > "/etc/apt/sources.list.d/saltstack.list"
+
+    if [ "$HTTP_VAL" = "https" ] ; then
+        __apt_get_install_noinput apt-transport-https ca-certificates || return 1
+    fi
+
+    __apt_key_fetch "$SALTSTACK_DEBIAN_URL/SALTSTACK-GPG-KEY.pub" || return 1
+
+    apt-get update
+
+}
+
 install_debian_deps() {
     if [ $_START_DAEMONS -eq $BS_FALSE ]; then
         echowarn "Not starting daemons on Debian based distributions is not working mostly because starting them is the default behaviour."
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -1244,18 +1244,40 @@ __gather_system_info() {
...
     else
         echoerror "dpkg: command not found."
         return 1
     fi
 
-    return 0
+    __REPO_ARCH="$DPKG_ARCHITECTURE"
+
+    if [ "$DPKG_ARCHITECTURE" = "i386" ]; then
+        echoerror "$_REPO_URL likely doesn't have all required 32-bit packages for Ubuntu $DISTRO_MAJOR_VERSION (yet?)."
```
vutny: I guess we should use `${DISTRO_NAME}` variable instead of hard coded `Ubuntu` word here, right?

bootstrap-salt.sh
```diff
@@ -1244,18 +1244,40 @@ __gather_system_info() {
...
+    elif [ "$DPKG_ARCHITECTURE" != "amd64" ]; then
+        echoerror "$_REPO_URL doesn't have packages for your system architecture: $DPKG_ARCHITECTURE."
+
+        if [ "$DPKG_ARCHITECTURE" != "armhf" ]; then
+            echoerror "Try git installation mode with pip and disable SaltStack apt repository, for example:"
+            echoerror "    sh ${__ScriptName} -r -P git v2016.11.5"
+        elif [ "$ITYPE" != "git" ]; then
+            echoerror "You can try git installation mode, i.e.: sh ${__ScriptName} git v2016.11.5"
+        fi
+    fi
```
vutny: As I see, this should work properly during stable installation on Debian/Raspbian `armhf`, but will output unnecessary error messages which may confuse users.
How about to simplify this whole block to a `case` statement?

bootstrap-salt.sh
```diff
@@ -2887,6 +2886,27 @@ install_ubuntu_check_services() {
 #
 #   Debian Install Functions
 #
+__install_saltstack_debian_repository() {
+    if [ "$DISTRO_MAJOR_VERSION" -eq 7 ]; then
+        DEBIAN_CODENAME="wheezy"
+    else
+        DEBIAN_CODENAME="jessie"
+    fi
```
vutny, line 2913: This is fine for now. But I guess in the future we would need a better checks for Debian distros codenames, something like we do for Ubuntu above and in the `__ubuntu_codename_translation` function.

~~Winter~~ Debian Stretch is coming...

vutny (changes requested):
This is super awesome! The code looks DRY and much cleaner now. Great work.

But I would like to offer a couple of suggestions, please take a look below.

s0undt3ch (approved):
As always, @vutny added valuable input!

bootstrap-salt.sh
rallytime: Yeah, I agree. I thought of this, too, but I think a separate PR will be better for this right now. I'll make an issue for it, especially since 7 is phasing out and we'll be moving to 9 very soon as you said. :)

bootstrap-salt.sh
rallytime: Ah, yes. Good catch. I will fix that.

bootstrap-salt.sh
rallytime: Good idea.

rallytime:
@vutny I have updated this PR with your suggestions. Please take another look.

bootstrap-salt.sh
```diff
@@ -1258,24 +1258,30 @@ __check_dpkg_architecture() {
...
+        "i386")
+            error_msg="$_REPO_URL likely doesn't have all required 32-bit packages for $DISTRO_NAME $DISTRO_MAJOR_VERSION."
+            # amd64 is just a part of repository URI, 32-bit pkgs are hosted under the same location
+            __REPO_ARCH="amd64"
+            ;;
+        "amd64")
+            error_msg=""
+            ;;
+        "armhf")
+            error_msg=""
```
vutny: Since there are no `armhf` packages for Ubuntu and Debian 7, maybe we should also check that here?

vutny:
@rallytime That looks really great. Just added some another suggestion above. I know I'm bothersome :smiley:

rallytime:
@vutny Don't worry, this is good feedback! 

My latest change adds warnings for when the architecture detected is `armhf` when installing on Ubuntu and Debian 7. I also moved the `__check_dpkg_architecture` function below the `__ubuntu_derivatives_translation` function to use the "DISTRO_NAME_L" value and added return codes to various warning cases since if we detect an unsupported architecture, we should fail sooner.

Let me know what you think.

rallytime:
Go Go Jenkins!

vutny (approved):
Looks solid!

rallytime:
Go Go Jenkins!

rallytime:
@s0undt3ch I am not sure why the clone job is failing on these tests. I am not as familiar with this jenkins set up. Do you know why that `ci/clone` build is failing?

PR #1084, opened 2017-05-30
Update Fedora support: 23 is EOL now.

rallytime:
Fedora 23 reached EOL in December of 2016, so this just updates the script and README to match that support.

FYI @vutny I grabbed this really quickly today while I was waiting for some other tests to run. :)


vutny (approved):
:smile: I did exactly the same yesterday but forgot to submit a PR.

rallytime:
@vutny Oh, haha! 😄

Issue #1086, opened 2017-06-01
Package unmask not working on Gentoo


ssplatt:
### Description of Issue/Question
When using salt-bootstrap on a new Gentoo VM, the build fails with salt requiring "~amd64" unmasked.

### Setup
tested with "kitchen converge" on a Linode running Gentoo 2017-01-05.

### Steps to Reproduce Issue
make a Linode, deploy Gentoo, login, download salt-bootstrap, run salt-bootstrap.
```
li1244-72 ~ # cd /tmp
li1244-72 tmp # ls
bootstrap-salt.log  bootstrap-salt.sh  stderr
li1244-72 tmp # sh bootstrap-salt.sh -D
 *  INFO: Running version: 2017.05.24
 *  INFO: Executed by: shell pipe
 *  INFO: Command line: 'bootstrap-salt.sh -D'

 *  INFO: System Information:
 *  INFO:   CPU:          GenuineIntel
 *  INFO:   CPU Arch:     x86_64
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   4.9.15-x86_64-linode81
 *  INFO:   Distribution: Gentoo 2.3

 * DEBUG: Binaries will be searched using the following $PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/bin:/usr/x86_64-pc-linux-gnu/gcc-bin/4.9.4
 *  INFO: Installing minion
 * DEBUG: install_gentoo_2_stable_deps not found....
 * DEBUG: install_gentoo_2_3_stable_deps not found....
 * DEBUG: install_gentoo_2_deps not found....
 * DEBUG: install_gentoo_2_3_deps not found....
 * DEBUG: install_gentoo_stable_deps not found....
 *  INFO: Found function install_gentoo_deps
 * DEBUG: DEPS_INSTALL_FUNC=install_gentoo_deps
 * DEBUG: config_gentoo_2_stable_salt not found....
 * DEBUG: config_gentoo_2_3_stable_salt not found....
 * DEBUG: config_gentoo_2_salt not found....
 * DEBUG: config_gentoo_2_3_salt not found....
 * DEBUG: config_gentoo_stable_salt not found....
 * DEBUG: config_gentoo_salt not found....
 *  INFO: Found function config_salt
 * DEBUG: CONFIG_SALT_FUNC=config_salt
 * DEBUG: preseed_gentoo_2_stable_master not found....
[...]

PR #1087, opened 2017-06-01
Ensure presence of CONFIG_PROTECT_MASK files

conn:
Right now, portage will treat absent files as protected.

In order to get around this, we need to make sure they're there.


conn:
This is relevant to #1086

conn:
I've also submitted a patch with portage to fix its behavior.

rallytime (approved):

rallytime:
Thanks @decomposite

conn:
Thanks for the merge! 🍺

PR #1088, opened 2017-06-02
Autounmask as separate step

conn:
When autounmask-write is set and there is a flag that needs to be
written, emerge will write the change and then exit without building.
This change should avoid cases where nothing will build because of this
behavior.


conn:
This is relevant to #1086

rallytime (approved):
Looks good. Thanks!

conn:
Thanks for the merge! 🍺

PR #1092, opened 2017-06-07
Fix -j/-J (JSON configs) options for Debian 8

vutny:
### What does this PR do?
Add `python-yaml` package before calling `apt-get`.
Also remove duplicate functions.

### Previous Behavior
Calling the bootsrap script with `-j` or `-J` option fails because of missing Python `yaml` module.

### New Behavior
Both options are working.


rallytime (approved):
Great! Nice job cleaning this up.

Issue #1094, opened 2017-06-11
Windows bootstrap incorrectly fetches latest version


absolutejam:
### Description of Issue/Question

Running the Windows `bootstrap-salt.ps1` without a version explicitly defined should query the Salt repo and find the latest version but does not.

### Setup

  - WIndows 10 Enterprise x64
  - `$PSVersionTable`:
```powershell
Name                           Value                                                                                                                                                                                                                             
----                           -----                                                                                                                                                                                                                             
PSVersion                      5.0.10586.672                                                                                                                                                                                                                     
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                                                                                                           
BuildVersion                   10.0.10586.672                                                                                                                                                                                                                    
CLRVersion                     4.0.30319.42000                                                                                                                                                                                                                   
WSManStackVersion              3.0                                                                                                                                                                                                                               
PSRemotingProtocolVersion      2.3                                                                                                                                                                                                                               
SerializationVersion           1.1.0.1       
```

### Steps to Reproduce Issue

I stripped out most of the steps to install the minion, and tested the results in `$version` and `$returnMatches`.

```powershell
 [string]$repourl= "https://repo.saltstack.com/windows"

If ([IntPtr]::Size -eq 4) { $arch = "x86" } Else { $arch = "AMD64" }

If (!$version) {
    # Find latest version of Salt Minion
    $repo = Invoke-Restmethod "$repourl"
    $regex = "<\s*a\s*[^>]*?href\s*=\s*[`"']*([^`"'>]+)[^>]*?>"
    $returnMatches = New-Object System.Collections.ArrayList
    $resultingMatches = [Regex]::Matches($repo, $regex, "IgnoreCase")
    foreach($match in $resultingMatches) {
        $cleanedMatch = $match.Groups[1].Value.Trim()
        [void] $returnMatches.Add($cleanedMatch)
    }
    If ($arch -eq 'x86') {
        $returnMatches = $returnMatches | Where {$_ -like "Salt-Minion*x86-Setup.exe"}
    }
    Else {
        $returnMatches = $returnMatches | Where {$_ -like "Salt-Minion*AMD64-Setup.exe"}
    }

    $version = $(($returnMatches | Sort-Object -Descending)[0]).Split(("n-","-A","-x"),([System.StringSplitOptions]::RemoveEmptyEntries))[1]
}
```

[...]

PR #1095, opened 2017-06-11
Removes faulty explicit version ordering

absolutejam:
### What does this PR do?
Changes the version ordering to rely on the correct ordering already present in the repo, and selects the last (Most recent) version. This removes the explicit ordering done with `Select-Object -Descending` as this works alphabetically and does not correctly order versions (eg. `.11` comes before `.2`).

#1094 

### Previous Behavior
Not passing `-version` would fetch the wrong version, where it should fetch the latest.

### New Behavior
Now correctly fetches latest version.

### DIFF
--- a/bootstrap-salt.ps1
+++ b/bootstrap-salt.ps1
@@ -211,7 +211,7 @@ If (!$version) {
         $returnMatches = $returnMatches | Where {$_ -like "Salt-Minion*AMD64-Setup.exe"}
     }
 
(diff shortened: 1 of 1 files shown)


rallytime:
@twangboy Can you review this PR when you have a moment?

rallytime:
@twangboy Bump for a review here. :)

twangboy (approved):

PR #1096, opened 2017-06-12
Support bootstrapping on upcoming Debian GNU/Linux 9.0

vutny:
### What does this PR do?
It happened. This is a quick and rough support for Debian 9 codename "stretch", which is going to become stable at June 17.
The packages would be installed from the SaltStack's repo URL for release 8 ("jessie"), but all dependencies should be already available in the distro.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -226,7 +226,7 @@ Debian and derivatives
 ~~~~~~~~~~~~~~~~~~~~~~
 
 - Cumulus Linux 2/3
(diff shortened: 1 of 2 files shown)


rallytime (approved):

rallytime:
@vutny Thanks for adding this support. Looks good. We'll need to adjust this once repo.saltstack.com has Debian 9 packages, but this works for now!

PR #1098, opened 2017-06-12
Add hash verification information to installation instructions

### LINKED ISSUE #1043: Installation instructions should include hash verification

There is no way to verify that the file downloaded is the correct one. The bootstrap installation procedure should list the correct SHA256 sum of the script or reference a file which contains that sum.

rallytime:
Fixes #1043

cc: @cachedout

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,13 +27,15 @@ are high that your issue was already fixed.
 Bootstrap
 =========
 
+In every two-step installation example, you would be well-served to **verify against the SHA256 sum**
+of the downloaded ``bootstrap-salt.sh`` file.
+
+The SHA256 sum of the ``bootstrap-salt.sh`` file, per release, is:
+
+- 2017.05.24: ``8c42c2e5ad3d4384ddc557da5c214ba3e40c056ca1b758d14a392c1364650e89``
+
 If you're looking for the *one-liner* to install Salt, please scroll to the bottom and use the
 instructions for `Installing via an Insecure One-Liner`_.
-
-.. note::
-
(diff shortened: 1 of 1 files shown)


vutny (commented):
This is OK in general.

But it would be really nice to have a hash sum in a separate file and be able to verify it automatically (and generate it the same way).
For example:
```bash
curl -L https://bootstrap.saltstack.com/bootstrap-salt.sh.sha512 | sha512sum -c -
```

rallytime:
@vutny Yes, I agree. I was thinking about doing it that way as well, but then I wasn't sure where to put it right now. We could add the sum files to repo.saltstack.com, one for each new release of the bootstrap script. This would be similar to how we do this for the windows executables: https://repo.saltstack.com/#windows

I think for now this is a good start, though, until we can get that added to repo.saltstack.com. I want to loop @Ch3LL in here for getting something like this added to the bootstrap release process. @Ch3LL what do you think?

Ch3LL:
We could possibly just add it to bootstrap.saltstack.com but I would like to get @UtahDave 's input here since he is more familiar with how that is setup to see if this is a possibility. That way the sha is in the same subdomain.

rallytime:
@Ch3LL Oh yes, you're right. The same subdomain is best. :) We're already doing this with the Windows sum files, so I think we should do this with the bootstrap releases as well.

rallytime:
I think this is good for now from a bootstrap script perspective. We'll have to follow up with the file repo conversation at a later time.

PR #1099, opened 2017-06-12
Update Debian and Ubuntu service file paths to be consistent with salt

### LINKED ISSUE #1085: Update service files to be consistent
Some discussion was raised in salt issue https://github.com/saltstack/salt/pull/41446 about where the service files are located for various services. The service files used should be the ones in `pkg/*.service` as @dmurphy18 confirmed in [this comment](https://github.com/saltstack/salt/pull/41446#issuecomment-304989819). There are some old service file references that need to be cleaned up so this is consistent.

This issue was originally specifically raised for Debian/Ubuntu service files, but it might be wise to take a look at some of the other paths to see if they're also out of date.

rallytime:
Fixes #1085

In https://github.com/saltstack/salt/pull/41446, the issue was raised about where the salt packages are pulling the various service files from.

There are the pkg/* files, the pkg/deb/* files, and the pkg/debian/* files to choose from. However, it was noted in the comment located here:
https://github.com/saltstack/salt/pull/41446#issuecomment-304989819 that only the pkg/* files are actively updated. The pkg/deb/* and pkg/debian/* files are old artifacts that have not yet been cleaned up.

Therefore, the bootstrap script should not use these service file locations and should match what the salt packages do.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2785,7 +2785,7 @@ install_ubuntu_git_post() {
         [ $fname = "syndic" ] && [ "$_INSTALL_SYNDIC" -eq $BS_FALSE ] && continue
 
         if [ -f /bin/systemctl ] && [ "$DISTRO_MAJOR_VERSION" -ge 16 ]; then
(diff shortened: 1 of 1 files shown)


rallytime:
FYI @dmurphy18 :)

PR #1101, opened 2017-06-12
Add debian_codename_translation function, similar to Ubuntu

### LINKED ISSUE #1083: Add Debian codename distro translation function
We should add a function similar to `__ubuntu_codename_translation()`, but for Debian distributions.

This was suggested in PR #1082, so this issue is to track that idea. Debian 7 is phasing out of support and Debian 9 is coming soon, so something like the ubuntu codename translation but for Debian distros will be useful.

rallytime:
Fixes #1083

This is a simpler function than the one needed for the Ubuntu codename translations, but will still be helpful moving forward with Debian support as Debian 9 comes out.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1444,6 +1444,27 @@ __debian_derivatives_translation() {
 
 
 #---  FUNCTION  -------------------------------------------------------------------------------------------------------
+#          NAME:  __debian_codename_translation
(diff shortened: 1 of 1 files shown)


rallytime:
@vutny Please look this over as it touches changes you made recently in the debian repo set up. :)

vutny (approved):
Looks good!

PR #1104, opened 2017-06-19
FEAT: add support for OS X in bootstrap script.

cournape:
### What does this PR do?

Add support for bootstrapping salt minions on OS X

While the PR adds support for 10.9 upward, since the latest installer only supports 10.11 upward, this effectively supports 10.11 and above only. If the salt minion package is built with https://github.com/saltstack/salt/pull/37858, then the bootstrap script will effectively support 10.9 and above.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1222,6 +1222,42 @@ __gather_bsd_system_info() {
 
 
 #---  FUNCTION  -------------------------------------------------------------------------------------------------------
+#          NAME:  __gather_darwin_system_info
+#   DESCRIPTION:  Discover Darwin (aka OSX aka macOS) systems information
+#----------------------------------------------------------------------------------------------------------------------
+__gather_darwin_system_info() {
+    DISTRO_NAME=${OS_NAME}
+    DISTRO_VERSION=$(sw_vers | grep ProductVersion |  cut -f 2 -d: | tr -d ' \t')
+
+    DISTRO_MAJOR_VERSION=$(echo ${DISTRO_VERSION} | cut -d. -f 1)
+    DISTRO_MINOR_VERSION=$(echo ${DISTRO_VERSION} | cut -d. -f 2)
+
+    if [ "${DISTRO_MAJOR_VERSION}" -ne "10" ]; then
+        echoerror "${DISTRO_NAME} ${DISTRO_VERSION} not supported.";
+        exit 1
+    fi;
+
+    case ${DISTRO_MINOR_VERSION} in
+        [1-8] )
+            echoerror "${DISTRO_NAME} ${DISTRO_VERSION} not supported."
+            ;;
+        9 )
+            echodebug "Detected Maverick"
+            LAUNCHCTL_API_VERSION="maverick"
+            ;;
+        1[0-9]* )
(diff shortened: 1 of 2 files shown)


bootstrap-salt.sh
```diff
@@ -1222,6 +1222,42 @@ __gather_bsd_system_info() {
...
 #---  FUNCTION  -------------------------------------------------------------------------------------------------------
+#          NAME:  __gather_darwin_system_info
+#   DESCRIPTION:  Discover Darwin (aka OSX aka macOS) systems information
+#----------------------------------------------------------------------------------------------------------------------
+__gather_darwin_system_info() {
+    DISTRO_NAME=${OS_NAME}
+    DISTRO_VERSION=$(sw_vers | grep ProductVersion |  cut -f 2 -d: | tr -d ' \t')
+
+    DISTRO_MAJOR_VERSION=$(echo ${DISTRO_VERSION} | cut -d. -f 1)
+    DISTRO_MINOR_VERSION=$(echo ${DISTRO_VERSION} | cut -d. -f 2)
```
vutny: Lint checker fails on the two lines above: remember to quote variable and command substitutions.

bootstrap-salt.sh
```diff
@@ -5161,6 +5205,57 @@ install_openbsd_restart_daemons() {
...
+
+    expected_md5=416321b985a17b2c0c6ee35914946bf9
+
+    tmpdir=$(mktemp -d -t 'salt-tempdir')
+    target="${tmpdir}/salt-installer.pkg"
+    __fetch_url "${target}" "$installer_url"
+
+    actual_md5=$(md5 -q "${target}")
+    if [ "$actual_md5" = "$expected_md5" ]; then
+        installer -pkg ${target} -target / || return 1
```
vutny: Same here, `${target}` should be surrounded with double quotes.

vutny (changes requested):
@cournape Finally! Good work.
There are several lint check violations, however they are trivial to fix. Please take a look at my comments below.

Also, would you mind to update the README file about OS X support as well?

cournape:
@vutny how does that look now ?

vutny (approved):
This is perfect. Thank you.

@rallytime, I think we are good to go.

rallytime:
Thank you for getting this support started @cournape! My main concern with this PR as it stands right now is that there are references to exact installer versions in the `install_darwin_stable` function. We should be able to install any version of salt, and not be pinned to a single version. 

Can you refactor that function to work for multiple versions of salt?

rallytime:
Bump @cournape - did you see my comment above?

rallytime:
@cournape bump

rallytime:
Closing due to no response.

We can certainly open this again when the concerns above are addressed.

tiyab:
Hi, sorry for digging out this topic.
As I was looking a way to automate dotfiles installation on MacOS + Linux via SaltStack, I stumble upon this.
@rallytime The issue is, it will be difficult to not hardcode the MacOS version in the script as there is no `latest` directory for OSX.
If you look how the release are done for Debian or RHEL, you will found a `latest` directory to which the script is referring to.
Until this is change I don't see a viable way to change the PR.

Issue #1105, opened 2017-06-21
Existing minion keys are not overwritten even if -F is specified.


vernondcole:
### Description of Issue/Question

I am framing this as a query rather than a pull request, because there may have been a very
good reason why the programmer decided to do the code the way it is, rather than the way
that would be very convenient for my use case: I want to take control of an existing machine.

As salt-bootstrap is now implemented, I would have to run the deploy command twice:
once to install salt,
a second time with `-C` to configure the minion as I need.

I tried adding a `-F` to the deploy command to do it all in one pass, which does not work.

I am updating the salt-cloud "saltify" driver so that it will be capable of re-assigning a hardware
machine to a different salt-master. 

The `salt-cloud -p` command creates a new key pair for a new minion which it is creating, and sends
that key pair to the minion along with the salt-bootstrap code and a new minion config file.
It loads the minion's public key into the master's pki directory so that it will be accepted
as soon as the minion becomes active.  It then executes the salt-bootstrap script with 
a `-c /tmp/xxx` option.
Unfortunately, if salt-minion were to have been started earlier, a different key pair will
have been placed in /etc/salt/pki/minion, and those files are not replaced. 
When the minion tries to connect to its new master, the public keys are different,
and the connection is immediately rejected.

The obvious solution is to add a `-F` option to the salt-bootstrap command line. [ the command is run about line 1541 of salt/utils/cloud.py ] That solution will not work because the key pair files are
explicitly excluded from copying unless the `-C` option has been specified. 

This exclusion was applied in [#848]( https://github.com/saltstack/salt-bootstrap/pull/848) which seems to me to be slightly odd.  
That PR changed a hard-wired `BS_TRUE` value into a reference to `_CONFIG_ONLY`. 
If the `BS_TRUE` value had simply been removed, so that the value of 
`_FORCED_OVERWRITE` was used, would that not have solved #830 without disabling `-F`?

### Setup

A Virtualbox VM is running Ubuntu 16.04 which has had Salt previously installed - as would be
normal in my use case. Salt-cloud is running in a Python virtual environment on my workstation,
(since I am updating it). For our purposes, the current stable version of salt-cloud should produce
the same result.  (My added options would not work, of course.)

```
# /etc/salt/cloud.profiles.d/saltify_demo_profiles.conf

quail16:   # example machine created by "vagrant up quail16"
  ssh_host: 192.168.8.3  # hard-wired in the Vagrantfile
  ssh_username: vagrant
[...]

PR #1110, opened 2017-06-29
Fix 1105

vernondcole:
### What does this PR do?
- Completes the correction of #848 
- update the Ubuntu version in the Vagrantfile to a supported release.

#1105 

### Previous Behavior

`salt-bootstrap.sh -F` caused overwrite of the minion config, but not the keys.

sh salt-bootstrap.sh failed with 
```
 $ sudo sh salt-bootstrap.sh -A salt.2txt.us -i sb1
 *  INFO: Running version: 2017.05.24
 *  INFO: Executed by: sh
 *  INFO: Command line: 'salt-bootstrap.sh -A salt.2txt.us -i sb1'

 *  INFO: System Information:
 *  INFO:   OS Version:   3.2.0-126-virtual
 *  INFO:   Distribution: Ubuntu 12.04

 * ERROR: End of life distributions are not supported.
 * ERROR: Please consider upgrading to the next stable. See:
 * ERROR:     https://wiki.ubuntu.com/Releases
```

### New Behavior
- Key files will be replaced, too.
- Can actually test using Vagrant

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -10,7 +10,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # please see the online documentation at vagrantup.com.
 
   # Every Vagrant virtual environment requires a box to build off of.
-  config.vm.box = "ubuntu/precise64"
+  config.vm.box = "ubuntu/xenial64"
 
   # The url from where the 'config.vm.box' box will be fetched if it
   # doesn't already exist on the user's system.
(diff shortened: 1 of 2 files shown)


vutny (approved):
Good job @vernondcole !

Since passing `-C` (config only) option implies `-F` (forced overwrite), this is backwards compatible and more useful.

rallytime:
@vernondcole Yes, this is the right way to do this. Thank you for fixing this!

rallytime (approved):

PR #1111, opened 2017-07-05
Add Debian 9 support using official repo.saltstack.com packages

### LINKED ISSUE #1109: Salt-bootstrap require dirmngr
Salt bootstrap require dirmngr but don't install it. Work fine if you install it manually.

### Setup
Install a fresh debian 9.

### Steps to Reproduce Issue
```
gpg: requesting key from 'https://repo.saltstack.com/apt/debian/8/amd64/latest/SALTSTACK-GPG-KEY.pub'
gpg: failed to start the dirmngr '/usr/bin/dirmngr': Aucun fichier ou dossier de ce type
gpg: connecting dirmngr at '/tmp/apt-key-gpghome.rgJCyrRIWX/S.dirmngr' failed: Aucun fichier ou dossier de ce type
gpg: WARNING: unable to fetch URI https://repo.saltstack.com/apt/debian/8/amd64/latest/SALTSTACK-GPG-KEY.pub: Pas de dirmngr
 * ERROR: Failed to run install_debian_deps()!!!
```

### Versions and Systems
./bootstrap-salt.sh -- Version 2017.05.24

rallytime:
### What does this PR do?
Adds more official support for Debian 9 packages to install salt from repo.saltstack.com. Also adds the new git dependency package installation function for Debian 9, since some of these packages have been renamed since Debian 8.

Fixes #1109

### Previous Behavior
Previously, the bootstrap script was using the packages for Debian 8 at repo.saltstack.com to install salt on Debian 9.

### New Behavior
Now that packages are available for Debian 9 for salt, this updates the bootstrap script to use Debian 9 packages instead of Debian 8.

ping @vutny

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1459,6 +1459,9 @@ __debian_codename_translation() {
         "8")
             DISTRO_CODENAME="jessie"
             ;;
+        "9")
+            DISTRO_CODENAME="stretch"
+            ;;
         *)
             DISTRO_CODENAME="jessie"
             ;;
@@ -2924,17 +2927,8 @@ install_ubuntu_check_services() {
 #   Debian Install Functions
 #
 __install_saltstack_debian_repository() {
-    if [ "$DISTRO_MAJOR_VERSION" -eq 9 ]; then
-        # Packages for Debian 9 at repo.saltstack.com are not yet available
-        # Set up repository for Debian 8 for Debian 9 for now until support
-        # is available at repo.saltstack.com for Debian 9.
-        DEBIAN_RELEASE="8"
-    else
-        DEBIAN_RELEASE="$DISTRO_MAJOR_VERSION"
-    fi
-
     # amd64 is just a part of repository URI, 32-bit pkgs are hosted under the same location
-    SALTSTACK_DEBIAN_URL="${HTTP_VAL}://${_REPO_URL}/apt/debian/${DEBIAN_RELEASE}/${__REPO_ARCH}/${STABLE_REV}"
+    SALTSTACK_DEBIAN_URL="${HTTP_VAL}://${_REPO_URL}/apt/debian/${DISTRO_MAJOR_VERSION}/${__REPO_ARCH}/${STABLE_REV}"
     echo "deb $SALTSTACK_DEBIAN_URL $DISTRO_CODENAME main" > "/etc/apt/sources.list.d/saltstack.list"
 
     if [ "$HTTP_VAL" = "https" ] ; then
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -2975,6 +2969,11 @@ install_debian_deps() {
     # YAML module is used for generating custom master/minion configs
     __PACKAGES="${__PACKAGES} python-yaml"
 
+    # Debian 9 needs the dirmgr package in order to import the GPG key later
+    if [ "$DISTRO_MAJOR_VERSION" -ge 9 ]; then
+        __PACKAGES="${__PACKAGES} dirmngr"
```
vutny, line 2974: This is interesting. I don't remember that I have needed this package.
My guess is that `dirmngr` could possibly replace dependency on `gnupg1-curl` we have been introduced before.
I will try to investigate that.

bootstrap-salt.sh
```diff
@@ -3093,6 +3092,57 @@ install_debian_8_git_deps() {
...
+    __PACKAGES="libzmq5 lsb-release python-apt python-crypto python-jinja2 python-msgpack"
+    __PACKAGES="${__PACKAGES} python-requests python-systemd python-yaml python-zmq"
+
+    if [ "$_INSTALL_CLOUD" -eq $BS_TRUE ]; then
+        # Install python-libcloud if asked to
+        __PACKAGES="${__PACKAGES} python-libcloud"
+    fi
+
+    __PIP_PACKAGES=''
+    if (__check_pip_allowed >/dev/null 2>&1); then
```
vutny: I think we don't need to install `tornado` with `pip` on Debian 9, even if requested. It has relatively new version.

vutny (changes requested):
Wow, finally, @rallytime ! 🤗  🎆 

I have just a small suggestion, please take a look below.

bootstrap-salt.sh
rallytime: Yeah when you install with bootstrap, it fails because that package is missing. The package gets included when installing the packages from repo.saltstack.com just fine, but not in bootstrap. It's a dependency of `debconf-utils`. I didn't think to check the `gnupg1-curl` dependency though.

bootstrap-salt.sh
rallytime: Good point. I'll check that and make appropriate changes.

dmurphy18: @rallytime FYI for Debian 9, we only needed to generate packages for Salt and raet / ioflo / timelib .  Everything else is supplied from Debian 9 itself.

rallytime: @dmurphy18 Thanks for pointing that out. I have removed this section. 👍

vutny (approved):
Looks solid!

PR #1114, opened 2017-07-13
Remove pip install requirement on Fedora git installs

rallytime:
The version of fedora available via the dnf package manager is now new enough that we don't need to require users to specify `-P` any longer to install `tornado`. We require a minimum of 4.2.1, but
the version available with dnf is 4.4.2.


vutny (approved):
Nice.

PR #1144, opened 2017-09-06
fedora: use dnf-utils for F26 and above

toanju:
### What does this PR do?
install dnf-utils instead of yum-utils for Fedora 26 and above

dnf-utils is mandatory for installation of salt* 2017* otherwise currently an older version gets installed

### Previous Behavior
older version of salt is installed

### New Behavior
latest version is installed

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -3319,7 +3319,13 @@ install_fedora_deps() {
         __install_saltstack_copr_salt_repository || return 1
     fi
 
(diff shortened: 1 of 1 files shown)


vutny (approved):
👍

rallytime (approved):

rallytime:
Thanks for submitting this @toanju. It looks like our Jenkins instance is restarting, so I want to wait for that to come back online and then run the tests. Pending the test run, this looks good to me.

Issue #1142, opened 2017-08-31
No dependencies installation function found.


ggmartins:
### Description of Issue/Question

Sorry to bug you guys again. I was installing salt stack on raspbian stretch and got this. It was working for jessie.

**sh bootstrap-salt.sh -r -P git v2016.11.5**
   INFO: Running version: 2017.05.24  (tested with 2017.08.17 too)
   INFO: Executed by: sh
   INFO: Command line: 'bootstrap-salt.sh -r -P git v2016.11.5'

   INFO: System Information:
   INFO:   CPU:          :
   INFO:   CPU Arch:     armv7l
   INFO:   OS Name:      Linux
   INFO:   OS Version:   4.9.41-v7+
   INFO:   Distribution: Raspbian 9.1

   INFO: Installing minion
bootstrap-salt.sh: 1: eval: raspbian_9_debian_base: parameter not set
   INFO: Found function config_salt
   INFO: Found function preseed_master
   INFO: Found function daemons_running
 ERROR: **No dependencies installation function found**. Exiting...
### Setup
rpi3 + internet

### Steps to Reproduce Issue
sh bootstrap-salt.sh -r -P git v2016.11.5

### Versions and Systems
./bootstrap-salt.sh -v
./bootstrap-salt.sh -- Version 2017.08.17
uname -a
Linux raspberrypi 4.9.41-v7+ #1023 SMP Tue Aug 8 16:00:15 BST 2017 armv7l GNU/Linux

ggmartins:
tried it on jessie and didn't work. I don't know how the hell I got it working before.

rallytime:
@ggmartins Support for Raspbian on a Debian 9/Stretch base hasn't been added yet. It was on my radar at some point, but since I hadn't filed an issue for it, I forgot. Thank you for filing this.

@vutny You might be interested in adding this initial support if you have some spare time. :)

vutny:
@rallytime Hmm.. I don't have spare RPi to test at the moment... However, I might be able to emulate it.

Actually, this is a broader issue with generic support of various Debian derivatives. I will try to apply a quick fix to unblock @ggmartins for now. Stay tuned.

ggmartins:
You guys are awesome! Thanks so much for considering fixing it. It's a big research project we are working on and it's only possible because of amazing tools like SS.

dmurphy18:
[...]

PR #1145, opened 2017-09-07
Follow symlinks when parsing distro release files

vutny:
### What does this PR do?
It enables `__gather_linux_system_info()` function to follow symlinks while it parses distribution release files to extract necessary information.
This is required because some distros (spotted on Debian 9) and their multiple derivatives have at least `/etc/os-release` file symlinked to `../usr/lib/os-release`.

### Previous Behavior
The script ignored data from release file which are symlinks.

### New Behavior
The script gathers distro data correctly.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1002,7 +1002,6 @@ __gather_linux_system_info() {
             echo redhat-release lsb-release
             )"); do
 
-        [ -L "/etc/${rsource}" ] && continue        # Don't follow symlinks
         [ ! -f "/etc/${rsource}" ] && continue      # Does not exist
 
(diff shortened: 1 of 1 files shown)


rallytime:
@s0undt3ch Can you review this really quickly when you get a moment? I am wondering if there is a historical reason for that line that I am not remembering that might cause problems. 

Also, thank you for looking into this @vutny. The Jenkins instance for salt-bootstrap is restarting, so the tests still need to run on this before it can be merged.

Related to #1142

s0undt3ch (approved):
Can't really remember why we weren't following symlinks. Its been too long...

rallytime:
Ok thanks @s0undt3ch! I just wanted to make sure I didn't miss something.

rallytime (approved):

vutny:
In case of related issues, this always could be reverted :-)

PR #1150, opened 2017-09-22
Declare EOL for SUSE Linux Enterprise Server releases

vutny:
### What does this PR do?
It drops support for SUSE versions less than **11 SP4** and for release 12 everything below SP2.

PR #1156, opened 2017-09-29
Remove patch-level-1 specific code for SLES installs

rallytime:
Anything below SLES 11 SP4 and SLES 12 SP2 is no longer supported by bootstrap, so we no longer need to specify these SP1-specific code.

Refs #1150

See https://www.suse.com/lifecycle/ for more information.


vutny (approved):
👍  Well done!

But needs rebase 😃

Issue #1066, opened 2017-05-04
Salt-bootstrap failed in SLES12-SP1-12.1-0


yogeshprasad:
Hello,
I have tried to run salt-bootstrap in SLES12-SP1-12.1-0 and it is failing.

Logs without -P:
 *  INFO: Running version: 2017.01.10
 *  INFO: Executed by: sh
 *  INFO: Command line: 'bootstrap-salt.sh '

 *  INFO: System Information:
 *  INFO:   CPU:          GenuineIntel
 *  INFO:   CPU Arch:     x86_64
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   3.12.49-11-default
 *  INFO:   Distribution: suse 12.1

 *  INFO: Installing minion
 *  INFO: Found function install_suse_12_stable_deps
 *  INFO: Found function config_salt
 *  INFO: Found function preseed_master
 *  INFO: Found function install_suse_12_stable
 *  INFO: Found function install_suse_12_stable_post
 *  INFO: Found function install_suse_12_restart_daemons
 *  INFO: Found function daemons_running
 *  INFO: Found function install_suse_check_services
 *  INFO: Running install_suse_12_stable_deps()
Repository 'SLES12-SP1-12.1-0' is up to date.
Repository 'Python Modules (SLE_12)' is up to date.
Repository 'SaltStack, dependencies, and addons (SLE_12)' is up to date.
All repositories have been refreshed.
 * ERROR: pip based installations were not allowed. Retry using '-P'
--------------------------------------------------------------------------------------
Logs with -P:
 *  INFO: Running version: 2017.01.10
 *  INFO: Executed by: sh
 *  INFO: Command line: 'bootstrap-salt.sh -P'

 *  INFO: System Information:
 *  INFO:   CPU:          GenuineIntel
 *  INFO:   CPU Arch:     x86_64
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   3.12.49-11-default
 *  INFO:   Distribution: suse 12.1

 *  INFO: Installing minion
 *  INFO: Found function install_suse_12_stable_deps
 *  INFO: Found function config_salt
 *  INFO: Found function preseed_master
 *  INFO: Found function install_suse_12_stable
 *  INFO: Found function install_suse_12_stable_post
 *  INFO: Found function install_suse_12_restart_daemons
 *  INFO: Found function daemons_running
 *  INFO: Found function install_suse_check_services
[...]

PR #1157, opened 2017-10-05
Use suse repo to install packages, rather than defining them

### LINKED ISSUE #714: Problems installing on SLES 12
Hi all,
### What was attempted

Created a new server in AWS using `salt-cloud`, with the image `ami-b90e6283` which is SLES 12.
### Result

salt-cloud was able to do everything except bootstrap salt because of `No dependencies installation function found.`
### How I went about fixing it manually

I had to manually follow the steps here:  
https://docs.saltstack.com/en/latest/topics/installation/suse.html#suse-linux-enterprise

```
zypper addrepo http://download.opensuse.org/repositories/devel:languages:python/SLE_12/devel:languages:python.repo
zypper refresh
zypper install salt salt-minion
```
### Other install steps needed

I think at some point, some packages were half-way through install and failed during the initial cloud bootstrap, so I had to perform the following steps:
1. had to install `python3` and `pip` using `zypper` if they weren't already there.
2. had to install `singledispatch` using `pip` with `--upgrade` flag.  
3. had to install `backports_abc` with `pip.  

After that, I ran salt-minion and it was fine, except things like the hostname had not been set by the bootstrap process, and the minion key came up as `ip-[redacted].ap-southeast-2.compute.internal` on the master when `salt-key -L` was checked.

rallytime:
### What does this PR do?
Refactors much of the SLES and SUSE support to follow some of the more established distro install paradigms.

Since we're setting up suse's package repo above, we should rely on that instead of defining each package explicitly in the `stable` installation functions.

The git installation functions however, still have those packages defined as is customary in other distributions' git install funcs. The relevant packages that weren't already installed with the `stable` function installs have been moved to the `git` install functions.

This change builds on the work done in #1156, #1150, and #1143.

Fixes #714
Refs #1066

### Previous Behavior
Support for SLES 11 wasn't working, and SLES 12 support was only partially working

### New Behavior
SLES 11 SP 4 and SLES 12 SP 2 now work correctly and the code is much cleaner.

Tested on:
- SLES 11 SP4 (stable & git)
- SLES 12 SP2 (stable & git)
- Leap 42.2 (stable & git)

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5425,7 +5425,7 @@ __set_suse_pkg_repo() {
 
 __check_and_refresh_suse_pkg_repo() {
     # Check to see if systemsmanagement_saltstack exists
-    __zypper repos | grep systemsmanagement_saltstack >/dev/null 2>&1
+    __zypper repos | grep -q systemsmanagement_saltstack
 
     if [ $? -eq 1 ]; then
         # zypper does not yet know anything about systemsmanagement_saltstack
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5402,6 +5402,27 @@ install_smartos_restart_daemons() {
 #
 __ZYPPER_REQUIRES_REPLACE_FILES=-1
 
+__set_suse_pkg_repo() {
+
+    # Set distro repo variable
+    if [ "${DISTRO_MAJOR_VERSION}" -gt 2015 ]; then
+        DISTRO_REPO="openSUSE_Tumbleweed"
+    elif [ "${DISTRO_MAJOR_VERSION}" -ge 42 ]; then
+        DISTRO_REPO="openSUSE_Leap_${DISTRO_MAJOR_VERSION}.${DISTRO_MINOR_VERSION}"
+    elif [ "${DISTRO_MAJOR_VERSION}" -lt 42 ]; then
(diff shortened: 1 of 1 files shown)


vutny (approved):
I haven't tested this, because I don't work with SUSEs too often, but looks solid for me!

bootstrap-salt.sh
```diff
@@ -5402,26 +5402,37 @@ install_smartos_restart_daemons() {
...
     else
         suse_pkg_url_base="${HTTP_VAL}://repo.saltstack.com/opensuse"
+        suse_pkg_url_path="${DISTRO_REPO}/systemsmanagement:saltstack:products.repo"
     fi
     SUSE_PKG_URL="$suse_pkg_url_base/$suse_pkg_url_path"
 }
 
+__check_and_refresh_suse_pkg_repo() {
+    # Check to see if systemsmanagement_saltstack exists
+    __zypper repos | grep systemsmanagement_saltstack >/dev/null 2>&1
```
vutny: Maybe it would be better to use `grep -q` instead of slurping errors.

rallytime: Good find! I didn't notice that when I needed to move this function up higher. How does that look now?

vutny: Perfect.

PR #1160, opened 2017-10-16
Use tabs consistently in ChangeLog

eradman:
### What does this PR do?

Convert mix of indentation to tabs. This appears to be the original intent based on the vim parameter `ts=4`, But use spaces to align with the formatting of the README

~~Add vim option `noexpandtab` to keep this more constant in the future~~

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -32,47 +32,47 @@ Version 2017.08.17:
     * Declare End-Of-Life for Ubuntu 12.04 LTS (vutny) #1080
 
 Version 2017.05.24:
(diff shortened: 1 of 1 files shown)


vutny (commented):
Unified formatting is definitely a Greater Good (tabs vs spaces, oh my! 😄 ).

However, in README we already have `expandtab` and `softtabstop=2` been set explicitly, so maybe it would be better to align with it?

eradman:
I think this change is ready to go (converted to spaces)

rallytime:
Thanks @eradman

PR #1164, opened 2017-11-07
Fix yum repo on AWS Linux

### LINKED ISSUE #1159: Wrong yum repo on AWS Linux
Booting a cloud instance with bootstrap-salt.sh version 2017.08.17 configures the old/wrong yum repo.

I tried using this in my cloud profile -- with and without it yields the same results:  `script_args: stable 2017.7`

Here's what I end up with:

```
$ cat /etc/yum.repos.d/saltstack-repo.repo
[saltstack-repo]
name=SaltStack repo for RHEL/CentOS 6
failovermethod=priority
priority=10
gpgcheck=1
gpgkey=https://repo.saltstack.com/yum/redhat/6/$basearch/2017.7/SALTSTACK-GPG-KEY.pub
baseurl=https://repo.saltstack.com/yum/redhat/6/$basearch/2017.7/
```
I see some logic in `install_amazon_linux_ami_deps` to configure the correct repo (`/yum/amazon` instead of `/yum/redhat`), but at the moment I don't get why it's not correct.

cmclaughlin:
If I understand the logic here the new(ish) aws repo url would only be
used in these two cases - when repo_rev:

* starts with latest or 2016.11
* starts with 0-9 and the year string is greater than 2016

If I'm reading this correct, the second case is wrong so I've removed
the check for "starts with 0-9".

### What does this PR do?
Fixes the yum repo URL on aws linux

Fixes #1159 

### Previous Behavior
`baseurl=https://repo.saltstack.com/yum/redhat/6/$basearch/2017.7/`

### New Behavior
`baseurl=https://repo.saltstack.com/yum/amazon/latest/$basearch/2017.7/`

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4423,7 +4423,7 @@ install_amazon_linux_ami_deps() {
     repo_rev="$(echo "${STABLE_REV}"  | sed 's|.*\/||g')"
 
     if echo "$repo_rev" | egrep -q '^(latest|2016\.11)$' || \
-           ( echo "$repo_rev" | egrep -q '^[0-9]+$' && [ "$(echo "$repo_rev" | cut -c1-4)" -gt 2016 ] ); then
+            [ "$(echo "$repo_rev" | cut -c1-4)" -gt 2016 ]; then
        _USEAWS=$BS_TRUE
        pkg_append="python27"
     fi


cmclaughlin:
@vutny you touched this line recently... would you please review?

vutny (approved):
Hi @cmclaughlin 

Looks like that `'^[0-9]+$'` RegEx was completely erroneous. Good job on removing it! Thanks.

FYI @rallytime

rallytime (approved):

rallytime:
Thanks @cmclaughlin - this looks great. We had some test troubles here, so my apologies for opening and closing your PR so much. Everything is all set now, though. Let's get this in!

Issue #1168, opened 2017-11-16
-b option for telling salt-bootstrap to skip installing dependencies causes it to error out instead.


arizvisa:
Documentation for the -b option specifies the following:
```
    -b  Assume that dependencies are already installed and software sources are
        set up. If git is selected, git tree is still checked out as dependency
        step.
```
This implies that the dependency installation step is skipped. However..the following code skips adding any `install_*_deps` functions to the `DEP_FUNC_NAMES` var.

bootstrap-salt.sh:6399
```
# Let's get the dependencies install function
if [ ${_NO_DEPS} -eq $BS_FALSE ]; then
    DEP_FUNC_NAMES="install_${DISTRO_NAME_L}${PREFIXED_DISTRO_MAJOR_VERSION}_${ITYPE}_deps"
    DEP_FUNC_NAMES="$DEP_FUNC_NAMES install_${DISTRO_NAME_L}${PREFIXED_DISTRO_MAJOR_VERSION}${PREFIXED_DISTRO_MINOR_VERSION}_${ITYPE}_deps"
    DEP_FUNC_NAMES="$DEP_FUNC_NAMES install_${DISTRO_NAME_L}${PREFIXED_DISTRO_MAJOR_VERSION}_deps"
    DEP_FUNC_NAMES="$DEP_FUNC_NAMES install_${DISTRO_NAME_L}${PREFIXED_DISTRO_MAJOR_VERSION}${PREFIXED_DISTRO_MINOR_VERSION}_deps"
    DEP_FUNC_NAMES="$DEP_FUNC_NAMES install_${DISTRO_NAME_L}_${ITYPE}_deps"
    DEP_FUNC_NAMES="$DEP_FUNC_NAMES install_${DISTRO_NAME_L}_deps"
elif [ "${ITYPE}" = "git" ]; then
    DEP_FUNC_NAMES="__git_clone_and_checkout"
else
    DEP_FUNC_NAMES=""
fi
```

At the end of the following code, `DEPS_INSTALL_FUNC` results in "null" due to `NO_DEPS` causing `DEP_FUNC_NAMES` to not include any of the `install_*_deps` functions.

bootstrap-salt.sh:6413
```
DEPS_INSTALL_FUNC="null"
for FUNC_NAME in $(__strip_duplicates "$DEP_FUNC_NAMES"); do
    if __function_defined "$FUNC_NAME"; then
        DEPS_INSTALL_FUNC="$FUNC_NAME"
        break
[...]

PR #1169, opened 2017-11-16
Fix fatal error with using the -b option.

arizvisa:
This fixes Issue #1168.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -6576,7 +6576,7 @@ if [ "$_CUSTOM_MASTER_CONFIG" != "null" ] || [ "$_CUSTOM_MINION_CONFIG" != "null
         _TEMP_CONFIG_DIR="$_SALT_ETC_DIR"
     fi
 
-    if [ ${NO_DEPS} -eq $BS_FALSE -a "$_CONFIG_ONLY" -eq $BS_TRUE ]; then
+    if [ ${_NO_DEPS} -eq $BS_FALSE -a "$_CONFIG_ONLY" -eq $BS_TRUE ]; then
         # Execute function to satisfy dependencies for configuration step
         echoinfo "Running ${DEPS_INSTALL_FUNC}()"
         $DEPS_INSTALL_FUNC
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -6576,7 +6576,7 @@ if [ "$_CUSTOM_MASTER_CONFIG" != "null" ] || [ "$_CUSTOM_MINION_CONFIG" != "null
         _TEMP_CONFIG_DIR="$_SALT_ETC_DIR"
     fi
 
-    if [ "$_CONFIG_ONLY" -eq $BS_TRUE ]; then
+    if [ ${NO_DEPS} -eq $BS_FALSE -a "$_CONFIG_ONLY" -eq $BS_TRUE ]; then
```
rallytime: `${NO_DEPS}` should be `${_NO_DEPS}` here, yes?

arizvisa:
yep, my bad. you're right. the variable should've been _NO_DEPS.

bootstrap-salt.sh
arizvisa: yep, my bad. you're right. the variable should've been _NO_DEPS.

rallytime:
@arizvisa Ok, now we have some lint errors to take care of. :)

https://jenkins-testing.saltstack.com/job/bootstrap/job/pr/job/1169/job/lint/5/violations/file/bootstrap-salt.sh/

Can you fix those up, and then we can get this in. Basically, we need to be using `&&` instead of `-a`.

arizvisa:
ah well, i don't care enough to opensource.

it's a shellscript. manpage for '[' and 'test' both say -a is for "and". if you want to fix -b, you know what to do. enjoy!

vutny:
@arizvisa Thanks for the fix, anyhow.

By the way, there's a sense behind having `&&` everywhere. Both `-a` and `-o` to mean AND and OR in a [ .. ] test expression are not well defined. They are obsolescent extensions in POSIX and the lack of additional quoting levels means that expressions involving these operators are often ambiguous. Values that begin with hyphens and the `!` string could become issues.

PR #1174, opened 2017-12-07
Use HTTPS URL for OpenSuse's saltstack repo

gdm85:
In line with #588, I propose using the HTTPS URL for OpenSuse's repository and the same for the list of OpenBSD mirrors.

I verified that the HTTPS versions work (they didn't around the time #588 was opened).


rallytime (commented):
Awesome! Thank you for following up here.

rallytime (approved):

PR #1175, opened 2017-12-12
Adding a new param to force ipv4

dubb-b:
### What does this PR do?
This PR will add an option for you to force IPV4 on all apt-get update commands.

Still needs to be created. Will tie this after...

### Previous Behavior
The cloud provider we use is having issues when we don't force IPv4.

### New Behavior
This check will add the option to force IPv4 when the -T option is passed.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -269,6 +269,7 @@ _QUIET_GIT_INSTALLATION=$BS_FALSE
 _REPO_URL="repo.saltstack.com"
 _PY_EXE=""
 _INSTALL_PY="$BS_FALSE"
+_APT_FORCEIPv4=$BS_FALSE
 
 # Defaults for install arguments
 ITYPE="stable"
@@ -398,12 +399,15 @@ __usage() {
             sh bootstrap.sh -P -y -x python2.7 git v2016.11.3
         The above will install python27 and install the git version of salt using the
         python2.7 executable. This only works for git and pip installations.
+    -T  Forces 'apt-get' to use IPv4. Currently this will default to whatever the machine
+        defults are. The option will add: -o Acquire::ForceIPv4=true to each apt-get update command.
+        
 
 EOT
 }   # ----------  end of function __usage  ----------
 
 
-while getopts ':hvnDc:g:Gyx:wk:s:MSNXCPFUKIA:i:Lp:dH:ZbflV:J:j:rR:aq' opt
+while getopts ':hvnDc:g:Gyx:wk:s:MSNXCPFUKIAT:i:Lp:dH:ZbflV:J:j:rR:aq' opt
 do
   case "${opt}" in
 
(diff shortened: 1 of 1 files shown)


vutny (changes requested):
@dubb-b First of all, thanks for the PR. The important topic has been raised. However, I have few objections against these changes. Anyhow, @rallytime should have the last word.

### TL;DR version
1. Submit support ticket to Cloud vendor
2. Fix your bare OS image
3. Make workaround global and reliable

I will put the long version of my brain dump in the comment below.

vutny:
1. This is only my personal opinion, but I think that the bootstrap script should not be the place to put fixes for all possible underlying infrastructure problems (at OS or even the Cloud level).
   Making IPv4/IPv6 connectivity to co-exist is sometimes a challenging job, especially it could be so in vendor-controlled environment (read Public Cloud). And introducing short-cut commands doesn't really help that. I'm not trying to say we should leave things as is, but rather think about such issues in different way.
   If mentioned Cloud provider has screwed IPv6 so badly, I assume the virtual machine was being broken and not completely usable even **prior to installing Salt and after that too**. It doesn't make any difference what are you trying to install there, right? I'd suggest to fill in support ticket, but I'm aware we usually need things to work by yesterday 😃
   I just might guess that the problem is only with some pre-configured APT repository which couldn't be reachable by IPv6 due to a routing chaos in those internets. That's why `apt-get update` may hang and forcing it to use IPv4 magically fixes such obstacle. So probably disabling that repo or using another mirror might be a solution. Or simply put `Acquire::ForceIPv4 "true";` to `/etc/apt/apt.conf.d/99ipv4` ahead. Below I explain why putting additional option to the command in the script is generally a bad idea.

2. I'm completely fine with the idea that you could add some option to just make things work without having to search internet forums for proposed fixes. But current implementation of the workaround is too much limited to very specific issue.
   * It works only for updating packages list. Installing or upgrading packages from affected repository(ies) will not work. Consider options `-U` or `-p`.
   * It will still not work after the bootstrapping as well, since the change doesn't persist. That's why I suggest to fix the root cause before attempting to install Salt.
   * It is limited to `apt`-based distros. What if `yum`, `dnf`, `zypper`, etc will be affected by the same issue? It would be better to place the workaround at the level below, i.e. for GLibC in `/etc/gai.conf`, where IPv4 preference will work across whole bunch of supported distributions. But that should be done very carefully, and it still would remain fragile to some extend.

I also would like to hear @s0undt3ch thoughts on this.

dubb-b:
@vutny Thanks for the candid feedback. After reading through your comments, I agree with everything you said. This PR would essentially be a hack around an issue with a cloud provider. I thought about that when I read your detailed explanation more. So I intend to do exactly what you propose, either update the `/etc/apt/apt.conf.d/99ipv4` file during the init of the machine or work with the cloud provider to have them update their image. 

Thanks again for your explanation and another perspective for this issue.

-Dubb-b

PR #1186, opened 2017-12-28
Add __wait_for_apt function: avoid locking on the apt-get process

### LINKED ISSUE #1163: Salt Minion install fails because dpkg lock is not released (using salt-cloud)
### Description
When using salt-cloud to create a new AWS EC2 instance, it works, but often fails to get salt-minion installed (error below). I see this often with AWS EC2 instances... if you try and run apt/dpkg too quickly after the EC2 instance was created, you get errors about the dpkg lock not being released. Basically, you need to sleep for a couple seconds and try again, and it'll work.

This was confirmed by taking the command from the salt error and just running it again manually... it worked. 

Is there a way to have salt-bootstrap sleep before trying this or have it retry a few times before dying in error? In the case of an Ubuntu box, I often use a function like this to make sure apt is not running and if it is, to sleep for a couple seconds and try again:

```
isAptGetRunning(){
    while [ $(ps -e | grep "apt" -c) -gt 0 ]
    do
        sleep 2
    done
}
```

Thanks!!

Error:
```
 *  INFO: Running version: 2017.08.17
 *  INFO: Executed by: /bin/sh
 *  INFO: Command line: '/tmp/.saltcloud-da5c774e-ffbf-42da-a79d-bb9b29307b28/deploy.sh -c /tmp/.saltcloud-da5c774e-ffbf-42da-a79d-bb9b29307b28'

 *  INFO: System Information:
 *  INFO:   OS Version:   4.4.0-1038-aws
 *  INFO:   Distribution: Ubuntu 16.04

 *  INFO: Installing minion
 *  INFO: Found function install_ubuntu_stable_deps
 *  INFO: Found function install_ubuntu_stable
 *  INFO: Found function install_ubuntu_stable_post
 *  INFO: Found function install_ubuntu_restart_daemons
 *  INFO: Found function install_ubuntu_check_services
 *  INFO: Running install_ubuntu_stable_deps()
Hit:1 http://eu-west-1.ec2.archive.ubuntu.com/ubuntu xenial InRelease
Hit:2 http://eu-west-1.ec2.archive.ubuntu.com/ubuntu xenial-updates InRelease
Hit:3 http://eu-west-1.ec2.archive.ubuntu.com/ubuntu xenial-backports InRelease
Hit:4 http://security.ubuntu.com/ubuntu xenial-security InRelease
Reading package lists...Connection to 10.22.29.170 closed.

E: Could not get lock /var/lib/dpkg/lock - open (11: Resource temporarily unavailable)
E: Unable to lock the administration directory (/var/lib/dpkg/), is another process using it?
 * ERROR: Failed to run install_ubuntu_stable_deps()!!!
Error: There was a profile error: Command 'ssh -t -t -oStrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null -oControlPath=none -oPasswordAuthentication=no -oChallengeResponseAuthentication=no -oPubkeyAuthentication=yes -oIdentitiesOnly=yes -oKbdInteractiveAuthentication=no -i /etc/salt/biren.pem -p 223 ubuntu@10.22.29.170 'sudo /tmp/.saltcloud-da5c774e-ffbf-42da-a79d-bb9b29307b28/deploy.sh -c '"'"'/tmp/.saltcloud-da5c774e-ffbf-42da-a79d-bb9b29307b28'"'"''' failed. Exit code: 1
```

### Versions Report
```
Salt Version:
           Salt: 2017.7.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Nov 19 2016, 06:48:10)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-1038-aws
         system: Linux
        version: Ubuntu 16.04 xenial
```

rallytime:
Fixes #1163

I also updated some of the example versions from 2016.x versions to use newer 2017.7 versions.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -299,13 +299,13 @@ __usage() {
   Examples:
     - ${__ScriptName}
     - ${__ScriptName} stable
-    - ${__ScriptName} stable 2016.3
-    - ${__ScriptName} stable 2016.3.1
+    - ${__ScriptName} stable 2017.7
+    - ${__ScriptName} stable 2017.7.2
     - ${__ScriptName} daily
     - ${__ScriptName} testing
     - ${__ScriptName} git
-    - ${__ScriptName} git 2016.3
-    - ${__ScriptName} git v2016.3.1
+    - ${__ScriptName} git 2017.7
+    - ${__ScriptName} git v2017.7.2
     - ${__ScriptName} git 06f249901a2e2f1ed310d58ea3921a129f214358
 
   Options:
@@ -395,7 +395,7 @@ __usage() {
         tested with Centos 6 and is considered experimental. This will install the
         ius repo on the box if disable repo is false. This must be used in conjunction
         with -x <pythonversion>.  For example:
-            sh bootstrap.sh -P -y -x python2.7 git v2016.11.3
+            sh bootstrap.sh -P -y -x python2.7 git v2017.7.2
         The above will install python27 and install the git version of salt using the
         python2.7 executable. This only works for git and pip installations.
 
@@ -1331,10 +1331,10 @@ __check_dpkg_architecture() {
     if [ "${error_msg}" != "" ]; then
         echoerror "${error_msg}"
         if [ "$ITYPE" != "git" ]; then
-            echoerror "You can try git installation mode, i.e.: sh ${__ScriptName} git v2016.11.5."
+            echoerror "You can try git installation mode, i.e.: sh ${__ScriptName} git v2017.7.2."
             echoerror "It may be necessary to use git installation mode with pip and disable the SaltStack apt repository."
             echoerror "For example:"
-            echoerror "    sh ${__ScriptName} -r -P git v2016.11.5"
+            echoerror "    sh ${__ScriptName} -r -P git v2017.7.2"
         fi
     fi
 
@@ -1766,11 +1766,40 @@ __function_defined() {
 
 
 #---  FUNCTION  -------------------------------------------------------------------------------------------------------
+#          NAME:  __wait_for_apt
+#   DESCRIPTION:  Check if any apt, apt-get, aptitude, or dpkg processes are running before
+#                 calling these again. This is useful when these process calls are part of
+#                 a boot process, such as on AWS AMIs. This func will wait until the boot
+#                 process is finished so the script doesn't exit on a locked proc.
+#----------------------------------------------------------------------------------------------------------------------
+__wait_for_apt(){
+    echodebug "Checking if apt process is currently running."
+
+    # Timeout set at 15 minutes
+    WAIT_TIMEOUT=900
+
+    while ps -C apt,apt-get,aptitude,dpkg >/dev/null; do
+        sleep 1
+        WAIT_TIMEOUT=$((WAIT_TIMEOUT - 1))
+
+        # If timeout reaches 0, abort.
+        if [ "$WAIT_TIMEOUT" -eq 0 ]; then
+            echoerror "Apt, apt-get, aptitude, or dpkg process is taking too long."
+            echoerror "Bootstrap script cannot proceed. Aborting."
+            return 1
+        fi
+    done
+
+    echodebug "No apt processes are currently running."
+}
+
+#---  FUNCTION  -------------------------------------------------------------------------------------------------------
 #          NAME:  __apt_get_install_noinput
 #   DESCRIPTION:  (DRY) apt-get install with noinput options
 #    PARAMETERS:  packages
 #----------------------------------------------------------------------------------------------------------------------
 __apt_get_install_noinput() {
+    __wait_for_apt
     apt-get install -y -o DPkg::Options::=--force-confold "${@}"; return $?
 }   # ----------  end of function __apt_get_install_noinput  ----------
 
@@ -1780,6 +1809,7 @@ __apt_get_install_noinput() {
 #   DESCRIPTION:  (DRY) apt-get upgrade with noinput options
 #----------------------------------------------------------------------------------------------------------------------
 __apt_get_upgrade_noinput() {
+    __wait_for_apt
     apt-get upgrade -y -o DPkg::Options::=--force-confold; return $?
 }   # ----------  end of function __apt_get_upgrade_noinput  ----------
 
@@ -1790,6 +1820,7 @@ __apt_get_upgrade_noinput() {
 #    PARAMETERS:  url
 #----------------------------------------------------------------------------------------------------------------------
 __apt_key_fetch() {
+    __wait_for_apt
     url=$1
 
     # shellcheck disable=SC2086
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -1765,12 +1765,25 @@ __function_defined() {
...
 
 
+#---  FUNCTION  -------------------------------------------------------------------------------------------------------
+#          NAME:  __wait_for_apt
+#   DESCRIPTION:  Check if any apt process is running before calling apt-get again. This is
+#                 useful when an apt-get call is part of a boot process and will wait until
+#                 wait until the apt-get process is finished, such as on AWS AMIs.
+#----------------------------------------------------------------------------------------------------------------------
+__wait_for_apt(){
+    while pgrep -l apt; do
```
vutny: The `pgrep` command is a part of `procps` package. It does not belong to `required` Debian archive section, so there's a possibility that it would not be installed by default on "tiny-core" systems.
More important, that this check does not cover raw `dpkg` operations, when it does not invoked by `apt-get` or similar tools.

I suggest to use some harmless APT operation to check if lock is in place, for example: `apt-get check 2>/dev/null`.
Also, having a timeout, even hard-coded, would be very convenient.

bootstrap-salt.sh
```diff
@@ -2576,6 +2591,7 @@ __install_saltstack_ubuntu_repository() {
 
     __apt_key_fetch "$SALTSTACK_UBUNTU_URL/SALTSTACK-GPG-KEY.pub" || return 1
 
+    __wait_for_apt
     apt-get update
```
vutny, line 2611: Maybe we could wrap `apt-get update` command with waiter in an "internal" function (like `__apt_get_install_noinput`) to keep the code DRY?

vutny (changes requested):
@rallytime Looks good overall, but I have few suggestions. Those are rather "good-to-have" items, though.

bootstrap-salt.sh
rallytime: Ah, I was worried about that. I did a little research to see if `pgrep` is included and the ones I looked at did have it. But I didn't think about the smaller options. I can update it to your suggestion.

For the time out, what would be useful here? I know we don't want to hang, but we also don't want to short-circuit the wait functionality.

bootstrap-salt.sh
rallytime: I thought of that, but we're calling bare `apt-get update` in many places. I thought this way would be better until we have time to do a larger refactor. I don't have time to do a larger change at this time, along with the requisite testing. Is that something you'd be interested in doing instead?

bootstrap-salt.sh
vutny: As for timeout I think 15 minutes (900 seconds) should be more than enough. Because longer time to wait would indicate that something really going too slow or bad things happen, so we could not help that in any case.

And I think wrapping this "waiter" function with INFO or DEBUG messages would be very useful as well. Or maybe I'm asking too much? 😃

bootstrap-salt.sh
vutny: No worries, probably I could do that later.

bootstrap-salt.sh
rallytime: > Or maybe I'm asking too much? 

Not at all! Great idea. I'll fix that.

rallytime:
@vutny Thanks for the review, as always. :)

Can you take a look at my changes from today?

bootstrap-salt.sh
```diff
@@ -1765,12 +1765,49 @@ __function_defined() {
...
+    # Timeout set at 15 minutes
+    APT_WAIT_TIMEOUT=900
+
+    # set the command to use to check for apt process
+    if __check_command_exists pgrep; then
+        APT_PROC_CHECK="pgrep -l apt"
+    else
+        # pgrep isn't available on smaller Debian systems
+        # shellcheck disable=SC2009
+        APT_PROC_CHECK="ps aux | grep apt"
```
vutny: Unfortunately, this will not work as expected. Check it out:
```bash
$ ps aux | grep apt
user    13686  0.0  0.0  13036   948 pts/11   S+   13:19   0:00 grep --color=auto apt
$ echo $?
0
$ APT_PROC_CHECK="ps aux | grep apt"
$ $APT_PROC_CHECK
error: garbage option

Usage:
 ps [options]

 Try 'ps --help <simple|list|output|threads|misc|all>'
  or 'ps --help <s|l|o|t|m|a>'
 for additional help text.

For more details see ps(1).
```

bootstrap-salt.sh
```diff
@@ -1765,12 +1765,49 @@ __function_defined() {
...
+    fi
+
+    # Check if any apt process is running
+    while ${APT_PROC_CHECK} >/dev/null; do
```
vutny: The problem here is that such substitution not guaranteed to work properly, but also checking for process names which contain "apt" sub-string may give lots of false positives. Trivial example is when `apt-cacher-ng` daemon already installed on a target system. I think we could find even more tools with "apt" in their names.

So I suggest to explicitly test for running `dpkg`, `apt`, `apt-get` and `aptitude` processes. And it appears that good old `ps` is capable to do what we need!
```sh
$ ps -C apt,apt-get,aptitude,dpkg
```

Here are some tests:
```bash
# apt-get update >/dev/null & ps -C apt,apt-get,aptitude,dpkg >/dev/null; echo $?
[1] 16096
0
# find / -name xxx &>/dev/null & ps -C apt,apt-get,aptitude,dpkg >/dev/null; echo $?
[2] 16419
[1]   Done                    apt-get update > /dev/null
1
```

vutny (changes requested):
Hey @rallytime . I have realized that my previous suggestion dramatically slows down the script execution, because the waiter would be called before each `apt` invocation.

However, combining both `pgrep` and `ps` seems not much reliable though. I propose to ether to use `pgrep` only and if it's not available proceed as is; or completely rely on `ps`. Please see detailed explanation below.

bootstrap-salt.sh
rallytime: @vutny I remember i ran into this before, but then through all of the changes, forgot to test and confirm this again. I'll fix that today. Thanks!

rallytime:
@vutny Ok, one more time here. I think that should do it now. ;)

vutny (approved):
Looks awesome @rallytime !

rallytime:
Hooray! Thanks again for the excellent reviews. It is very much appreciated.

PR #1191, opened 2018-01-03
Install swig30 as freebsd dep instead of swig.

abednarik:
Looks like swig was removed from FreeBSD Ports. See https://www.freshports.org/commit.php?category=devel&port=swig13&files=yes&message_id=201712201528.vBKFSwpR091538@repo.freebsd.org

### What does this PR do?

Install swig30 package instead of swig

No

### Previous Behavior

Salt fails to install in FreeBSD 

```
       pkg: No packages available to install matching 'swig' have been found in the repositories

        * ERROR: Failed to run install_freebsd_git_deps()!!!

       salt-call: not found

       No salt-minion installed, install must have failed!!

       salt_install = bootstrap

       salt_url = http://bootstrap.saltstack.org
```

### New Behavior

Salt installs just fine

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4898,9 +4898,9 @@ install_freebsd_9_stable_deps() {
         __configure_freebsd_pkg_details || return 1
     fi
 
-    # Now install swig
+    # Now install swig30
     # shellcheck disable=SC2086
-    /usr/local/sbin/pkg install ${FROM_FREEBSD} -y swig || return 1
+    /usr/local/sbin/pkg install ${FROM_FREEBSD} -y swig30 || return 1
 
     # YAML module is used for generating custom master/minion configs
     # shellcheck disable=SC2086


bootstrap-salt.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh -
+/#!/bin/sh -
```
rallytime: Was this additional slash intentional? ;)

rallytime (changes requested):

bootstrap-salt.sh
abednarik: Sorry, my bad. Fixed.

abednarik:
Looks like I did a squash and is no longer signed with my gpg. Let me see how I can fix that

abednarik:
Please let me know if there anything else I can do. Since we use FreeBSD, our Pipelines are broken and we need to do dirty workarounds to install salt.

Thanks.

rallytime (approved):

rallytime:
@abednarik Merged - thank you for updating the support for this!

abednarik:
Thanks to you @rallytime !

abednarik:
HI @rallytime 

 Sorry to bother, do you have an idea when this patch will released? We still have hacks everywhere to install our FreeBSD servers.

Thanks.

rallytime:
We just very recently did a bootstrap release, so it probably wont be for a month or so. However, you can pull down the version in develop and use the patched version. If you're using salt-cloud, you just place it in the `/etc/salt/cloud.deploy.d/` directory and it will pick up the version you put in there.

abednarik:
Hi @rallytime 

 This is still an issue. Do you plan to merge this to master and release a new version? Looks like this one didn't went to stable in the last merge.

Thanks.

rallytime:
Hi @abednarik Yes, I still plan to merge this into stable and release a new version. I got busy with the 2018.3.0 release for salt and this took a back seat. I'm working on slating some time to release in the next week or so. Thanks!

PR #1197, opened 2018-01-25
OpenBSD has a cdn which handles selecting the best mirror

jasperla:
### What does this PR do?

Set the mirror by overwriting the installurl file which can only
contain a single line so the original behaviour of appending
might not have worked reliably.
The user can pass '-r' to disable setting the repos already to
opt-out of this behaviour.

### Previous Behavior

It went through a process of selecting a mirror based on ping times.

### New Behavior

Use the OpenBSD CDN and offload best-mirror selection to it.

This was tested on OpenBSD 6.2.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5147,35 +5147,11 @@ install_freebsd_restart_daemons() {
 #   OpenBSD Install Functions
 #
 
-__choose_openbsd_mirror() {
-    OPENBSD_REPO=''
-    MINTIME=''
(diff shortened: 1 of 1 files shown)


rallytime:
@vutny and @eradman - you have both been involved with this function. Can you guys review this change?

eradman:
I tested this change in a VM and it works fine. Very nice simplification.

jasperla:
Thanks!

rallytime:
Thank you!

PR #1205, opened 2018-03-08
Use integer parameter for sleep command

bdrung:
The bootstrap-salt.sh script uses a "bashism" for the sleep command. Bash allows real numbers as sleep time, but other POSIX shell interpreters only support an integer as parameter. Thus raise the sleep timeout from 0.1 to 1 second.

The git commits baa54a64, acbd9842, 1aea3037, and 03186a6d that introduce the sleep commands do not explain why a sleep command is used in the code. You might be able to remove the sleep command completely.

Bug-Debian: https://bugs.debian.org/772406


rallytime:
Ah, I was hoping there would be more explanation there as well. Let's leave them for now. Thanks for fixing this @bdrung!

PR #1206, opened 2018-03-08
Add shellcheck test to travis config file

rallytime:
Remove the other testing lines - travis for bootstrap has been disabled for years. Let's just replace the old set up and use a new one now instead.

I'm not sure if this will work correctly - let's see how this run goes.


rallytime:
Actually, I just spoke with @gtmanfred and we're going to be moving these tests over to the new Jenkins anyway, so I'll just close this.

Issue #1137, opened 2017-08-23
Bootstrap of Debian family broken in 2017.08.17


EvaSDK:
### Setup

I run minions in a network with Internet access only available through an HTTP proxy.
After upgrading to 2017.08.17, boostrapping new minions fails due to changes introduced in commit 0e45ba1c72c674562c9336716d6bc741ad649227.

### Logs

```
[...]
Dépaquetage de gnupg1-curl (1.4.21-4) ...
Sélection du paquet gnupg1-l10n précédemment désélectionné.
Préparation du dépaquetage de .../gnupg1-l10n_1.4.21-4_all.deb ...
Dépaquetage de gnupg1-l10n (1.4.21-4) ...
Paramétrage de gnupg1 (1.4.21-4) ...
Paramétrage de gnupg1-l10n (1.4.21-4) ...
Traitement des actions différées (« triggers ») pour man-db (2.7.6.1-2) ...
Paramétrage de gnupg1-curl (1.4.21-4) ...
Warning: apt-key output should not be parsed (stdout is not a terminal)
Executing: /tmp/apt-key-gpghome.tlIJzNCn0a/gpg.1.sh --keyserver-options ca-cert-file=/etc/ssl/certs/ca-certificates.crt --fetch-keys https://repo.saltstack.com/apt/debian/9/amd64/2016.3/SALTSTACK-GPG-KEY.pub
gpg: keyserver option 'ca-cert-file' is obsolete; please use 'hkp-cacert' in dirmngr.conf
gpg: requesting key from 'https://repo.saltstack.com/apt/debian/9/amd64/2016.3/SALTSTACK-GPG-KEY.pub'
gpg: WARNING: unable to fetch URI https://repo.saltstack.com/apt/debian/9/amd64/2016.3/SALTSTACK-GPG-KEY.pub: Connexion terminée par expiration du délai d'attente
Ign:1 http://ftp.fr.debian.org/debian stretch InRelease
Atteint:2 http://ftp.fr.debian.org/debian stretch-updates InRelease
Atteint:3 http://security.debian.org/debian-security stretch/updates InRelease
Atteint:4 http://ftp.fr.debian.org/debian stretch Release
Réception de:6 https://repo.saltstack.com/apt/debian/9/amd64/2016.3 stretch InRelease [2 841 B]
Ign:6 https://repo.saltstack.com/apt/debian/9/amd64/2016.3 stretch InRelease
[...]

PR #1214, opened 2018-04-04
Workaround gnupg2 not being proxy friendly, refs #1137

EvaSDK:
GnuPG2 does not support proxies using the CONNECT method which means only
transparent proxies are supported for encrypted traffic or one must degrade
security by disabling encryption during provisioning.

This commit switches back to curl piped to apt-key which is found on several software projects homepage, even salstack as a matter of fact: https://repo.saltstack.com/#debian, and works for all cases.

### What does this PR do?

Workaround GnuPG2 limitations when used behind proxies.

Issue originally reported as #1137.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1793,7 +1793,7 @@ __apt_key_fetch() {
     url=$1
 
     # shellcheck disable=SC2086
-    apt-key adv ${_GPG_ARGS} --fetch-keys "$url"; return $?
+    curl ${_CURL_ARGS} -f -s -S -L "$url" | apt-key add - ; return $?
 }   # ----------  end of function __apt_key_fetch  ----------
 
 
@@ -2557,7 +2557,7 @@ __install_saltstack_ubuntu_repository() {
 
     # Install downloader backend for GPG keys fetching
     if [ "$DISTRO_VERSION" = "16.10" ] || [ "$DISTRO_MAJOR_VERSION" -gt 16 ]; then
-        __PACKAGES="${__PACKAGES} gnupg2 dirmngr"
+        __PACKAGES="${__PACKAGES} gnupg2 dirmngr curl"
     else
         __PACKAGES="${__PACKAGES} gnupg-curl"
     fi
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -1793,7 +1793,7 @@ __apt_key_fetch() {
     url=$1
 
     # shellcheck disable=SC2086
-    apt-key adv ${_GPG_ARGS} --fetch-keys "$url"; return $?
+    curl ${_CURL_ARGS} -f -s -S -L "$url" | apt-key add - ; return $?
```
vutny, line 1796: You don't need to explicitly install and use `curl`, we have special wrapper function for downloading things: `__fetch_url` [here](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L755).

vutny (changes requested):
Okay, `dirmngr` in Debian "jessie" is broken, that's why `apt-key adv` command fails silently.

But it seems that future versions of Debian and recent Ubuntus seems unaffected (I might be wrong, but assume GNU folks fix things, eventually). And of course it works if proxy is not the case.

So how about to add a fallback action piping downloaded key to `apt-key add` when `adv` command fails to connect?

I see no reason to completely sacrifice Dirmngr, because it serves it's security hardening purpose: checking for publicly revoked keys, when configured properly. So if SaltStack's signing key would be compromised (hope this never would happen!), the dirmngr will catch that.

I suggest to check the output of `apt-key adv` and then download the key only if `gpg` complains about "unable to fetch URI".

rallytime:
In addition to the comments above, this needs to be submitted against the `develop` branch and not `stable`. Stable only receives PRs when we merge `develop` into `stable` in preparation for a release.

EvaSDK:
Sorry, I worked against stable as this is what I needed for my production environment.

EvaSDK:
@vutny, not just Jessie, Stretch is affected as well. And I only provision LTS Ubuntu releases and Xenial ships with an older version of gnupg2 than Stretch. I'm ok with trying to use dirmngr if possible but my goal was to unbreak stable for me.

vutny:
@EvaSDK Indeed, I was wrong. A buggy version of `dirmngr` is present in `stretch`. `jessie` is not affected because there is older gnupg version used with `gnupg-curl` plugin, which apparently works with proxies.

You also could pre-accept downloaded SaltStack key before running the bootstrap script to workaround the issue in current stable.

rallytime:
@EvaSDK Can you update your PR with the suggestion from @vutny to use the `__fetch_url` function and rebase this against the `stable` branch? That way we can get your fix in.

rallytime:
Since we haven't heard back here for a little while, I am going to close this as we do not merge changes in the `stable` branch directly.

If you want to rebase this change against the `develop` branch and incorporate the changes suggested above, we would certainly love to get this in.

We can of course continue the discussion here or on the new PR. Thanks all!

PR #1219, opened 2018-04-19
Install py-tornado4 for FreeBSD.

abednarik:
### What does this PR do?

Install py-tornado4

See https://github.com/saltstack/salt/issues/45790

### Previous Behavior

Salt fails to run, because tornado 5 is not yet supported

### New Behavior

Salt works just fine


rallytime (approved):
This is good to get you going for now, but once Tornado 5.0 support is fixed, we should revert this back. I'll make an issue to track that.

Thanks @abednarik!

abednarik:
Sure, thanks!

Issue #1217, opened 2018-04-11
-b option causes bootstrap-salt.sh not to download git repository


jheidbrink:
### Description of Issue/Question

When installation type ("$ITYPE") is "git", the "-b" switch prevents bootstrap-salt.sh from downloading the repository (contrary to the documentation). This breaks the installation.

### Steps to Reproduce Issue
Take an Ubuntu 16.04 system with git and python installed. Place the bootstrap-salt.sh somewhere and execute:
`sh bootstrap-salt.sh -D -b git`

### Explanation

First, the `DEPS_INSTALL_FUNC` is set to `__git_clone_and_checkout`. This function is never called because of the check for `$_NO_DEPS` introduced in commit 958fb6.

### output of failing bootstrap-salt.sh call
`sh bootstrap-salt.sh -D -b git`
` *  INFO: Running version: 2017.12.13`
` *  INFO: Executed by: sh`
` *  INFO: Command line: 'bootstrap-salt.sh -D -b git'`
` *  WARN: Running the unstable version of bootstrap-salt.sh`
``
` *  INFO: System Information:`
` *  INFO:   CPU:          GenuineIntel`
` *  INFO:   CPU Arch:     x86_64`
` *  INFO:   OS Name:      Linux`
` *  INFO:   OS Version:   4.14.32-1-lts`
` *  INFO:   Distribution: Ubuntu 16.04`
``
` * DEBUG: Binaries will be searched using the following $PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`
` *  INFO: Installing minion`
` *  INFO: Found function __git_clone_and_checkout`
` * DEBUG: DEPS_INSTALL_FUNC=__git_clone_and_checkout`
` * DEBUG: config_ubuntu_16_git_salt not found....`
` * DEBUG: config_ubuntu_16_04_git_salt not found....`
` * DEBUG: config_ubuntu_16_salt not found....`
` * DEBUG: config_ubuntu_16_04_salt not found....`
` * DEBUG: config_ubuntu_git_salt not found....`
` * DEBUG: config_ubuntu_salt not found....`
` *  INFO: Found function config_salt`
` * DEBUG: CONFIG_SALT_FUNC=config_salt`
` * DEBUG: preseed_ubuntu_16_git_master not found....`
` * DEBUG: preseed_ubuntu_16_04_git_master not found....`
` * DEBUG: preseed_ubuntu_16_master not found....`
[...]

PR #1226, opened 2018-05-07
Call __git_clone_and_checkout directly

jheidbrink:
### What does this PR do?

The function `__git_clone_and_checkout` is not called from the various `install_xxx_deps` functions, but "top level" from the `bootstrap-salt.sh` script.

issue #1217 

### Previous Behavior

When `bootstrap-salt.sh` is invoked with the `-b` switch, the git tree is not checked out (contrary to the documentation).

### New Behavior

The `-b` switch does not prevent the script from checking out the git tree.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -6416,6 +6416,7 @@ daemons_running() {
 #======================================================================================================================
 
 # Let's get the dependencies install function
+DEP_FUNC_NAMES=""
 if [ ${_NO_DEPS} -eq $BS_FALSE ]; then
     DEP_FUNC_NAMES="install_${DISTRO_NAME_L}${PREFIXED_DISTRO_MAJOR_VERSION}_${ITYPE}_deps"
     DEP_FUNC_NAMES="$DEP_FUNC_NAMES install_${DISTRO_NAME_L}${PREFIXED_DISTRO_MAJOR_VERSION}${PREFIXED_DISTRO_MINOR_VERSION}_${ITYPE}_deps"
@@ -6423,10 +6424,6 @@ if [ ${_NO_DEPS} -eq $BS_FALSE ]; then
     DEP_FUNC_NAMES="$DEP_FUNC_NAMES install_${DISTRO_NAME_L}${PREFIXED_DISTRO_MAJOR_VERSION}${PREFIXED_DISTRO_MINOR_VERSION}_deps"
     DEP_FUNC_NAMES="$DEP_FUNC_NAMES install_${DISTRO_NAME_L}_${ITYPE}_deps"
     DEP_FUNC_NAMES="$DEP_FUNC_NAMES install_${DISTRO_NAME_L}_deps"
-elif [ "${ITYPE}" = "git" ]; then
-    DEP_FUNC_NAMES="__git_clone_and_checkout"
-else
-    DEP_FUNC_NAMES=""
 fi
 
 DEPS_INSTALL_FUNC="null"
@@ -6571,6 +6568,7 @@ if [ "$INSTALL_FUNC" = "null" ]; then
     exit 1
 fi
 
+
 # Install dependencies
 if [ ${_NO_DEPS} -eq $BS_FALSE ] && [ $_CONFIG_ONLY -eq $BS_FALSE ]; then
     # Only execute function is not in config mode only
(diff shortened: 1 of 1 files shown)


rallytime:
Hi @jheidbrink - Thank you for submitting this PR. I ran this through a quick test and unfortunately this approach breaks current functionality.

If you run the bootstrap script on a new system (no dependencies are already installed), the script fails:
```
root@localhost:~# sh salt.sh git v2018.3.0
 *  INFO: Running version: 2018.04.25
 *  INFO: Executed by: sh
 *  INFO: Command line: 'salt.sh git v2018.3.0'

 *  INFO: System Information:
 *  INFO:   OS Version:   4.15.13-x86_64-linode106
 *  INFO:   Distribution: Ubuntu 16.04

 *  INFO: Installing minion
 *  INFO: Found function install_ubuntu_git_deps
 *  INFO: Found function install_ubuntu_git
 *  INFO: Found function install_ubuntu_git_post
 *  INFO: Found function install_ubuntu_restart_daemons
 *  INFO: Found function install_ubuntu_check_services
salt.sh: 1882: salt.sh: git: not found
 *  INFO: Git revision matches a Salt version tag, shallow cloning enabled.
salt.sh: 1975: salt.sh: git: not found
```
A different approach will be needed here to preserve the default functionality.

jheidbrink:
Hi @rallytime, thanks for pointing that out. It should be fixed with my latest commit (tested by running the following in a ubuntu:xenial container: `bootstrap-salt.sh -f -d -N -M -X -P git`).

Note that calling that command with the `-b` switch fails as git doesn't get installed, but I think that's ok, because if we pass the `-b` switch, we assume that git is already installed.

rallytime:
@jheidbrink Thanks for fixing that. I tested this again to double check and all is well again.

> Note that calling that command with the -b switch fails as git doesn't get installed, but I think that's ok, because if we pass the -b switch, we assume that git is already installed.

Maybe it would be good to do some checking if `-b` is passed and git is missing? We can exit early if this is the case so people know what the problem is with an `echoerror`.

jheidbrink:
Hi @rallytime,

I'm not sure if I like the idea of letting the script check missing dependencies. After all, it does install the dependencies per default. If people decide to not use that mechanism but to take care of dependency installation themselves, I think it's ok not to add extra code for checks (that would also have to be maintained when dependencies change).
Anyway, I'd say it's out of scope of this pull request, with is merely addressing issue 1217.

rallytime:
Yeah, that's fair. We weren't doing any error checking there before, just thinking of future bugs people might file. :)

This is good as-is. We can always cross that bridge later if needed.

rallytime (approved):

rallytime:
@vutny Do you want to look over this one as well?

vutny (changes requested):
I generally like the idea of calling git clone from a single place, but it seems to be achieved not that easy...

These changes apparently break usage of `-a` parameter -- install all deps from `pip`. Because "deps" functions read the requirements file(s) from locally checked out sources to determine what to install.

So you need to have a local copy of Salt git repo before executing dependency installation, and at the same time have `git` to be installed on particular platform to be able to pull from GitHub.

I currently don't have any good solution in mind to resolve this properly.

jheidbrink:
I have updated the pull request:

The `__git_clone_and_checkout` function is again called in the indivudual `install_{distro}_git_deps` functions as it was originally done. When the script is called with NO_DEPS (so `install_{distro}_deps` is not executed) then it is called "standalone".

Here is another possible approach, that would keep the call to `__git_clone_and_checkout` in one place:

```
def gitinstall_ubuntu():
    ...

[... more gitinstall_{distro} functions ...]

if not NO_DEPS:
    gitinstall_{distro}

if ITYPE == 'git':
    __git_clone_and_checkout

if not NO_DEPS:
    install_{distro}_deps
```

vutny (approved):
Thanks @jheidbrink . This is good to go in. You may want to rebase on latest `develop` head as well.

I think we wouldn't want to introduce more complexity by slicing and dicing functions across all supported distros. The current state may be sub-optimal from DRY coding perspective, but it is much easier to follow a flow of operations.

jheidbrink:
Hi @vutny,

I have now squashed the commits into one; it's based on the develop branch.

PR #1229, opened 2018-05-10
Fix the variable ref from #803

rallytime:
The version of shellcheck that was previously running on bootstrap PRs was out of date and not catching the fact that the VIRTUAL_ENV variable didn't exist. That variable should be _VIRTUALENV_DIR instead.

Refs #803

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2394,7 +2394,7 @@ __create_virtualenv() {
 __activate_virtualenv() {
     set +o nounset
     # Is virtualenv empty
-    if [ -z "$VIRTUAL_ENV" ]; then
+    if [ -z "$_VIRTUALENV_DIR" ]; then
         __create_virtualenv || return 1
         # shellcheck source=/dev/null
         . "${_VIRTUALENV_DIR}/bin/activate" || return 1


bootstrap-salt.sh
```diff
@@ -2394,9 +2394,9 @@ __create_virtualenv() {
 __activate_virtualenv() {
     set +o nounset
     # Is virtualenv empty
-    if [ -z "$VIRTUAL_ENV" ]; then
+    if [ -z "$_VIRTUALENV_DIR" ]; then
         __create_virtualenv || return 1
-        # shellcheck source=/dev/null
+        # shellcheck disable=34
```
vutny: I think this is rule [SC1090](https://github.com/koalaman/shellcheck/wiki/SC1090).
What is `34` by the way? I have tried that with ShellCheck 0.4.7 and it does not work.

vutny (changes requested):

bootstrap-salt.sh
rallytime: You're right. I was using an old version of shellcheck.

rallytime:
@vutny Can you review one more time?

vutny (approved):
:+1:

PR #1234, opened 2018-05-23
Update to scripting, different pr tests

KaiSforza:
### What does this PR do?

Change from declarative jenkinsfile to scripted jf. Doesn't actually do anything for the kitchen (just echoes) because we don't have any kitchen stuff set up, but it does them in parallel, and limits them for PR's (to 3 os's).


dubb-b (approved):
Looks good

Issue #1207, opened 2018-03-09
Move PR lint tests from old jenkins to new jenkins


rallytime:
The lint checks are set up on a Jenkins master that doesn't do a lot of recent testing besides lint checks for bootstrap PRs. We need to move these checks to the new Jenkins. This will also help start prepping for #1180, as those tests, once created, should run in the new Jenkins, too.

@gtmanfred I just wanted to file this after what we discussed yesterday about moving these to new jenkins instead of using travis like I was originally thinking. Now we can track this task!

gtmanfred:
:+1: sounds good

rallytime:
This is done.

PR #1239, opened 2018-05-31
Shellcheck fixes found when running latest version

rallytime:
### What does this PR do?
Fixes a bunch of shellcheck errors found when running the latest version of shellcheck:
```
shellcheck --version
ShellCheck - shell script analysis tool
version: 0.4.7
license: GNU General Public License, version 3
website: http://www.shellcheck.net
```

Fixes the following error types:
- SC2196: egrep is non-standard and deprecated. Use grep -E instead.
- SC1117: Backslash is literal in "\n". Prefer explicit escaping: "\\n".
- SC2086: Double quote to prevent globbing and word splitting.
- SC2213: getopts specified -Z, but it's not handled by this 'case'.

This is needed as part of #1207, since we'll have a newer versions of shellcheck on the new test runs.

TODO: There is still one major shellcheck error to take care of before a clean run is possible. But this PR is a good first step. Another PR will be needed to clean up the `SC2181: Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.` errors as there are many.

@vutny Can you look over this one as well?

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -400,7 +400,7 @@ EOT
 }   # ----------  end of function __usage  ----------
 
 
-while getopts ':hvnDc:g:Gyx:wk:s:MSNXCPFUKIA:i:Lp:dH:ZbflV:J:j:rR:aq' opt
+while getopts ':hvnDc:g:Gyx:wk:s:MSNXCPFUKIA:i:Lp:dH:bflV:J:j:rR:aq' opt
 do
   case "${opt}" in
 
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -6656,8 +6656,8 @@ fi
 # Drop the master address if passed
 if [ "$_SALT_MASTER_ADDRESS" != "null" ]; then
     [ ! -d "$_SALT_ETC_DIR/minion.d" ] && mkdir -p "$_SALT_ETC_DIR/minion.d"
-    cat <<_eof > $_SALT_ETC_DIR/minion.d/99-master-address.conf
-master: $_SALT_MASTER_ADDRESS
+    cat <<_eof > "$_SALT_ETC_DIR/minion.d/99-master-address.conf
+master: $_SALT_MASTER_ADDRESS"
 _eof
 fi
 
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -61,11 +61,11 @@ __DEFAULT_SLEEP=3
...
-        RC="\033[1;31m"
-        GC="\033[1;32m"
-        BC="\033[1;34m"
-        YC="\033[1;33m"
-        EC="\033[0m"
+        RC="\\033[1;31m"
+        GC="\\033[1;32m"
+        BC="\\033[1;34m"
+        YC="\\033[1;33m"
+        EC="\\033[0m"
```
vutny: Since there are no variable references, we could just use single quotes here, right?

vutny: Since there are no variable references, we could just use single quotes here, right?

bootstrap-salt.sh
```diff
@@ -1102,7 +1103,7 @@ __install_python() {
         exit 1
     fi
 
-    PY_PKG_V=$(echo "$_PY_EXE" | sed -r "s/\.//g")
+    PY_PKG_V=$(echo "$_PY_EXE" | sed -r "s/\\.//g")
```
vutny: Same here, single quotes is the better way to do it.

bootstrap-salt.sh
```diff
@@ -6619,8 +6620,8 @@ fi
 # Drop the master address if passed
 if [ "$_SALT_MASTER_ADDRESS" != "null" ]; then
     [ ! -d "$_SALT_ETC_DIR/minion.d" ] && mkdir -p "$_SALT_ETC_DIR/minion.d"
-    cat <<_eof > $_SALT_ETC_DIR/minion.d/99-master-address.conf
-master: $_SALT_MASTER_ADDRESS
+    cat <<_eof > "$_SALT_ETC_DIR/minion.d/99-master-address.conf
+master: $_SALT_MASTER_ADDRESS"
```
vutny: I think here the quote must end on the file name, but "heredoc" content should remain unquoted.

vutny (changes requested):
Good job on fixing new shellcheck tests, @rallytime.
I have couple of suggestion to make it just little better :smiley:

bootstrap-salt.sh
rallytime: Oh, yep! Thanks!

rallytime:
@vutny Thanks for the review! Can you run through this one more time?

vutny (approved):
Excellent @rallytime !

Issue #1127, opened 2017-08-10
Option -y -x python3 does not install using python3


ninja-:
### Description of Issue/Question

Expected:
bash bootstrap-salt.sh -y -x python3 -M git develop

installs salt with python3 support

Currently:
It will still install using python2.7's pip and the /usr/bin/salt* files will have #!/usr/bin/python instead of python3
 
### Setup
(Please provide relevant configs (Be sure to remove sensitive info).)

### Steps to Reproduce Issue
(Include debug logs if possible, `bootstrap-salt.sh -D`.)

Execute mentioned command and watch for "byte-compiling /usr/lib/python2.7/**..."

### Versions and Systems
(`salt --versions-report`, `bootstrap-salt.sh -v`, system type and version,
cloud/VM provider as appropriate.)
Ubuntu 16.04, latest salt-bootstrap

ninja-:
When running with set -x I can see PY_EXE being set to python3[.5] so I am not sure where things go wrong here.

```
+ _INSTALL_PY=0
+ getopts :hvnDc:g:Gyx:wk:s:MSNXCPFUKIA:i:Lp:dH:ZbflV:J:j:rR:aq opt
+ case "${opt}" in
+ _INSTALL_PY=1
+ getopts :hvnDc:g:Gyx:wk:s:MSNXCPFUKIA:i:Lp:dH:ZbflV:J:j:rR:aq opt
+ case "${opt}" in
+ _PY_EXE=python3.5
+ getopts :hvnDc:g:Gyx:wk:s:MSNXCPFUKIA:i:Lp:dH:ZbflV:J:j:rR:aq opt
+ case "${opt}" in
+ _INSTALL_MASTER=1
+ getopts :hvnDc:g:Gyx:wk:s:MSNXCPFUKIA:i:Lp:dH:ZbflV:J:j:rR:aq opt
+ shift 4
```

ninja-:
Well there's a bunch of issues I see here, all parts of script should use ${PY_EXE} and there is 'python' hardcoded in some parts...

Needs a fix here
https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L2173
https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L2771

I changed these these and still can't get it to work with python3 it crashes with  no module named yaml...

ninja-:
[...]

PR #1242, opened 2018-06-05
 Add Python3 package support for CentOS 7

rallytime:
### What does this PR do?
Adds support to bootstrap to install Salt's Python 3 packages for CentOS 7 using the `-x` option. For example:
```
sh bootstrap-salt.sh -x python3
```

Refs #1127

### Previous Behavior
Could not install Salt with Python 3 packages.

### New Behavior
Can now install Salt with Python 3 packages

@vutny Can you review this one as well?

I tested this with the following options, for the record, and to make sure I didn't regress something:

- `sh bootstrap-salt.sh -x python3`
- `sh bootstrap-salt.sh -x python3 git v2018.3.0`
- `sh bootstrap-salt.sh` (Python 2 stable install)
- `sh bootstrap-salt.sh git v2018.3.0` (Python2 git install)

And it all went well!

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -3632,13 +3632,14 @@ install_centos_stable_deps() {
         yum -y update || return 1
     fi
 
-    if [ "$_DISABLE_REPOS" -eq "$BS_FALSE" ] || ( [ -n "$_PY_EXE" ] && [ "$_PY_MAJOR_VERSION" -eq 3 ] ); then
+    if [ "$_DISABLE_REPOS" -eq "$BS_TRUE" ] && [ -n "$_PY_EXE" ] && [ "$_PY_MAJOR_VERSION" -eq 3 ]; then
+        echoerror "Detected -r or -R option while installing Salt packages for Python 3."
+        echoerror "Python 3 packages for Salt require the EPEL repository to be installed."
+        echoerror "The -r and -R options are incompatible with -x and Python 3 bootstrap installs."
+        return 1
+    fi
 
-        if [ "$_DISABLE_REPOS" -eq "$BS_TRUE" ]; then
-            echowarn "Detected -r or -R option while installing Salt packages for Python 3."
-            echowarn "Python 3 packages for Salt require the EPEL repository to be installed."
-        fi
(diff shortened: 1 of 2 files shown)


bootstrap-salt.sh
```diff
@@ -3627,7 +3632,13 @@ install_centos_stable_deps() {
         yum -y update || return 1
     fi
 
-    if [ $_DISABLE_REPOS -eq $BS_FALSE ]; then
+    if [ "$_DISABLE_REPOS" -eq "$BS_FALSE" ] || ( [ -n "$_PY_EXE" ] && [ "$_PY_MAJOR_VERSION" -eq 3 ] ); then
+
+        if [ "$_DISABLE_REPOS" -eq "$BS_TRUE" ]; then
+            echowarn "Detected -r or -R option while installing Salt packages for Python 3."
+            echowarn "Python 3 packages for Salt require the EPEL repository to be installed."
+        fi
```
vutny: Here I think `-r` or `-R` options would not do what they supposed to do. Because whenever we detect `python3`, the script will forcibly configure the repositories.
I suggest to simplify the logic: below at line 3638 (`if [ "$_CUSTOM_REPO_URL" != "null" ]; then`) which the `-R` option triggers, add inner test for Py3 and add EPEL (+ warning message) on success.
This would be straight forward and more readable.

vutny (changes requested):
Great job @rallytime 

I only have one comment to sort this out.

bootstrap-salt.sh
rallytime: So really what I was trying to do here is rather than fail when the `-r/-R` options are used in tandem with `-x`, just warn the user and keep going.

But maybe a better approach is to exit with a `warnerror`. Because the Python3 packages on repo.saltstack.com for RHEL require EPEL to be enabled. So until that is no longer the case, it's probably best to check for this and fail with a warning (documented) that the arguments are incompatible.

rallytime:
@vutny Thank you for the review! How does this look now?

vutny (approved):
This is good to go :+1:

PR #1246, opened 2018-06-19
Disable shell error about unbound variable during daily install

icy:
### What does this PR do?

Disable shell error about unbound variable during daily install

When using with `daily` option on `ubuntu` system, the `bootstraps` fails due to `shell` error

```
./bootstrap-salt.sh: 2598: ./bootstrap-salt.sh: STABLE_REV: parameter not set
```

### Previous Behavior

The bootstrap fails with `daily` option on Ubuntu system.

### New Behavior

The bootstrap continues with `daily` option on Ubuntu system.


vutny (approved):
Thanks for the fix @icy !

rallytime (approved):

PR #1251, opened 2018-06-24
Allow arm64 installation on Debian systems

### LINKED ISSUE #1240: arm64 installation from repositories is not allowed, even with custom repos

Upstream commit [5] broke Salt bootstrap on AArch64, by
introducing an architecture condition that is too strict to allow
Debian package installation (even if we provide our own repo) at [1].
Note that since most packages required by Salt are arch-indep (with some exceptions for RAET, see [2]),
we actually use the amd64 repos just fine with Salt in OPNFV on AArch64.

We (OPNFV) add "arm64" to the list of supported architectures on the fly to circumvent this,
but this needs some ugly hacks (e.g. [3, 4], as the bootstrap script is fetched using
`curl` from <salt-master-setup.sh> in certain scenarios (e.g. while using salt-formula-salt to
provision the virtual control plane virtual machines).

### Setup
Ubuntu 16.04 (arm64), but it applies to any Debian distro on arm64 really.

### Steps to Reproduce Issue
N/A

### Versions and Systems
2018.04.25 (but it affects everything past commit [5]).

[1] https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L1293
[2] https://jira.opnfv.org/browse/ARMBAND-382
[3] https://github.com/opnfv/fuel/blob/master/mcp/patches/scripts/0001-salt-master-setup.sh-Allow-arm64-salt-bootstrap.patch
[4] https://github.com/opnfv/armband/blob/master/patches/opnfv-fuel/0003-mcp-salt-formulas-armband-Extend-libvirt_domain.patch#L85
[5] https://github.com/saltstack/salt-bootstrap/commit/caa6d7d

alexandruavadanii:
Previously, AArch64 machines ('arm64' DPKG architecture) were not
supported due to missing repository metadata.

- relax bail-out condition in DPKG machine architecture check to
  allow arm64; print a warning about experimental support;
- if no custom repo is defined, default to re-using amd64 repo on
  arm64 since all packages are arch-independent;

Fixes: https://github.com/saltstack/salt-bootstrap/issues/1240

Signed-off-by: Alexandru Avadanii <Alexandru.Avadanii@enea.com>

### What does this PR do?
Allows salt-bootstrap to run on arm64 (AArch64) Debian systems. The default repository used on such machines is the amd64 one, since all packages are architecture independent.

https://github.com/saltstack/salt-bootstrap/issues/1240

### Previous Behavior
Bootstrapping Salt on arm64 machines was not possible, as the script bailed during __dpkg_check_architecture with a return code of 1.

### New Behavior
arm64 bootstrapping is allowed, a warning is printed that this is experimental.
If custom repos are used, they should have the expected structure and contents.
If repos are left to default, amd64 (with an extra '[deb=amd64]' prefixed to the URI) are used, allowing us to re-use arch-independent Salt packages.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1315,6 +1315,7 @@ __check_dpkg_architecture() {
     fi
 
     __REPO_ARCH="$DPKG_ARCHITECTURE"
+    __REPO_ARCH_DEB='deb'
     __return_code=0
 
     case $DPKG_ARCHITECTURE in
@@ -1324,6 +1325,18 @@ __check_dpkg_architecture() {
             __REPO_ARCH="amd64"
             ;;
         "amd64")
+            error_msg=""
+            ;;
+        "arm64")
+            if [ "$_CUSTOM_REPO_URL" != "null" ]; then
+                warn_msg="Support for arm64 is experimental, make sure the custom repository used has the expected structure and contents."
+            else
+                # Saltstack official repository does not yet have arm64 metadata,
+                # use amd64 repositories on arm64, since all pkgs are arch-independent
+                __REPO_ARCH="amd64"
+                __REPO_ARCH_DEB="deb [arch=$__REPO_ARCH]"
+                warn_msg="Support for arm64 packages is experimental and might rely on architecture-independent packages from the amd64 repository."
+            fi
             error_msg=""
             ;;
         "armhf")
@@ -1340,6 +1353,11 @@ __check_dpkg_architecture() {
             ;;
     esac
 
+    if [ "${warn_msg}" != "" ]; then
+        # AArch64: Do not fail at this point, but warn the user about experimental support
+        # See https://github.com/saltstack/salt-bootstrap/issues/1240
+        echowarn "${warn_msg}"
+    fi
     if [ "${error_msg}" != "" ]; then
         echoerror "${error_msg}"
         if [ "$ITYPE" != "git" ]; then
@@ -2611,7 +2629,7 @@ __install_saltstack_ubuntu_repository() {
 
     # SaltStack's stable Ubuntu repository:
     SALTSTACK_UBUNTU_URL="${HTTP_VAL}://${_REPO_URL}/${__PY_VERSION_REPO}/ubuntu/${UBUNTU_VERSION}/${__REPO_ARCH}/${STABLE_REV}"
-    echo "deb $SALTSTACK_UBUNTU_URL $UBUNTU_CODENAME main" > /etc/apt/sources.list.d/saltstack.list
+    echo "$__REPO_ARCH_DEB $SALTSTACK_UBUNTU_URL $UBUNTU_CODENAME main" > /etc/apt/sources.list.d/saltstack.list
 
     __apt_key_fetch "$SALTSTACK_UBUNTU_URL/SALTSTACK-GPG-KEY.pub" || return 1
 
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -1324,6 +1324,18 @@ __check_dpkg_architecture() {
...
             error_msg=""
             ;;
+        "arm64")
+            if [ "$_CUSTOM_REPO_URL" != "null" ]; then
+                warn_msg="Support for arm64 is experimental, make sure the custom repository used has the expected structure and contents."
+            else
+                # Saltstack official repository does not yet have arm64 metadata,
+                # use amd64 repositories on arm64, since all pkgs are arch-independent
+                __REPO_ARCH="amd64"
+                __REPO_ARCH_DEB="[arch=$__REPO_ARCH]"
```
vutny: I think we should declare some default value for `__REPO_ARCH_DEB` and redefine it here.
Let it be `__REPO_ARCH_DEB='deb'` before the case statement and here we would assign it
`deb [arch=$__REPO_ARCH]` value.

bootstrap-salt.sh
```diff
@@ -2618,7 +2634,7 @@ __install_saltstack_ubuntu_repository() {
 
     # SaltStack's stable Ubuntu repository:
     SALTSTACK_UBUNTU_URL="${HTTP_VAL}://${_REPO_URL}/${__PY_VERSION_REPO}/ubuntu/${UBUNTU_VERSION}/${__REPO_ARCH}/${STABLE_REV}"
-    echo "deb $SALTSTACK_UBUNTU_URL $UBUNTU_CODENAME main" > /etc/apt/sources.list.d/saltstack.list
+    echo "deb $__REPO_ARCH_DEB $SALTSTACK_UBUNTU_URL $UBUNTU_CODENAME main" > /etc/apt/sources.list.d/saltstack.list
```
vutny: So here and below just use:
```
echo "$__REPO_ARCH_DEB $SALTSTACK_UBUNTU_URL ...
```

bootstrap-salt.sh
```diff
@@ -1338,7 +1350,11 @@ __check_dpkg_architecture() {
             ;;
     esac
 
-    if [ "${error_msg}" != "" ]; then
+    if [ "${warn_msg}" != "" ]; then
+        # AArch64: Do not fail at this point, but warn the user about experimental support
+        # See https://github.com/saltstack/salt-bootstrap/issues/1240
+        echowarn "${warn_msg}"
+    elif [ "${error_msg}" != "" ]; then
```
vutny: Do we really want to display either `{warn_msg}` or `${error_msg}`? Maybe better to output them both when they are defined...

bootstrap-salt.sh
```diff
@@ -1324,6 +1324,18 @@ __check_dpkg_architecture() {
         "amd64")
             error_msg=""
             ;;
+        "arm64")
+            if [ "$_CUSTOM_REPO_URL" != "null" ]; then
+                warn_msg="Support for arm64 is experimental, make sure the custom repository used has the expected structure and contents."
+            else
+                # Saltstack official repository does not yet have arm64 metadata,
+                # use amd64 repositories on arm64, since all pkgs are arch-independent
+                __REPO_ARCH="amd64"
```
vutny, line 1336: Probably adding `amd64` as a "foreign" architecture is missing here.
You need to check if it is not listed in `dpkg --print-foreign-architectures` and then add it with the
`dpkg --add-architecture` command.

vutny (changes requested):
Thanks @alexandruavadanii , this is good PR.

I have few of suggestions to implement and it would be great. Please see my inline comments.

rallytime:
Hi @alexandruavadanii - did you get a chance to review the feedback above?

alexandruavadanii:
Hi and sorry for the lack of activity, I've been on vacation. I'll try to address all of the above when I get back.

bootstrap-salt.sh
alexandruavadanii: Adding amd64 as a foreign architecture is not needed. I actually think that would be a bad idea to apply here, as we don't want any x86 binaries on arm64, we just tell APT to fetch the repository metadata from the x86-specific URL (which includes 'binary-amd64'), but we only care about the arch-independent packages in that repo.

alexandruavadanii (commented):
Thank you, I will push -f shortly.

rallytime:
@vutny Can you give this another look?

bootstrap-salt.sh
vutny: Oh, you're absolutely right, indeed. First I thought that there might be some arch specific dependencies between the packages which are mandatory for `apt` to proceed. It was some issue with `i386` arch in Salt 2015.8 for whezzy back when I believe...
Now things are much cleaner in SaltStack's repo. Thanks for pointing me out on this!

vutny (approved):
@alexandruavadanii Looks great.

@rallytime This is ready to merge.

rallytime (approved):

PR #1256, opened 2018-07-10
Ensure pipe is absent before mkfifo/mknod

noelmcloughlin:
### What does this PR do?

Before invoking `mknod/mkfifo`  ensure `pipe` is absent on the file system.

When `bootstrap-salt.sh` script terminates abnormally (SIGKILL, kernel crash, etc), subsequent runs could fail - in my case, a FreeBSD VM crash (caused by a [ACPI kernel bug](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=209222))
```
kernel: ACPI Error No installed handler for fixed event - P_Timer, disabling
kernel: ACPI Error No installed handler for fixed event - Realtimeclock, disabling
kernel: ACPI Error No installed handler for fixed event - GPE 0
etc ... <fatal kernel crash>
``` 
### Previous Behavior

Execution Error:
```
root@:~ # sh bootstrap_salt.sh
 * ERROR: Failed to create the named pipe required to log
root@:~ # sh bootstrap_salt.sh -D
 * ERROR: Failed to create the named pipe required to log
```

### New Behavior

The error is gone.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -457,6 +457,8 @@ shift $((OPTIND-1))
 # Define our logging file and pipe paths
 LOGFILE="/tmp/$( echo "$__ScriptName" | sed s/.sh/.log/g )"
 LOGPIPE="/tmp/$( echo "$__ScriptName" | sed s/.sh/.logpipe/g )"
+# Ensure no residual pipe exists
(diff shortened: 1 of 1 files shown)


rallytime (approved):
👍

PR #1262, opened 2018-07-11
Remove support for openSUSE Leap 42.2

rallytime:
Leap 42.2 is EOL and has been replaced by 42.3

This commit also simplifies some of the logic in checking for suse versions.

openSUSE 13.X is no longer supported and hasn't been for some time, so we don't need to worry about checking for those versions any longer. The script will bail out at the `__check_end_of_life_versions` functions early on in the install process.

I will have a couple more changes coming in today for some other suse support items, but this is a good start and is a discrete change that can be it's own PR.

@noelmcloughlin You've been working in this space recently. Can you review this one?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -266,7 +266,7 @@ Red Hat family
 SUSE family
 ~~~~~~~~~~~
 
-- openSUSE Leap 42.2/42.3
+- openSUSE Leap 42.3
 - openSUSE Tumbleweed 2015
 - SUSE Linux Enterprise Server 11 SP4, 12 SP2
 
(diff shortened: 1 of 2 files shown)


noelmcloughlin:
LTGM!! Improvement over my conservatives update in #1244. Approved!

btw, I don't have write access on this repo.

README.rst
```diff
@@ -266,7 +266,7 @@ Red Hat family
 SUSE family
 ~~~~~~~~~~~
 
-- openSUSE Leap 42.2/42.3
+- openSUSE Leap 42.3
```
vutny, line 269: I guess Leap releases `15.0` and `15.1` are missing here, since the script attempts to proceed on them when detected.

rallytime: Yeah, I am working on 15 support today. @noelmcloughlin Made some improvements there already, but I am making it a bit more robust with git support, etc. So I will update the README about that in another PR.

PR #1265, opened 2018-07-20
Updting checkstyle to fail on warnings

dubb-b:
### What does this PR do?
Fixes the checkstyle plugin to fail on Warnings

PR's not failing shellcheck tests

Im pretty sure this will fail anything that is a warning.


dubb-b:
@rallytime view the build and see if that is what you are after.

rallytime:
Yep! That looks good. Thanks for fixing that. I have created https://github.com/dubb-b/salt-bootstrap/pull/1 to fix the errors so we can see them pulled in and turn the build back to green. Please merge my PR into your branch when you get a moment and then we should be able to see those changes pull in here.

rallytime (approved):

PR #1283, opened 2018-08-25
FIX #1237 SmartOS should use pkgin show-deps

### LINKED ISSUE #1237: SmartOS dependency conflict
The function which installs dependencies on SmartOS (`install_smartos_deps`) explicitly installs the package `py27-crypto`. However when pkgsrc is used later on to install the `salt` package this fails because the `salt` package now depends on `py27-cryptodome` which conflicts with `py27-crypto`.

The 18.1.0 SmartOS image (` 221635c4-3b85-11e8-b6ba-23f68c9bf2c4`) has a recent enough version of pkgsrc for this problem to manifest itself.

sjorge:
### What does this PR do?
The install_smartos_deps function is used for both stable and git.
Only the m2crypt package is not included in the automatic dependacies
for the salt package.

We simply query the salt package to figure out the dependancies. That we
we have the right set for both git and stable. Even if pacakge names
change in the future.

#1237 

### Previous Behavior
Hardcoded list of dependacies

### New Behavior
We query the OS provided package so we always grab the right set of dependacies.

### Testing notes
This was minimally tested on in a zone only using both stable and git and it seems to work.
I did hit a error a few times because the error handling makes no difference between package install failure and config file download in the function. Which made me wast a bit of time as the zone was originally in a super restricted environment.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -322,7 +322,7 @@ UNIX systems
 
 **SunOS**:
 
-- SmartOS
+- SmartOS (2015Q4 and later)
 
 Unsupported Distributions
 -------------------------
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5336,7 +5336,8 @@ install_openbsd_restart_daemons() {
 #   SmartOS Install Functions
 #
 install_smartos_deps() {
-    pkgin -y install zeromq py27-crypto py27-m2crypto py27-msgpack py27-yaml py27-jinja2 py27-zmq py27-requests || return 1
+    smartos_deps="$(pkgin show-deps salt | grep '^\s' | grep -v '\snot' | xargs) py27-m2crypto"
+    pkgin -y install "${smartos_deps}" || return 1
 
     # Set _SALT_ETC_DIR to SmartOS default if they didn't specify
     _SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/opt/local/etc/salt}


sjorge:
@rallytime sorry it took so long for me to test this and provide a PR, been rather busy at work and such so I spend less time behind the computer at home.

rallytime:
No problem @sjorge!

There is a lint error here that will need to be fixed before moving forward, however. Can you take a look?

https://jenkinsci.saltstack.com/job/salt-bootstrap/job/PR-1283/1/checkstyleResult/

sjorge:
I guess in this case we do want word splitting though. According to some googling this seems to be the correct way to make shellcheck happy in this case. *crosses fingers*

sjorge:
Ugh this made it 10x worse -_-

sjorge:
Newish pkgin seems to be OK with this, I think this will not work on older zones (2014Q4) because I had issues with that when I was scripting my installs before I used salt.

Then again I don't think salt will work on those either, so that is probably OK.

rallytime:
@sjorge Maybe that would be a good thing to document in the [README file](https://github.com/saltstack/salt-bootstrap/blob/develop/README.rst) under the Unix Support section?

sjorge:
I added a version to the README, older zone I had access to right now is 2016Q4 which bootstraps fine. 

I know the stable version shipped for 2015Q4 also works, so I put the mininum at that. I also double checked the stable package is still available: https://pkgsrc.joyent.com/packages/SmartOS/2015Q4/x86_64/All/ which is is, salt-2015.8.12.tgz  was relatively recently update. Given the YYYYQ4 are LTS for 4(?) years.

rallytime (approved):
Thanks for this @sjorge!

Issue #1194, opened 2018-01-19
Doesn't work with Amazon Linux 2 ami-4f89f53


nodakai:
### Description of Issue/Question

```
$ sudo bash bootstrap-salt.sh
 *  INFO: Running version: 2017.12.13
 *  INFO: Executed by: bash
 *  INFO: Command line: 'bootstrap-salt.sh '

 *  INFO: System Information:
 *  INFO:   CPU:          GenuineIntel
 *  INFO:   CPU Arch:     x86_64
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   4.9.76-38.79.amzn2.x86_64
 *  INFO:   Distribution: Amazon Linux AMI 2.0

 * ERROR: End of life distributions are not supported.
 * ERROR: Please consider upgrading to the next stable. See:
 * ERROR:     https://aws.amazon.com/amazon-linux-ami/
```

`bash -x` displays sth like this

```
+ __check_end_of_life_versions
+ case "${DISTRO_NAME_L}" in
+ '[' 2 -lt 2012 ']'
+ echoerror 'End of life distributions are not supported.'
```

Apparently the script picked up 2 from "2.0" and complained it's less than 2012.
* https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L1606

Given the new versioning scheme for Amazon Linux 2 series, maybe anything below 10 should pass the check?

### Versions and Systems
```
$ bash install-salt.sh -v
install-salt.sh -- Version 2017.12.13
```

```
$ cat /etc/system-release
Amazon Linux release 2.0 (2017.12) LTS Release Candidate
$ cat /etc/os-release
NAME="Amazon Linux"
VERSION="2.0 (2017.12)"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2.0"
PRETTY_NAME="Amazon Linux 2.0 (2017.12) LTS Release Candidate"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2.0"
HOME_URL="https://amazonlinux.com/"
```

nodakai:
Hmm, the RPM doesn't support Amazon Linux 2 yet after all... saltstack/salt#45154

rallytime:
@nodakai my apologies for the late response here, but yeah, the packages for Amazon Linux 2 haven't been released yet. Once they are, we definitely should get the bootstrap script to support it.

[...]

PR #1287, opened 2018-09-14
use official amazon linux 2 repo for amazon linux 2

mchugh19:
### What does this PR do?
~"Supports" Amazon Linux 2 by using the saltstack centos7 yum repo
Don't know if this is an upstream-able change, but it works for us.~
Update to use new amazon linux 2 yum repo

#1194

### Previous Behavior
bootstrap-salt.sh fails on amazon linux 2

### New Behavior
bootstrap-salt.sh successfully installs and runs salt-minion

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -46,6 +46,11 @@ platforms:
       run_command: /sbin/init
       provision_command:
         - yum install -y upstart
+  - name: amazon-2
+    driver_config:
+      image: amazonlinux:2
+      platform: rhel
+      run_command: /usr/lib/systemd/systemd
   - name: ubuntu-18.04
     driver_config:
       run_command: /lib/systemd/systemd
@@ -83,6 +88,8 @@ suites:
   - name: py2-git-2017.7
     provisioner:
       salt_version: 2017.7
+    excludes:
+      - arch
   - name: py2-git-2018.3
     provisioner:
       salt_version: 2018.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ env:
   - PLATFORM=centos-7
   - PLATFORM=centos-6
   - PLATFORM=amazon-1
+  - PLATFORM=amazon-2
   - PLATFORM=ubuntu-1804
   - PLATFORM=ubuntu-1604
   - PLATFORM=ubuntu-1404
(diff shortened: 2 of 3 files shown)


rallytime:
Hi @mchugh19! My apologies for taking so long to respond here. I see you have a checkstyle error that needs to be fixed.

However, the salt repo has added amazon linux 2 support now: http://repo.saltstack.com/#amzn

I think we should use the official packages, rather than the CentOS 7 method you've provided here. Thoughts?

mchugh19:
I think this should resolve the lint error and also use the new official amazon linux 2 repo. 

I wouldn't mind a double-check on this as I don't have the ability to spin up an ec2 instance at the moment, and it is copying a large block of code from the amazon linux 1 deps function (with some package name updates).

rallytime:
Hey @mchugh19 - I added a test to our runner for the new amazonlinux 2 builds and it looks like there's an error here. The test is failing. Can you take a look?

https://travis-ci.com/saltstack/salt-bootstrap/jobs/158212816

mchugh19:
@rallytime Gah! Looks like it was a missing trailing slash typo. Let's see how travis likes this version...

mchugh19:
Looks like the minion was successfully installed in the amazon linux 2 test, but had an issue on the integration tests. So I think it validated what this PR covers, but let me know if something else is needed.

rallytime:
@mchugh19 The reason this is failing is because you didn't add support for the `git` install options, only the stable options. So something like `sh bootstrap-salt.sh git v2018.3.3` isn't working. The minion and master daemons can't start because there are packages missing. Would you be willing to add support for the git option as well? There should be some examples to follow of amazon linux 1 that just need to be adapted a bit for amazon linux 2.

The stable options are working swimmingly! So thank you for adding that support.

rallytime:
@mchugh19 Getting closer! Thank you for adding that support. Looks like a couple of tweaks may still be needed: https://travis-ci.com/saltstack/salt-bootstrap/jobs/158600374

mchugh19:
Setup free tier aws to validate amazon linux 2. Git installs should be fixed now.

rallytime:
This looks great! Thanks @mchugh19!

rallytime (approved):

Issue #1288, opened 2018-09-26
salt-cloud minion bootstrap fails due to dpkg lock


ripesensor:
### Description of Issue/Question

Same issue as #1163 when bootstrapping an Amazon Deep Learning AMI with salt-cloud.  

Error:
<pre><code>
 *  INFO: Running version: 2018.08.15
 *  INFO: Executed by: /bin/sh
 *  INFO: Command line: '/tmp/.saltcloud-64aafe3b-1eff-47e5-a70c-4d58e2c227b0/deploy.sh -c /tmp/.saltcloud-64aafe3b-1eff-47e5-a70c-4d58e2c227b0 -D'

 *  INFO: System Information:
 *  INFO:   CPU:          GenuineIntel
 *  INFO:   CPU Arch:     x86_64
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   4.4.0-1066-aws
 *  INFO:   Distribution: Ubuntu 16.04

 * DEBUG: Binaries will be searched using the following $PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
 *  INFO: Installing minion
 * DEBUG: install_ubuntu_16_stable_deps not found....
 * DEBUG: install_ubuntu_16_04_stable_deps not found....
 * DEBUG: install_ubuntu_16_deps not found....
 * DEBUG: install_ubuntu_16_04_deps not found....
 *  INFO: Found function install_ubuntu_stable_deps
 * DEBUG: DEPS_INSTALL_FUNC=install_ubuntu_stable_deps
 * DEBUG: config_ubuntu_16_stable_salt not found....
 * DEBUG: config_ubuntu_16_04_stable_salt not found....
 * DEBUG: config_ubuntu_16_salt not found....
 * DEBUG: config_ubuntu_16_04_salt not found....
 * DEBUG: config_ubuntu_stable_salt not found....
 * DEBUG: config_ubuntu_salt not found....
 *  INFO: Found function config_salt
 * DEBUG: CONFIG_SALT_FUNC=config_salt
 * DEBUG: preseed_ubuntu_16_stable_master not found....
 * DEBUG: preseed_ubuntu_16_04_stable_master not found....
 * DEBUG: preseed_ubuntu_16_master not found....
 * DEBUG: preseed_ubuntu_16_04_master not found....
 * DEBUG: preseed_ubuntu_stable_master not found....
[...]

PR #1291, opened 2018-11-01
Modify wait_for_apt function

ripesensor:
### What does this PR do?
Modifies the __wait_for_apt function (and others based on it) to rely on the apt exit code to know if the apt process is locked or not.

#1288 

### Previous Behavior
__wait_for_apt() would do the right thing, but it was possible for another apt process to sneak in-between the function and the following apt call.

### New Behavior
Loop our apt commands until our exit code != 100.

May want to edit the comment for the function if this is a worthwhile PR.


bootstrap-salt.sh
```diff
@@ -18,7 +18,7 @@
 #======================================================================================================================
 set -o nounset                              # Treat unset variables as an error
 
-__ScriptVersion="2018.08.15"
+__ScriptVersion="2018.11.01"
```
vutny: Please do not bump the version number here, it would be done by a maintainer from SaltStack when the stable version will come out.

bootstrap-salt.sh
```diff
@@ -1798,24 +1798,26 @@ __function_defined() {
...
-    echodebug "Checking if apt process is currently running."
-
     # Timeout set at 15 minutes
     WAIT_TIMEOUT=900
 
-    while ps -C apt,apt-get,aptitude,dpkg >/dev/null; do
-        sleep 1
-        WAIT_TIMEOUT=$((WAIT_TIMEOUT - 1))
+    # Set an initial exit code to 100; which matches the exit code of a locked apt process
+    (exit 100)
```
vutny: I think, could we execute the passed in command first, instead of simulating "locked" exit code?
This way no need to wait an additional second for each `apt` call. Although it requires to save the exit code and return it later if not equals `100`.

vutny (changes requested):
This is great, @ripesensor !
Much cleaner implementation which is not relying on any external tool, but only on the apt utils return code.
I just have few minor comments. Thanks!

bootstrap-salt.sh
ripesensor: I guess I was trying to not have repeated code, but it also makes more sense to at least try it instead of faking it.

bootstrap-salt.sh
ripesensor: Got it, thanks.

bootstrap-salt.sh
```diff
@@ -1798,24 +1798,29 @@ __function_defined() {
...
+          return 1
+      else
+          # Try running apt again until our return code != 100
+	  "${@}"
+    	  APT_RETURN=$?
+      fi
     done
-
-    echodebug "No apt processes are currently running."
 }
```
vutny, line 1826: Looks like you need to return the apt exit code in the end. It is required for the caller to be able to determine if it was not lock-related failure.
```suggestion

    return $APT_RETURN
}
```

vutny (approved):
Looks great!

rallytime (approved):

PR #1293, opened 2018-11-15
Split up some tests to avoid travis timeout

rallytime:
This PR splits up some of the test platforms into multiple instances so we can avoid hitting the timeout errors.

Some distros were in the "allowed failures" category because sometimes they would hit timeouts.

By splitting the distros up into `git` and `stable` tests, we use more executors, but we allow the tests to complete all of the time.


rallytime:
Oh, heh. I made some assumptions here about how this works and this isn't correct. Closing.

PR #1296, opened 2018-11-15
Add Py3 tests for supported platforms (git installs)

rallytime:
- Ubuntu 16


rallytime:
Bah, these will need to be added later. I need to learn more about kitchen-salt for this to work correctly.

PR #1298, opened 2018-11-29
Fixes debian SysV init scripts

twe-syde:
### What does this PR do?
- removes download of init scripts
- instead copy them from source
- copy them from correct source
- removes Debian >=8 equals systemd comments

- on sysv init systems (e.g. docker images) the installation of the scripts fails and thus the service check.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -3320,16 +3320,16 @@ install_debian_git_post() {
         [ "$fname" = "minion" ] && [ "$_INSTALL_MINION" -eq $BS_FALSE ] && continue
         [ "$fname" = "syndic" ] && [ "$_INSTALL_SYNDIC" -eq $BS_FALSE ] && continue
 
-        # Configure SystemD for Debian 8 "Jessie" and later
+        # Install SystemD services
         if [ -f /bin/systemctl ]; then
             if [ ! -f /lib/systemd/system/salt-${fname}.service ] || \
                 { [ -f /lib/systemd/system/salt-${fname}.service ] && [ $_FORCE_OVERWRITE -eq $BS_TRUE ]; }; then
-                if [ -f "${_SALT_GIT_CHECKOUT_DIR}/pkg/salt-${fname}.service" ]; then
-                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/salt-${fname}.service" /lib/systemd/system
-                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/salt-${fname}.environment" "/etc/default/salt-${fname}"
+                if [ -f "${_SALT_GIT_CHECKOUT_DIR}/pkg/deb/salt-${fname}.service" ]; then
+                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/deb/salt-${fname}.service" /lib/systemd/system
+                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/deb/salt-${fname}.environment" "/etc/default/salt-${fname}"
                 else
                     # workaround before adding Debian-specific unit files to the Salt main repo
-                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/salt-${fname}.service" /lib/systemd/system
+                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/deb/salt-${fname}.service" /lib/systemd/system
                     sed -i -e '/^Type/ s/notify/simple/' /lib/systemd/system/salt-${fname}.service
                 fi
             fi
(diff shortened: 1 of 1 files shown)


Ch3LL:
thanks for this @pluhmen :) just need to clarify something with our packager.

ping @dmurphy18 is `salt/pkg/deb/salt-*.service` the accurate service files now with salt-pack in the picture?

dmurphy18:
@Ch3LL @pluhmen All Salt building & packaging uses salt-pack for Python 2 and salt-pack-py3 for Python 3.  At some point when time allows, I need to refer the changes back into salt on GitHub, however both salt-pack and salt-pack-py3 are open source and contributions are gladly accepted.

Come April once Ubuntu 14.04 EOL's I intend to do so.

Ch3LL:
@dmurphy18  can you point to me where the debian file is on salt-pack? I could not find it using a find command: `find ~/git/salt-pack/file_roots -name \*serv\* |grep -i deb` but it found the service files for the other operating systems. I am just trying to diff the current service file we use in salt-pack and the one in the salt repo to make sure we aren't missing anything important before we accept these changes into the bootstrap.

Ch3LL:
ping @dmurphy18 see above ^

dmurphy18:
@Ch3LL @pluhmen The files you want are in the spec directory as a tarbal, expand , edit, re-compress.
Note the following is for the nightly build on the current 2018.3 branch, there are similar for 2017.7 branch and 2018.11 branch is supposed to be Fluorine but that version number could change given we are in December and still to do a release candidate.
https://github.com/saltstack/salt-pack/blob/develop/file_roots/pkg/salt/2018_3/debian8/spec/salt_debian.tar.xz

There are similar files for Debian 9 and versions of Ubuntu supported.  The tagged release is generated from these nightly build branches.  Fortunately SaltStack does not support Debian 8 on Python 3, so changes are limited to salt-pack.

There is a task to replace the pkg files in salt with current versions and update the files as examples with text directing to salt-pack and salt-pack-py3, but that is a low lying task at the moment.

Ch3LL:
@dmurphy18 thanks for pointing that out. When i extract that file the only reference to a salt-<service>.service file is in `salt-master.install` and it shows: `pkg/salt-master.service /lib/systemd/system`. Is that just grabbing that from the salt repo anyways?

Also when i diff the init files between what is in salt-pack and whats in salts repo they are only differing in the names. So as long as you can clear up the above question I think i'm okay with these changes.

bootstrap-salt.sh
```diff
@@ -3320,16 +3320,16 @@ install_debian_git_post() {
...
-                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/salt-${fname}.service" /lib/systemd/system
-                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/salt-${fname}.environment" "/etc/default/salt-${fname}"
+                if [ -f "${_SALT_GIT_CHECKOUT_DIR}/pkg/deb/salt-${fname}.service" ]; then
+                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/deb/salt-${fname}.service" /lib/systemd/system
+                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/deb/salt-${fname}.environment" "/etc/default/salt-${fname}"
                 else
                     # workaround before adding Debian-specific unit files to the Salt main repo
-                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/salt-${fname}.service" /lib/systemd/system
+                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/deb/salt-${fname}.service" /lib/systemd/system
                     sed -i -e '/^Type/ s/notify/simple/' /lib/systemd/system/salt-${fname}.service
```
vutny, line 3333: Since the installation of `python*-systemd` package on Debian >= 8 is mandatory, no need for this `sed` anymore I think.

dmurphy18:
@Ch3LL the files in the tar ball are the ones used, hence init file changes should also be made in salt-pack (and salt-pack-py3).  The systemd service files used are from salt repo, hence for salt-master.install it uses salt's pkg/salt-master.service.  Any changes in Salt's deb/files needs to be reflected back into salt-pack and salt-pack-py3.  I usually go over them before a release to make sure all is in sync.

Ch3LL:
thanks for clarifying @dmurphy18 i think i'm okay with these changes as the init files are pretty much the same, and as you stated the service file is the one used in salt's repo, although we will probably want to keep the `pkg/salt-master.service` path instead of `pkg/deb/..`

bootstrap-salt.sh
```diff
@@ -3320,16 +3320,16 @@ install_debian_git_post() {
...
 
-        # Configure SystemD for Debian 8 "Jessie" and later
+        # Install SystemD services
         if [ -f /bin/systemctl ]; then
             if [ ! -f /lib/systemd/system/salt-${fname}.service ] || \
                 { [ -f /lib/systemd/system/salt-${fname}.service ] && [ $_FORCE_OVERWRITE -eq $BS_TRUE ]; }; then
-                if [ -f "${_SALT_GIT_CHECKOUT_DIR}/pkg/salt-${fname}.service" ]; then
-                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/salt-${fname}.service" /lib/systemd/system
-                    __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/salt-${fname}.environment" "/etc/default/salt-${fname}"
+                if [ -f "${_SALT_GIT_CHECKOUT_DIR}/pkg/deb/salt-${fname}.service" ]; then
```
Ch3LL, line 3327: The debian packages we build are using `pkg/salt-${fname}.service` instead of `pkg/deb` So I think its better to keep this at that path to ensure we use any changes that might occur.

Ch3LL (changes requested):

vutny (commented):
I think this PR was superseded by recently merged #1359 .
Since it is outdated, I think we can close this out @s0undt3ch .

Thanks for your contribution, anyway @pluhmen !

Issue #1299, opened 2018-12-29
Bunsenlabs Debian derivative support


kevinquinnyo:
### Description of Issue/Question
Doesn't support Bunsenlabs debian derivative

### Setup
(Please provide relevant configs (Be sure to remove sensitive info).)
```
$> sudo find /etc/ -name \*-release -print -exec cat {} \;
/etc/lsb-release
DISTRIB_ID="BunsenLabs"
DISTRIB_DESCRIPTION="BunsenLabs GNU/Linux 9.5 (Helium)"
DISTRIB_RELEASE="9.5"
DISTRIB_CODENAME="helium"
/etc/os-release
ANSI_COLOR="0;34"
BUG_REPORT_URL="https://github.com/BunsenLabs"
HOME_URL="https://bunsenlabs.org"
ID=bunsenlabs
ID_LIKE="debian"
NAME="BunsenLabs GNU/Linux"
PRETTY_NAME="BunsenLabs GNU/Linux 9.5 (Helium)"
SUPPORT_URL="https://forums.bunsenlabs.org"
VERSION="9.5 (Helium)"
VERSION_ID="9.5"
[22:52:33] ~/dev/kenv 
$> command lsb_release -a
No LSB modules are available.
Distributor ID:	BunsenLabs
Description:	BunsenLabs GNU/Linux 9.5 (Helium)
Release:	9.5
Codename:	helium
```
### Steps to Reproduce Issue
(Include debug logs if possible, `bootstrap-salt.sh -D`.)
Use this OS, and try the commands listed here:
https://docs.saltstack.com/en/latest/topics/tutorials/quickstart.html#bootstrap-salt-minion

Which are at the time of writing:
```
curl -L https://bootstrap.saltstack.com -o bootstrap_salt.sh
sudo sh bootstrap_salt.sh
```
### Versions and Systems
(`salt --versions-report`, `bootstrap-salt.sh -v`, system type and version,
cloud/VM provider as appropriate.)

N/A

So this is pretty minor and I'm using a derivative debian distro that is probably very uncommon.  I was able to resolve this by patching the `bootstrap_salt.sh` file I downloaded via the url listed above, but it seems to be different than the `bootstrap-salt.sh` file listed in this repo.  Probably a difference in the branch/release.

Anyway, the diff that fixes this for me is simple:

```
$> diff bootstrap_salt.sh /tmp/bootstrap_salt.sh 
1428c1428
[...]

PR #1300, opened 2018-12-29
Add support for bunsenlabs debian derivative

kevinquinnyo:
### What does this PR do?
Adds support for bunsenlabs distro (a superficial debian derivative).

I did not bother adding support for bunsenlabs (formerly crunchbang) that are based on older debians, because I'm using 9.

~~I don't think it would be hard to support that with a few changes, but I don't know how many people besides me even use this "distro".~~ I doubt any changes would be needed, because this "distro" is just plain debian with a pre-configured openbox for the UI basically.

https://github.com/saltstack/salt-bootstrap/issues/1299

### Previous Behavior
Failed to detect OS info:
```

```

### New Behavior
Installs without issues.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1431,7 +1431,7 @@ __debian_derivatives_translation() {
     # If the file does not exist, return
     [ ! -f /etc/os-release ] && return
 
-    DEBIAN_DERIVATIVES="(cumulus_.+|devuan|kali|linuxmint|raspbian)"
+    DEBIAN_DERIVATIVES="(cumulus_.+|devuan|kali|linuxmint|raspbian|bunsenlabs)"
     # Mappings
     cumulus_2_debian_base="7.0"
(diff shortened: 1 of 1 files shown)


kevinquinnyo:
Now that I've gotten masterless salt installed, I think this PR is probably not a good idea without additional support or documentation.  In my case, `os_family` is still "Bunsenlabs" post install (obviously) so things like pkg.installed fail.

It would be an OK idea if the supplemental changes were made to detect "Bunsenlabs" as debian, but obviously saltstack can't support every ridiculous rebranded linux distro.

Feel free to close, but if anyone knows the best practice for this situation, let me know.

Ch3LL:
@kevinquinnyo the situation you are explaining with this distro not working with salt is not related to the bootstrap, but would need to be added into salts code base, so i am okay merging this. i'm just going to update the branch and have the tests re-run.

Ch3LL (approved):

PR #1311, opened 2019-01-28
Adding changes for kitchen pytest params and the drone build.

dubb-b:
### What does this PR do?
Updates pytest not to cache locally. This should not matter unless testing locally.

Also making ubuntu builds run and not ignore failure.


Ch3LL (approved):

dubb-b:
@Ch3LL you can merge this. There are some issues that we found with performance I am working through with Drone. But if I run all of these individually they all pass.

PR #1316, opened 2019-02-06
Fix bootstrap install for git/develop on FreeBSD

hackacad:
### What does this PR do?
1Fix bootstrap install for git/develop on FreeBSD

none

### Previous Behavior
1: dependency detecting on FreeBSD was broken
2: installation failed when python meta package wasn't installed

### New Behavior
fixed that

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5231,6 +5231,8 @@ install_freebsd_git_deps() {
     SALT_DEPENDENCIES=$(/usr/local/sbin/pkg search ${FROM_FREEBSD} -R -d sysutils/py-salt | grep -i origin | sed -e 's/^[[:space:]]*//' | tail -n +2 | awk -F\" '{print $2}' | tr '\n' ' ')
     # shellcheck disable=SC2086
     /usr/local/sbin/pkg install ${FROM_FREEBSD} -y ${SALT_DEPENDENCIES} || return 1
+    # install python meta package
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -5231,6 +5231,8 @@ install_freebsd_git_deps() {
     SALT_DEPENDENCIES=$(/usr/local/sbin/pkg search ${FROM_FREEBSD} -R -d sysutils/py-salt | grep -i origin | sed -e 's/^[[:space:]]*//' | tail -n +2 | awk -F\" '{print $2}' | tr '\n' ' ')
     # shellcheck disable=SC2086
     /usr/local/sbin/pkg install ${FROM_FREEBSD} -y ${SALT_DEPENDENCIES} || return 1
+    # install python meta package
+    /usr/local/sbin/pkg install -y lang/python || return 1
```
vutny, line 5235: @hackacad You could add `lang/python` prior to `${SALT_DEPENDENCIES}` at line 5233.
So there would be a single command execution to install everything required.

hackacad: Though of that before, but it's a dependency for setup.py (no runtime dep for py*-salt). 
Other PKGs are installed in the lines below as well.

vutny: No worries, I suggested that just as matter of little more efficiency. Nothing really changes though.

s0undt3ch (approved):

Issue #1327, opened 2019-02-28
With Amazon Linux, choosing stable always is latest stable


puluanau:
### Description of Issue/Question
When using the installer with amazon linux, specifying a stable version does nothing, as it has been hardcoded to latest there.

### Setup
Get the latest script and attempt to run in amazon linux 2.

### Steps to Reproduce Issue
```bootstrap-salt.sh stable 2017.7```
Running this on an AWS linux machine will install something newer (2019.2 when I ran).

### Versions and Systems
``` ./bootstrap-salt.sh -v
./bootstrap-salt.sh -- Version 2019.02.27```

**Comment by PaulChristophel** [comment]
I am seeing this exact issue as well when I try to bootstrap 2018.3

```
     *  INFO: Running version: 2019.01.08
     *  INFO: Executed by: 
     *  INFO: Command line: '/tmp/bootstrap-salt.sh stable 2018.3'
    
     *  INFO: System Information:
     *  INFO:   CPU:          GenuineIntel
     *  INFO:   CPU Arch:     x86_64
     *  INFO:   OS Name:      Linux
     *  INFO:   OS Version:   4.15.0
     *  INFO:   Distribution: Amazon Linux AMI 2
    
     *  INFO: Installing minion
     *  INFO: Found function install_amazon_linux_ami_2_deps
     *  INFO: Found function config_salt
     *  INFO: Found function preseed_master
     *  INFO: Found function install_amazon_linux_ami_2_stable
     *  INFO: Found function install_amazon_linux_ami_2_stable_post
     *  INFO: Found function install_amazon_linux_ami_2_restart_daemons
     *  INFO: Found function daemons_running
     *  INFO: Running install_amazon_linux_ami_2_deps()
    Loaded plugins: ovl, priorities
    Package yum-utils-1.1.31-46.amzn2.0.1.noarch already installed and latest version
    Nothing to do
    Loaded plugins: ovl, priorities
    Package python-crypto-2.6.1-2.el7.x86_64 already installed and latest version
    Package PyYAML-3.11-1.el7.x86_64 already installed and latest version
    Package python-msgpack-0.4.6-1.el7.x86_64 already installed and latest version
    Package python-requests-2.6.0-1.amzn2.noarch already installed and latest version
    Package python-zmq-15.3.0-3.el7.x86_64 already installed and latest version
[...]

PR #1328, opened 2019-02-28
Allow stable version selection for amazon linux

puluanau:
### What does this PR do?
This allows amazon linux to specify a stable release as other linuxes do.

1327

### Previous Behavior
Choosing stable with amazon linux always reverted to "latest" repo.

### New Behavior
Choosing stable and a release version will choose that version.


vutny (approved):
Excellent fix @puluanau , thank you! Resolves #1327

@s0undt3ch Test failures seems to be unrelated.

s0undt3ch (approved):
I trust on @vutny's review and it does look like the right thing to do.

s0undt3ch:
@vutny, yes, we switched to drone and it has issues that need to be solved, just haven't had time to look into it.

Issue #1331, opened 2019-03-10
Typos in usage text


wolfpackmars2:
### Description of Issue/Question
bootstrap-salt.sh -h shows ${BS_FALSE} as default for -d option

### Setup
N/A

### Steps to Reproduce Issue
sh bootstrap-salt.sh -h

### Versions and Systems
develop

s0undt3ch:
It has been a while since I wrote that, but the idea is to actually show the env variable name, not the value, to suggest how you can tweak it.

wolfpackmars2:
Closed PR - works as intended.

PR #1333, opened 2019-03-10
Fix display of defaults for bootstrap usage

wolfpackmars2:
What does this PR do?
Fixes typos in usage text

What issues does this PR fix or reference?
#1331 

Previous Behavior
Usage text:
...
-d Disables checking if Salt services are enabled to start on system boot.
You can also do this by touching /tmp/disable_salt_checks on the target
host. Default: ${BS_FALSE}
...

New Behavior
Usage text:
...
-d Disables checking if Salt services are enabled to start on system boot.
You can also do this by touching /tmp/disable_salt_checks on the target
host. Default: 0
...


wolfpackmars2:
Close. Works as intended.

PR #1334, opened 2019-03-19
install into virtualenv is broken

gregjhogan:
### What does this PR do?
fix several issues related to venv installation on ubuntu 16.04 using the following options:
```
sh bootstrap-salt.sh -D -n -V /usr/local/venv/salt -a -X git v2019.2.0
```
issues related to installation in virtualenv:
* actually create virtualenv (revert [bad commit](https://github.com/saltstack/salt-bootstrap/commit/6d595996766841766b4a0e0559bfb49231cf73ad) causing it to silently skip creating the virtualenv)
* there is no libzmq3 apt package for ubuntu 16.04
* libapt-pkg-dev apt package is required by [apt](https://pypi.org/project/apt/) pip package
* an ancient version of the [apt](https://pypi.org/project/apt/) pip package is on pypi so I installed it from source (but needs to be updated to support ubuntu 16.04 and 18.04 version)
* systemd service support

Some work is probably required to get this upstreamed (could really use a unit test), but since the functionality seems completely broken I thought I would share my work and hopefully save someone else from having to work through the parts that are broken.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2443,7 +2443,7 @@ __create_virtualenv() {
 __activate_virtualenv() {
     set +o nounset
     # Is virtualenv empty
-    if [ -z "$_VIRTUALENV_DIR" ]; then
+    if [ -z "$VIRTUAL_ENV" ]; then
         __create_virtualenv || return 1
         # shellcheck source=/dev/null
         . "${_VIRTUALENV_DIR}/bin/activate" || return 1
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2427,9 +2427,9 @@ __create_virtualenv() {
     if [ ! -d "$_VIRTUALENV_DIR" ]; then
         echoinfo "Creating virtualenv ${_VIRTUALENV_DIR}"
         if [ $_PIP_ALL -eq $BS_TRUE ]; then
-            virtualenv --no-site-packages "${_VIRTUALENV_DIR}" || return 1
+            /usr/local/bin/virtualenv --no-site-packages "${_VIRTUALENV_DIR}" || return 1
         else
-            virtualenv --system-site-packages "${_VIRTUALENV_DIR}" || return 1
+            /usr/local/bin/virtualenv --system-site-packages "${_VIRTUALENV_DIR}" || return 1
         fi
     fi
     return 0
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2888,6 +2888,11 @@ install_ubuntu_git_post() {
 
         if [ -f /bin/systemctl ] && [ "$DISTRO_MAJOR_VERSION" -ge 16 ]; then
             __copyfile "${_SALT_GIT_CHECKOUT_DIR}/pkg/salt-${fname}.service" "/lib/systemd/system/salt-${fname}.service"
+            # Set service to know about virtualenv
+            if [ "${_VIRTUALENV_DIR}" != "null" ]; then
+                mkdir -p "/etc/systemd/system/salt-${fname}.service.d" || return 1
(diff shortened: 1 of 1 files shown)


vutny (approved):
Thanks for the fix @gregjhogan .
The hard-coding of `python-apt` URL and version does not look nice, but I guess it's the only working option, right?

gregjhogan:
> The hard-coding of `python-apt` URL and version does not look nice, but I guess it's the only working option, right?

Yes, and a different branch/tag should be used for ubuntu 18.04 support, so ideally there would be logic to automatically choose between the two.  A better solution would be convincing the python apt team to push updated packages to pypi, or build and push the packages yourself.

bootstrap-salt.sh
```diff
@@ -2427,9 +2427,9 @@ __create_virtualenv() {
     if [ ! -d "$_VIRTUALENV_DIR" ]; then
         echoinfo "Creating virtualenv ${_VIRTUALENV_DIR}"
         if [ $_PIP_ALL -eq $BS_TRUE ]; then
-            virtualenv --no-site-packages "${_VIRTUALENV_DIR}" || return 1
+            /usr/local/bin/virtualenv --no-site-packages "${_VIRTUALENV_DIR}" || return 1
```
Ch3LL, line 2430: why are we hard coding the path here?

vutny: Because this is the binary installed by `pip` explicitly from the script. This eliminates possibility of running with distro-provided `virtualenv`, depending on `PATH` settings.

Ch3LL: ahhh make sense. thanks for the explanation.

Ch3LL (approved):

Ch3LL:
looks like there is spellcheck error related to `line='2895'` on your PR @gregjhogan 

Heres the test failures: https://drone.saltstack.com/saltstack/salt-bootstrap/83/1/2

bootstrap-salt.sh
gregjhogan: I found that if I run this script at the end of ubuntu installation (preseed late_command hook) via `sh` it could not resolve `virtualenv` without the full path

gregjhogan:
@Ch3LL the **shellcheck** issue is fixed, but something I don't understand still fails in your tests.  Can you add a test for installing into a virtualenv on 18.04?

Ch3LL:
@gregjhogan can you add this to your PR:

```
diff --git a/.kitchen.yml b/.kitchen.yml
index 71a43d1..ac98d43 100644
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -97,6 +97,20 @@ suites:
   - name: py2-git-2019.2
     provisioner:
       salt_version: 2019.2
+  - name: py2-venv-2019.2
+    provisioner:
+      salt_version: 2019.2
+      salt_bootstrap_options: -D -n -V /usr/local/venv/salt -a -X git %s
+    excludes:
+      - fedora
+      - centos-7
+      - centos-6
+      - amazon-1
+      - amazon-2
+      - debian-8
+      - debian-9
+      - arch
+      - opensuse
   - name: py2-git-develop
     provisioner:
       salt_version: develop
```

that should run virtualenv tests for you and going forward.

vutny:
@gregjhogan Would you be able to continue on this with the input from @Ch3LL ?
Thanks for your work!

vutny (commented):
@s0undt3ch This PR is little bit outdated, but still has working changes and no conflicts so far.
We could add more tests on this later, but merge it now. What do you think?

gregjhogan:
Sorry, I won't have time to devote any time to this in the near future. Feel free to take it over, shouldn't be much work left to do.

bryceml:
closing since there are merge conflicts and this hasn't been touched since feb.  We're happy to re-open if this get's worked on again.

PR #1335, opened 2019-03-20
Apt wait for lock modification

ripesensor:
### What does this PR do?

Apt command wait based on apt error output rather than exit code.

#1288 

### Previous Behavior
Loop based on apt exit code which may not always be due to a file lock.

### New Behavior
Loop based on apt error output.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1818,27 +1818,27 @@ __wait_for_apt(){
     WAIT_TIMEOUT=900
 
     # Run our passed in apt command
-    "${@}"
+    APT_ERR=$(mktemp /tmp/apt_error.XXXX)
+    "${@}" 2>$APT_ERR
     APT_RETURN=$?
 
-    # If our exit code from apt is 100, then we're waiting on a lock
-    while [ $APT_RETURN -eq 100 ]; do
+    # Make sure we're not waiting on a lock
+    while [ $APT_RETURN -ne 0 ] && [ $(grep -c "^E: Could not get lock" $APT_ERR) -ge 1 ]; do
       echoinfo "Aware of the lock. Patiently waiting $WAIT_TIMEOUT more seconds..."
       sleep 1
       WAIT_TIMEOUT=$((WAIT_TIMEOUT - 1))
 
-      # If timeout reaches 0, abort.
       if [ "$WAIT_TIMEOUT" -eq 0 ]; then
           echoerror "Apt, apt-get, aptitude, or dpkg process is taking too long."
           echoerror "Bootstrap script cannot proceed. Aborting."
           return 1
       else
-          # Try running apt again until our return code != 100
 	  "${@}"
     	  APT_RETURN=$?
       fi
     done
 
+    rm $APT_ERR
     return $APT_RETURN
 }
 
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -1813,27 +1819,26 @@ __function_defined() {
...
 
     # Run our passed in apt command
-    "${@}"
+    "${@}" 2>$APT_ERR
     APT_RETURN=$?
 
-    # If our exit code from apt is 100, then we're waiting on a lock
-    while [ $APT_RETURN -eq 100 ]; do
+    # Make sure we're not waiting on a lock
+    while [ $APT_RETURN -ne 0 ] && [ $(grep -c "^E: Could not get lock" $APT_ERR) -ge 1 ]; do
```
```suggestion
    while [ $APT_RETURN -ne 0 ] && grep -q -v '^E: Could not get lock' "$APT_ERR"; do
```
@ripesensor We could make it little bit more clear by removing the command substitution.
If pattern not found, this is a success.

ripesensor: Agreed.

bootstrap-salt.sh
```diff
@@ -1813,27 +1819,26 @@ __function_defined() {
...
 
     # Run our passed in apt command
-    "${@}"
+    "${@}" 2>$APT_ERR
     APT_RETURN=$?
 
-    # If our exit code from apt is 100, then we're waiting on a lock
-    while [ $APT_RETURN -eq 100 ]; do
+    # Make sure we're not waiting on a lock
+    while [ $APT_RETURN -ne 0 ] && grep -q -v '^E: Could not get lock' "$APT_ERR"; do
```
vutny: It appears that `grep` with `-v` always returns `0`. And this is really not that we expect. We need to keep trying only on non-zero code from `apt` and the lock message appeared. Please omit `-v`, I misunderstood the intended behavior.

bootstrap-salt.sh
```diff
@@ -1813,27 +1819,26 @@ __function_defined() {
...
       WAIT_TIMEOUT=$((WAIT_TIMEOUT - 1))
 
-      # If timeout reaches 0, abort.
       if [ "$WAIT_TIMEOUT" -eq 0 ]; then
           echoerror "Apt, apt-get, aptitude, or dpkg process is taking too long."
           echoerror "Bootstrap script cannot proceed. Aborting."
           return 1
       else
-          # Try running apt again until our return code != 100
 	  "${@}"
```
vutny: What missing here is `stderr` redirect to temporary log file, because for now we only check the message for the very first `apt` invocation.

vutny (changes requested):
My apologies, @ripesensor 
I suggested you a wrong improvement. Please see details below.

bootstrap-salt.sh
ripesensor: I think this is still correct? Without `-v` we'd return an exit code of `0` from `grep` if we found the string 'E: Could not get lock'. We want the inverse of that.

bootstrap-salt.sh
ripesensor: True. Fixed.

bootstrap-salt.sh
ripesensor: Fixed.

vutny (approved):
Very good, @ripesensor !

@s0undt3ch Could you please take a look?

s0undt3ch (approved):

s0undt3ch:
Shellcheck isn't happy

ripesensor:
Looks OK now.

s0undt3ch:
Thank You.

PR #1340, opened 2019-04-14
`python-xml` or `python2-xml` is required in Opensuse 15 git installations

s0undt3ch:
### What does this PR do?
Fix docker based git bootstraps of Opensuse 15


s0undt3ch:
CI runs are not even targeting Opensuse right now, merging.

PR #1344, opened 2019-04-26
apt_key_fetch: use curl

greut:
### What does this PR do?

using curl + apt-key rather than keyserver-options no-check-cert which is not always supported

unknown

### Previous Behavior

the saltstack GPG key is downloaded by apt-key

### New Behavior

the saltstack GPG key is downloaded by curl

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1867,6 +1867,26 @@ __apt_get_upgrade_noinput() {
 
 
 #---  FUNCTION  -------------------------------------------------------------------------------------------------------
+#          NAME:  __temp_gpg_pub
+#   DESCRIPTION:  Create a temporary file for downloading a GPG public key.
+#----------------------------------------------------------------------------------------------------------------------
+__temp_gpg_pub() {
+    if __check_command_exists mktemp; then
+        tempfile="$(mktemp /tmp/salt-gpg-XXXXXXXX.pub 2>/dev/null)"
+
+        if [ -z "$tempfile" ]; then
+            echoerror "Failed to create temporary file in /tmp"
+            return 1
+        fi
+    else
+        tempfile="/tmp/salt-gpg-$$.pub"
+    fi
+
+    echo $tempfile
+}   # ----------- end of function __temp_gpg_pub  -----------
+
+
+#---  FUNCTION  -------------------------------------------------------------------------------------------------------
 #          NAME:  __apt_key_fetch
 #   DESCRIPTION:  Download and import GPG public key for "apt-secure"
 #    PARAMETERS:  url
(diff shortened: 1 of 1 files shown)


vutny (changes requested):
Thanks for the PR, @greut !

Please use the `__fetch_url` function, which is a wrapper for `curl` and other tools, since cURL may not be available on some minimal distributions.
Also. no need for passing the command to `__wait_for_apt`, because `apt-key` does not set any locks.
I think it was added previously by mistake.

bootstrap-salt.sh
```diff
@@ -1875,7 +1875,7 @@ __apt_key_fetch() {
     url=$1
 
     # shellcheck disable=SC2086
-    __wait_for_apt apt-key adv ${_GPG_ARGS} --fetch-keys "$url"; return $?
+    __wait_for_apt curl ${_CURL_ARGS} -sSL "$url" | apt-key add -; return $?
```
```suggestion
    __fetch_url - "$url" | apt-key add -; return $?
```

greut:
@vutny sadly the `-` trick doesn't work and it ends up looking exactly like the `__rpm_import_gpg` function. Shall I copy it or factor out the `tempfile` part?

vutny:
@greut Yeah. Appears that stdin alias does not work when passed in to a function.

I think having the unified function for handling downloads to a temporary file/dir would be the best option. And we could reuse it in both `__rpm_import_gpg` and `__apt_key_fetch`.
Would be great if you could implement that. Thanks for your efforts!

bootstrap-salt.sh
```diff
@@ -1874,8 +1894,13 @@ __apt_get_upgrade_noinput() {
 __apt_key_fetch() {
     url=$1
 
-    # shellcheck disable=SC2086
-    __wait_for_apt apt-key adv ${_GPG_ARGS} --fetch-keys "$url"; return $?
+    tempfile = __temp_gpg_pub || return 1
```
vutny: This is not valid variable assignment syntax, no spaces allowed.

bootstrap-salt.sh
```diff
@@ -1874,8 +1894,13 @@ __apt_get_upgrade_noinput() {
 __apt_key_fetch() {
     url=$1
 
-    # shellcheck disable=SC2086
-    __wait_for_apt apt-key adv ${_GPG_ARGS} --fetch-keys "$url"; return $?
+    tempfile=__temp_gpg_pub || return 1
```
vutny: The shell is very weird "programming language" :smile: 
You can't just assign an output of a function to a variable like that. The Command Substitution must be used: `$(...)`

greut: @vutny you've got so much patience and me, not enough coffee. thanks!

vutny: No worries, @greut

bootstrap-salt.sh
```diff
@@ -1874,8 +1894,13 @@ __apt_get_upgrade_noinput() {
 __apt_key_fetch() {
     url=$1
 
-    # shellcheck disable=SC2086
-    __wait_for_apt apt-key adv ${_GPG_ARGS} --fetch-keys "$url"; return $?
+    tempfile="$(__temp_gpg_pub)" || return 1
```
vutny: One small thing to finish this. The return statement becomes superfluous, because the function will be executed in a subshell. Result will always be `0` for a variable assingment. But later lines would catch errors anyway.

vutny (approved):
Good job @greut ! Thanks.

@s0undt3ch Could you please take a look?

s0undt3ch (approved):

s0undt3ch:
Fixed builds and triggered another build on this PR. Let's see how it goes.

s0undt3ch:
Builds passed.

s0undt3ch:
Thanks!

PR #1346, opened 2019-05-15
Do not install m2crypto 0.33.0, it fails to install on CentOS a PIP installs

s0undt3ch:
### What does this PR do?
See title.


s0undt3ch:
Merged as part of https://github.com/saltstack/salt-bootstrap/pull/1347

PR #1369, opened 2019-09-18
Fix busybox mktemp compatibility

stanzgy:
Command `mktemp` accepts templates like /tmp/name.XXXXXX to create
a temporary file. Standard GNU mktemp requires at least 3 'X's
while busybox's version requires at least 6 'X's.

For compatibility we should use at least 6 'X's in mktemp templates.


vutny (approved):
Good to have `busybox` shell compatibility, thanks @stanzgy

vutny:
@stanzgy Your branch has become slightly outdated. Would you mind to rebase over latest `develop`? Thanks.

And we will kindly ask @s0undt3ch to merge this :)

s0undt3ch (approved):

PR #1376, opened 2019-10-16
fix bootstrap-salt.sh for FreeBSD

kgbsd:
FreeBSD has py36-salt packages which handle all dependencies, etc - simple as pkg install -y py36-salt and then enabling it.  Don't need dependencies or git hacks.

### What does this PR do?
able to bootstrap salt on FreeBSD

able to bootstrap salt on FreeBSD

### Previous Behavior
bootstrap FreeBSD did not work, failed with logic and dependency hell

### New Behavior
installs FreeBSD built py36-salt package with correct paths, etc.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5369,9 +5369,6 @@ install_freebsd_stable_post() {
 }
 
 install_freebsd_git_post() {
-    if [ -f $salt_conf_file ]; then
-        rm -f $salt_conf_file
-    fi
     install_freebsd_stable_post || return 1
     return 0
 }
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -239,9 +239,9 @@ _CONFIG_ONLY=$BS_FALSE
 _PIP_ALLOWED=${BS_PIP_ALLOWED:-$BS_FALSE}
 _PIP_ALL=${BS_PIP_ALL:-$BS_FALSE}
 if uname -a | grep FreeBSD > /dev/null; then
-  _SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/usr/local/etc/salt}
+    _SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/usr/local/etc/salt}
 else
-  _SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/etc/salt}
+    _SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/etc/salt}
 fi
 _SALT_CACHE_DIR=${BS_SALT_CACHE_DIR:-/var/cache/salt}
 _PKI_DIR=${_SALT_ETC_DIR}/pki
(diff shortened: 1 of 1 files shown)


kgbsd:
https://github.com/saltstack/salt-bootstrap/issues/1377

bootstrap-salt.sh
```diff
@@ -238,7 +238,11 @@ _ECHO_DEBUG=${BS_ECHO_DEBUG:-$BS_FALSE}
 _CONFIG_ONLY=$BS_FALSE
 _PIP_ALLOWED=${BS_PIP_ALLOWED:-$BS_FALSE}
 _PIP_ALL=${BS_PIP_ALL:-$BS_FALSE}
-_SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/etc/salt}
+if uname -a | grep FreeBSD > /dev/null; then
+  SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/usr/local/etc/salt}
```
bryceml: are you missing a `_` here?

bryceml:
The git installation option is not optional to implement.  It allows for installing from any arbitrary hash/branch/tag from the salt repo.  That functionality can't go away.

bootstrap-salt.sh
```diff
@@ -238,7 +238,11 @@ _ECHO_DEBUG=${BS_ECHO_DEBUG:-$BS_FALSE}
 _CONFIG_ONLY=$BS_FALSE
 _PIP_ALLOWED=${BS_PIP_ALLOWED:-$BS_FALSE}
 _PIP_ALL=${BS_PIP_ALL:-$BS_FALSE}
-_SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/etc/salt}
+if uname -a | grep FreeBSD > /dev/null; then
+  _SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/usr/local/etc/salt}
```
vutny: There are different indentation style of 2 spaces instead of 4 in all lines you have changed.

vutny (changes requested):
I guess, installing OS distro supplied package (or port) is OK when no arguments given to the bootstrap script, or the `latest`. Probably, it would be complicated to support earlier versions, but there could be a an error message about that.
If git installations are badly broken, I think it should be addressed separately.

At very least, you could just use one-liner specific to your OS to install available packages.

kgbsd:
pulled latest from git
Fixed indentation back to 4
merged my changes, ran this on FreeBSD 12, and it installed, and properly got salt minion running.
I cleaned up a lot of [really old] hacks, which partially were just wrong, and partially are not needed. This makes the FreeBSD part much cleaner, simpler and KISS compliant.
This is 100% better than having a non working script, I haven't had time to test install from git or other FreeBSD versions. I don't see why other version would fail.
With FreeBSD and pkg stuff just works, no dependencies to manually sort.

bootstrap-salt.sh
```diff
@@ -5268,113 +5272,22 @@ __freebsd_get_packagesite() {
...
 }
 
-install_freebsd_12_stable_deps() {
-    install_freebsd_9_stable_deps
+install_freebsd_deps() {
+    pkg install -y pkg
 }
 
 install_freebsd_git_deps() {
     install_freebsd_9_stable_deps || return 1
```
vutny: Looks like this function was replaced with `install_freebsd_deps()`: 
```suggestion
    install_freebsd_stable_deps || return 1
```

bootstrap-salt.sh
```diff
@@ -238,7 +238,11 @@ _ECHO_DEBUG=${BS_ECHO_DEBUG:-$BS_FALSE}
 _CONFIG_ONLY=$BS_FALSE
 _PIP_ALLOWED=${BS_PIP_ALLOWED:-$BS_FALSE}
 _PIP_ALL=${BS_PIP_ALL:-$BS_FALSE}
-_SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/etc/salt}
+if uname -a | grep FreeBSD > /dev/null; then
+    _SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/usr/local/etc/salt}
+else
+    _SALT_ETC_DIR=${BS_SALT_ETC_DIR:-/etc/salt}
+fi
```
vutny: I think overriding default values for `_SALT_ETC_DIR` and `_PKI_DIR` should be done later, let it be in `__configure_freebsd_pkg_details()` function, for the sake of keeping related code close to where it would be used.
So, no need for an `if-else` statement here, it would look much cleaner.

bootstrap-salt.sh
```diff
@@ -5268,130 +5272,31 @@ __freebsd_get_packagesite() {
...
-    return 0
-}
-
-install_freebsd_10_stable_deps() {
-    install_freebsd_9_stable_deps
-}
-
-install_freebsd_11_stable_deps() {
-    install_freebsd_9_stable_deps
+    echo 'not hijacking pkg repo configurations'
```
vutny: This functions appears to be dummy and not being referenced at all.
I suggest it would be a perfect place for overriding `_SALT_ETC_DIR` and `_PKI_DIR` variables.
Also, it should be sourced from `install_freebsd_deps()`.

bootstrap-salt.sh
```diff
@@ -5268,130 +5272,31 @@ __freebsd_get_packagesite() {
...
 
 install_freebsd_git_deps() {
-    install_freebsd_9_stable_deps || return 1
+    install_freebsd_stable_deps || return 1
 
     # shellcheck disable=SC2086
-    SALT_DEPENDENCIES=$(/usr/local/sbin/pkg search ${FROM_FREEBSD} -R -d sysutils/py-salt | grep -i origin | sed -e 's/^[[:space:]]*//' | tail -n +2 | awk -F\" '{print $2}' | tr '\n' ' ')
+    SALT_DEPENDENCIES=$(/usr/local/sbin/pkg search -R -d sysutils/py-salt | grep -i origin | sed -e 's/^[[:space:]]*//' | tail -n +2 | awk -F\" '{print $2}' | tr '\n' ' ')
+    SALT_DEPENDENCIES=$(/usr/local/sbin/pkg search -R -d py36-salt | grep 'origin:' \
+        | tail -n +2 | awk -F\" '{print $2}' | sed 's#.*/py-#py36-#g')
```
vutny: The latter line will override the value set in the previous one. Is there something missed?

vutny (changes requested):
Thanks for addressing comments and keep working on this, @kgbsd !

I just have few small inquires to make up the changes here. Looks good overall.

bootstrap-salt.sh
```diff
@@ -5268,130 +5268,30 @@ __freebsd_get_packagesite() {
...
 }
 
-install_freebsd_11_stable_deps() {
-    install_freebsd_9_stable_deps
-}
-
-install_freebsd_12_stable_deps() {
-    install_freebsd_9_stable_deps
+install_freebsd_deps() {
+    __configure_freebsd_pkg_details()
```
vutny: This is a syntax error, I guess just some copy-pasting mistake. Remove parentheses.
```suggestion
    __configure_freebsd_pkg_details
```

vutny (changes requested):
This looks much better now, thanks for your dedication @kgbsd !

We're approaching to the final cut: the function `__freebsd_get_packagesite()` has got unused.
I think you want to drop it as well.

bootstrap-salt.sh
```diff
@@ -5509,34 +5361,18 @@ install_freebsd_9_stable_post() {
...
-}
-
-install_freebsd_12_stable_post() {
-    install_freebsd_9_stable_post
-}
-
 install_freebsd_git_post() {
     if [ -f $salt_conf_file ]; then
         rm -f $salt_conf_file
     fi
```
vutny: We're close!
It seems that the `salt_conf_file` variable is not set anymore. That will cause an error. Let's drop the whole `if` statement here.

vutny:
@kgbsd We have merged a fix just a few moments ago for CI to pass.
Please rebase on latest `develop`, let's look at the build status and we will make this through.
Thanks!

vutny:
@s0undt3ch The build is pending here after the rebase, would you mind to "APPROVE" it on the Drone? Thanks!

s0undt3ch:
Done

vutny (approved):
Good job, @kgbsd !

Last CI run has passed with the exception of flaky Opensuse test, which could be ignored.
Although the branch became even with `develop` (unfortunately again), I think it is well done to be merged from the general code quality perspective and contains no regressions. Would you mind to rebase once more just for consistency?

I don't have access to the resources to provision FreeBSD machine for now, but it would be great to do a basic testing for this change. Could you please give it a shot, @Ch3LL ?

Anyway, I think this is good to go, @s0undt3ch .

vutny:
The build has stuck for some reason... But anyway, @s0undt3ch , this is good and could be merged.

vutny:
:tada:

PR #1381, opened 2019-11-01
Add fedora 31 git install support

Ch3LL:
### What does this PR do?
Add git install support for fedora 31. Stable salt packages are not yet available so currently git install is the only option.

### Previous Behavior
Could not git install salt on fedora 31

### New Behavior
Can now do a git install on fedora 31. This does require `-P` since we need to install tornado<5 and only tornado>5 packages are available on fedora 31.  Can now install with args: `./bootstrap_salt.sh -x python3 -P git 2019.2`


vutny (changes requested):
There is the issue on line [3616](https://github.com/saltstack/salt-bootstrap/blob/97d7829b00ceb511961532d7e5a2e41b09763898/bootstrap-salt.sh#L3616).
It seems that `python2-tornado` package is not present in latest Fedora. I think another check for distro version is required there.

Also, the Fedora 31 image at Drone doesn't contain `procps-ng` package, which provides `ps` binary needed for the exit traps and daemons running checks work correctly. Probably, the script should install it early, like that is done for Debian/Ubuntu.

bootstrap-salt.sh
```diff
@@ -3589,6 +3589,9 @@ install_fedora_stable_post() {
 
 install_fedora_git_deps() {
     if [ -n "$_PY_EXE" ] && [ "$_PY_MAJOR_VERSION" -eq 3 ]; then
+        if [ "$DISTRO_MAJOR_VERSION" -eq 31 ]; then
+            __install_tornado_pip "${_PY_EXE}"|| return 1
```
vutny, line 3610: This function seems to be not universal and does not support Fedora properly.

The script returns this: ` Error: Unknown repo: 'epel'`, because of the call to `yum` with `epel` enabled instead of `dnf`. I guess we need more precise logic or, even better, make `__yum_install_noinput()` function more generic to support `dnf` (without epel) as well.

Ch3LL: hmm weird  it worked when i tested locally. i'll take a look thanks

vutny:
Ubuntu 16.04 tests seem to be stuck.

s0undt3ch:
@bryceml could you please resolve the conflicts and rebase?

bryceml:
> @bryceml could you please resolve the conflicts and rebase?

I re-merged from develop resolving the conflicts.

arizvisa:
This PR is linked to saltstack/salt#54392 as the expected workaround when deploying minions with salt-cloud is to use salt-bootstrap's option to install using pip modules to force the downgrading of python3-tornado to the 5.0-series, however, to install modules using pip requires both git install support and virtual environment support.

On another note, virtual environment support is only available for ubuntu which is broken for some versions until #1334 gets merged.

(just linking issues)

bootstrap-salt.sh
s0undt3ch: Have you looked at this @Ch3LL ?

Ch3LL: i haven't had time to come back to this quit yet, i know @bryceml did some work here though

I can probably come back to this later this week/next week

arizvisa:
Thank guys!

vutny:
This one got conflicts with `develop`.

Ch3LL:
thanks for the heads up, will get to this this week

s0undt3ch:
Superseded by https://github.com/saltstack/salt-bootstrap/pull/1424

PR #1385, opened 2019-11-04
Add 2019.11.04 sha and tornado warning 

Ch3LL:
### What does this PR do?
Add sha for 2019.11.04 release and add warning about tornado <5.0 pip installs


s0undt3ch (approved):

README.rst
```diff
@@ -207,6 +208,15 @@ Installing the latest develop branch of Salt:
 
   curl -L https://bootstrap.saltstack.com | sudo sh -s -- git develop
 
+Tornado 5/6 Workaround
```
vutny: How about to move this nearby `Python 3 Support` section below?
I think it would be more logical since most distros providing latest Tornado already have switched to Py3 by default. Also, Tornado 6.0 has [dropped support for Python 2.7](https://www.tornadoweb.org/en/stable/releases/v6.0.0.html).

vutny (approved):
Looks great.

Issue #1395, opened 2019-11-27
Drone CI: which configuration file is in use and how to update it properly?


vutny:
### Description of Issue/Question
This is a question: which file serves as a source of configuration for Drone CI,
`.drone.yml` or `.drone.jsonnet`?
If some of it is legacy, we need to drop it to avoid confusion and duplication of efforts. If the one of it is somehow generated from the other one, it should be documented how to make this conversion.

Also it appears that after the changes was made to the CI configs, all subsequent builds require manual approval from authorized person to kick off the build.

@s0undt3ch, could we change the CI behavior somehow to allow config changes applied automatically once being propagated to the `develop`? Because for now the build for the tip of `develop` is stuck:
https://drone.saltstack.com/saltstack/salt-bootstrap/282

s0undt3ch:
`.drone.jsonnet` is used to generate `.drone.yml`

#1396 fixes this issue.

PR #1396, opened 2019-11-27
Fix `.drone.yml` signature

s0undt3ch:
### What does this PR do?
See title.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1182,6 +1182,6 @@ depends_on:
 
 ---
 kind: signature
-hmac: 906ceab2f0ec44b6e9c99f698feeb4874c503fd9db2a8525ed2916445069c403
+hmac: 6f9e82fb6d1fc4166a76b998fecb456cb5e931f5c2914d023008e30fba1c8822
 
 ...


s0undt3ch:
Fixes https://github.com/saltstack/salt-bootstrap/issues/1395

.drone.yml
```diff
@@ -1182,6 +1182,6 @@ depends_on:
 
 ---
 kind: signature
-hmac: 906ceab2f0ec44b6e9c99f698feeb4874c503fd9db2a8525ed2916445069c403
+hmac: 6f9e82fb6d1fc4166a76b998fecb456cb5e931f5c2914d023008e30fba1c8822
```
vutny, line 1185: I got it. The maintainer needs to sign-off the modified YAML file each time it changes.

vutny (commented):
How about `.drone.jsonnet`, is it being in use at all?

.drone.yml
s0undt3ch: Yup

vutny (approved):
Alright, so the `.drone.jsonnet` is the template source of truth for CI, and the `.drone.yml` need to be generated and signed from it for changes to be propagated.

I'm thinking about describing this procedure in README or so...

vutny:
@s0undt3ch All green, let's merge it!

Issue #1391, opened 2019-11-22
No Support of Python3 with Amazon Linux 2


johnvox:
### Description of Issue/Question
Hello, I can't managed to install salt with python3 support on AmazonLinux 2

### Setup
(Please provide relevant configs (Be sure to remove sensitive info).)
Tested on Amazon Linux AMI  amzn2-ami-hvm-2.0.20191116.0-x86_64-gp2 (ami-00068cd7555f543d5) 

### Steps to Reproduce Issue
curl -o /tmp/bootstrap-salt.sh -L https://bootstrap.saltstack.com
sh /tmp/bootstrap-salt.sh -x python3

### Versions and Systems
salt -V
Salt Version:
           Salt: 2019.2.2

Dependency Versions:
           cffi: 1.6.0
       cherrypy: Not Installed
       dateutil: 2.6.0
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.31.0
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.16 (default, Sep 13 2019, 20:28:21)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist:
         locale: UTF-8
        machine: x86_64
        release: 4.14.152-127.182.amzn2.x86_64
         system: Linux
        version: Not Installed

dmurphy18:
@Ch3LL Can confirm this to be the case

dmurphy18:
Working on this since nightly builds also needs this resolved.

dmurphy18:
Closing since fix is merged into develop

PR #1401, opened 2019-12-11
Fix Amazon Linux 2 Python 3 support and other clean up

dmurphy18:
### What does this PR do?
Provides bootstrap support for Amazon Linux 2 Python 3 support, and other minor cleanup
Work-in-progress to utilize testing 

https://github.com/saltstack/salt-bootstrap/issues/1391

### Previous Behavior
Installed Python 2 Amazon Linux 2 salt-minion, when Python 3 Amazon Linux 2 salt-minion was desired

### New Behavior
Installs Python 3 Amazon Linux 2 salt-minion via bootstrap with '-x python3' parameter.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -47,6 +47,7 @@ platforms:
       image: archlinux/base
       run_command: /usr/lib/systemd/systemd
       provision_command:
+        - pacman-key --refresh-keys
         - pacman -Syu --noconfirm --needed systemd grep awk procps which
         - systemctl enable sshd
   - name: centos-8
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -47,7 +47,9 @@ platforms:
       image: archlinux/base
       run_command: /usr/lib/systemd/systemd
       provision_command:
-        - pacman-key --refresh-keys
+        - rm -fR /etc/pacman.d/gnupg
+        - pacman-key --init
+        - pacman-key --populate archlinux
         - pacman -Syu --noconfirm --needed systemd grep awk procps which
         - systemctl enable sshd
   - name: centos-8
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -47,9 +47,6 @@ platforms:
       image: archlinux/base
       run_command: /usr/lib/systemd/systemd
       provision_command:
-        - rm -fR /etc/pacman.d/gnupg
-        - pacman-key --init
-        - pacman-key --populate archlinux
         - pacman -Syu --noconfirm --needed systemd grep awk procps which
         - systemctl enable sshd
   - name: centos-8
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -131,6 +131,7 @@ suites:
       - centos-6
       - debian-8
       - opensuse-15
+      - amazon-2
   - name: py3-git-2019-2
     provisioner:
       salt_version: 2019.2
@@ -149,6 +150,7 @@ suites:
       - centos-6
       - debian-8
       - opensuse-15
+      - amazon-2
   - name: py3-stable-2019-2
     provisioner:
       salt_version: 2019.2
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -135,7 +135,7 @@ suites:
   - name: py3-git-2019-2
     provisioner:
       salt_version: 2019.2
-      salt_bootstrap_options: -y -x python3 -MPfq git %s
+      salt_bootstrap_options: -x python3 -MPfq git %s
     excludes:
       - amazon-1
       - centos-6
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -40,6 +40,8 @@ platforms:
       image: amazonlinux:2
       platform: rhel
       run_command: /usr/lib/systemd/systemd
+      provision_command:
+        - yum install -y procps-ng
   - name: arch
     driver_config:
       image: archlinux/base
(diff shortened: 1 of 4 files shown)


bootstrap-salt.sh
```diff
@@ -2542,7 +2542,7 @@ __install_pip_pkgs() {
 #----------------------------------------------------------------------------------------------------------------------
 __install_tornado_pip() {
     # OS needs tornado <5.0 from pip
-    __check_pip_allowed "You need to allow pip based installations (-P) for Tornado <5.0 in order to install Salt on Python 3"
+    __check_pip_allowed "You need to allow pip based installations \(-P\) for Tornado <5.0 in order to install Salt on Python 3"
```
s0undt3ch: The `()` were not being shown on the output?

dmurphy18: bash -n was complaining about the un-escaped ()
So cleaned it up, since 'bash -n'

s0undt3ch: Hmm, I really don't want to see back slashes on the output...
Can you confirm the echo's don't include the `\`'s please.

dmurphy18: will remove them

dmurphy18: done

bootstrap-salt.sh
```diff
@@ -4830,15 +4830,22 @@ install_amazon_linux_ami_2_git_deps() {
         yum -y install ca-certificates || return 1
     fi
 
-    PIP_EXE='pip'
-    if __check_command_exists python2.7; then
-        if ! __check_command_exists pip2.7; then
-            __yum_install_noinput python2-pip
+    if __check_command_exists python3; then
```
vutny, line 4839: When this check is being executed, it seems to be that Python3 is not yet installed on Amazon Linux.
I think you want to call `install_amazon_linux_ami_2_deps` before this (not afterwards) to make sure everything related to Py3 is in place.

dmurphy18: I am in two minds about that, since the user would need to have '-y -x python3' to install Python 3 versions and install **Python 3** itself as it currently is, but what if the user want to install the Python 3 support but not use it yet, that is, not install **Python 3** itself just yet.

Noting that **Python 3** is not automatically installed on Amazon Linux 2 ( that is Python 3.7.4) but that could change after this month, and also Amazon might follow after Redhat 7 (which it is based on bar Python 3.6) and have Python 3 support installed, as per Redhat 7.7.

Could cut both ways as to which is better for the user.  Any insight gladly accepted.

dmurphy18: In the end followed your suggestion

dmurphy18:
re-run all

bradrydzewski:
Just a heads up that you will want to remove the `node` selector section when using Drone cloud, otherwise the pipelines will sit in a pending state until they match a runner with this label. Ideally Drone would error when we know a certain configuration will never match a runner. To address this I created https://github.com/drone/drone/issues/2894.

.kitchen.yml
```diff
@@ -40,11 +40,16 @@ platforms:
...
   - name: arch
     driver_config:
       image: archlinux/base
       run_command: /usr/lib/systemd/systemd
       provision_command:
+        - rm -fR /etc/pacman.d/gnupg
+        - pacman-key --init
+        - pacman-key --populate archlinux
         - pacman -Syu --noconfirm --needed systemd grep awk procps which
         - systemctl enable sshd
```
vutny: There is an issue with generating SSH host keys for Arch in earlier stage.
The command appears to be `ssh-keygen -A -t rsa -f /etc/ssh/ssh_host_rsa_key`.
But, the man page says:
```
If -f has also been specified (with -A), its argument is used as a prefix to the
default path for the resulting host key files.
```
The `-f` parameter is for special use in that case, I don't think we need to specify it.

dmurphy18: Trying to get why ssh fails to connect. Thought it was the key errors, hence  attempted to insure right keys, but the issue with key errors occurs before any provision_command is run.  Also examining @Ch3LL successful run on her PR, she sees the same issue with key errors but runs successfully, hence removed those commands.  Retrying with my VPN connected to SaltStack but would not expect that to be needed.  The lack of golden images to test against, rather than just latest, bothers me, since the OS can change from one moment to the next: but then it is Arch and that would be a losing battle to keep track of.

bootstrap-salt.sh
```diff
@@ -23,7 +23,7 @@
 #======================================================================================================================
 set -o nounset                              # Treat unset variables as an error
 
-__ScriptVersion="2019.11.04"
+__ScriptVersion="2019.12.19"
```
vutny: Is it supposed to be changes on releases, isn't it?

vutny:
Looks good, the CI test seems to be passed.

waynew (approved):

bootstrap-salt.sh
s0undt3ch: Agreed, please don't _bump_ the version.

s0undt3ch (changes requested):
Please remove the version _bump_.

Akm0d (approved):

bootstrap-salt.sh
dmurphy18: Whoops, reverted back to original value, got carried away with update, similar to AIX / Solaris build scripts which are updated whenever touched.

s0undt3ch (approved):

vutny (approved):
LGTM

PR #1413, opened 2020-01-22
Use package "origin" for all FreeBSD packages

cedwards:
### What does this PR do?
This reverts an earlier change hard-coding "py36-salt". FreeBSD has moved to py37.

The issue I'm reporting right now with this patch. Referring to packages by specific version will cause issues as the Python versions increase. Referring to packages by their port name (origin) will avoid any future version-specific issues and automatically default to installing the latest.

### Previous Behavior
bootstrap doesn't work on FreeBSD

### New Behavior
bootstrap works on FreeBSD

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -328,7 +328,7 @@ UNIX systems
 **BSD**:
 
 - OpenBSD (``pip`` installation)
-- FreeBSD 9/10/11
+- FreeBSD 9/10/11/12
 
 **SunOS**:
 
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5296,7 +5296,7 @@ install_freebsd_deps() {
 }
 
 install_freebsd_git_deps() {
-    install_freebsd_stable_deps || return 1
+    install_freebsd_deps || return 1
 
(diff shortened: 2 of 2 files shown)


cedwards:
This patch works but has the unintended side-effect of installing two sets of python dependencies.

Researching solutions... not quite ready to merge.

cedwards:
This should be ready now.

Using the "port origin" as the install target has the unexpected side-effect of installing any variant built from the port (ie; py27 and py37, etc).

s0undt3ch (approved):

README.rst
```diff
@@ -329,7 +329,7 @@ UNIX systems
 **BSD**:
 
 - OpenBSD (``pip`` installation)
-- FreeBSD 9/10/11
+- FreeBSD 9/10/11/12
```
vutny, line 332: I guess the releases 9 and 10 would not be supported for real...
Need to adjust deprecation warnings in the script, but that is for another PR.

s0undt3ch:
@cedwards, can you confirm what @vutny mentioned?

s0undt3ch:
And, as always @vutny, thank you for the review.

cedwards:
9 and 10 are now EOL so we don't need to support them. It's a matter for another PR.

s0undt3ch:
Thank You @cedwards!

PR #1421, opened 2020-01-29
Final changes for stable release

s0undt3ch:
### What does this PR do?
* Remove duplicate entry in pre-commit config
* Update ChangeLog for release
* Bump version for release
* FreeBSD versions lower than 11 are EOL and unsupported


s0undt3ch:
The GH Actions failure should be ignored

Issue #1425, opened 2020-02-04
salt-bootstrap does not work on Ubuntu 20.04


vernondcole:
### Description of Issue/Question

The anticipated LTS of Ubuntu is not yet supported.

### Setup

- Download a pre-release of Ubuntu 20.04 from http://cdimage.ubuntu.com/daily-live/current/

- install it

### Steps to Reproduce Issue

- download bootstrap_salt.sh as usual

- `sudo sh bootstrap-salt.sh -D -A 192.168.88.2 -x python3 -P git`

As expected, the bootstrap fails.

### Versions and Systems
(after the supplied patch below)
```
vernon@hplt:~/salt-bootstrap$ salt --versions-report
/usr/local/lib/python3.7/dist-packages/salt/ext/tornado/httputil.py:107: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
  class HTTPHeaders(collections.MutableMapping):
Salt Version:
           Salt: 3000
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.7.3
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
         Jinja2: 2.11.1
        libgit2: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.6.2
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.7.6 (default, Jan 16 2020, 18:23:42)
   python-gnupg: Not Installed
         PyYAML: 5.3
          PyZMQ: 18.1.1
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.3.2
 
System Versions:
           dist: Ubuntu 20.04 focal
         locale: UTF-8
        machine: x86_64
        release: 5.4.0-12-generic
         system: Linux
        version: Ubuntu 20.04 focal
```
```
sh bootstrap-salt.sh -v
bootstrap-salt.sh -- Version 2020.01.29
```
fix by v v v
[ubuntu_20_04.patch.txt](https://github.com/saltstack/salt-bootstrap/files/4150825/ubuntu_20_04.patch.txt)

PR #1424, opened 2020-02-03
Final Neon Support Fix

s0undt3ch:
### What does this PR do?
See title.
Additionally:
* Drop support for Fedora < 30
* Drop support for Debian < 8
* Adds preliminary support for Fedora 32, although disabled since it defaults to Py3.8 which salt doesn't yet support
* Locks kitchen config for Arch Linux to python 2 since Arch is also on Py3.8

### What issues does this PR fix or reference?

Closes https://github.com/saltstack/salt-bootstrap/issues/1395
Closes https://github.com/saltstack/salt-bootstrap/issues/1400
Closes https://github.com/saltstack/salt-bootstrap/issues/1423
Closes https://github.com/saltstack/salt-bootstrap/pull/1381

PR #1426, opened 2020-02-04
first changes to support Ubuntu 20.04

vernondcole:
### What does this PR do?

Loads a git version of salt minion onto a pre-release Ubuntu 20.04 machine

#1425

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1,3 +1,4 @@
+
 #!/bin/sh -
 
 # WARNING: Changes to this file in the salt repo will be overwritten!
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -1,3 +1,4 @@
+
```
vutny: I guess the blank line here was added occasionally, right?

vernondcole: Yes, I suppose so. At least, I did not intend to add it.

```suggestion
```
Better to remove it, cause it breaks lint checker.

vutny (commented):
The PR looks good, @vernondcole . But unfortunately, after merging #1424 it has conflicts now.
Would you mind to update your changes in line with current `develop`?

vernondcole:
I committed these and pushed them straight from my 20.04 test machine, which does not have gpg keys registered.  
@vutny Do I need to re-submit from a registered workstation?

s0undt3ch (approved):

s0undt3ch:
It's desirable but I can override the key issue.

Issue #1418, opened 2020-01-24
The bootstrap script should install the python3 version of salt without needing to provide -x python3 on oses that don't support python2


bryceml:
### Description of Issue/Question
Debian 10, Rhel 8, and Fedora 32 we no longer support python2 on.  We should not require `-x python3` in order for it to correctly install a salt on those oses by default.  For oses that don't support python2, it should automatically install the python3 version.

darkn3rd:
The default should be python3.  Python2 is offiicially deprecated in 2020.  

If users need compatibility to `python2`, then they could specify `-x python2`.

bryceml:
Debian 10, Rhel 8, and Fedora 32 all default to py3 now.

Issue #1350, opened 2019-05-29
Ubuntu regression: ERROR: Failed to run install_ubuntu_stable_deps()!!!


ipmb:
### Description of Issue/Question

Bootstrap is now failing on Ubuntu (tested with both Xenial and Bionic).

### Setup

Grab a copy of the `2019.05.20` release.

### Steps to Reproduce Issue

I'm seeing this in a more complex LXC environment, but this is the easiest way to reproduce:

```
docker run --rm -v `pwd`/bootstrap-salt.sh:/sbin/bootstrap.sh ubuntu:bionic sh /sbin/bootstrap.sh -D stable
```

*Debug output*

```
 *  INFO: Running version: 2019.05.20
 *  INFO: Executed by:
 *  INFO: Command line: '/sbin/bootstrap.sh -D stable'

 *  INFO: System Information:
 *  INFO:   CPU:          GenuineIntel
 *  INFO:   CPU Arch:     x86_64
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   4.9.125-linuxkit
 *  INFO:   Distribution: Ubuntu 18.04

 * DEBUG: Binaries will be searched using the following $PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 *  INFO: Installing minion
 * DEBUG: install_ubuntu_18_stable_deps not found....
 * DEBUG: install_ubuntu_18_04_stable_deps not found....
 * DEBUG: install_ubuntu_18_deps not found....
 * DEBUG: install_ubuntu_18_04_deps not found....
 *  INFO: Found function install_ubuntu_stable_deps
 * DEBUG: DEPS_INSTALL_FUNC=install_ubuntu_stable_deps
 * DEBUG: config_ubuntu_18_stable_salt not found....
 * DEBUG: config_ubuntu_18_04_stable_salt not found....
 * DEBUG: config_ubuntu_18_salt not found....
 * DEBUG: config_ubuntu_18_04_salt not found....
 * DEBUG: config_ubuntu_stable_salt not found....
 * DEBUG: config_ubuntu_salt not found....
 *  INFO: Found function config_salt
 * DEBUG: CONFIG_SALT_FUNC=config_salt
 * DEBUG: preseed_ubuntu_18_stable_master not found....
 * DEBUG: preseed_ubuntu_18_04_stable_master not found....
 * DEBUG: preseed_ubuntu_18_master not found....
[...]

PR #1442, opened 2020-03-27
Set python version when unset

### LINKED ISSUE #1388: Bootstrap doesn't work for Debian 10 (Buster)
Debian 10 Salt packages are only available for Python 3.
But, the Debian 10 default is Python 2.
Bootstrap Python version detection find 2, and try to install packages for this version.
  
### Setup
```
$ docker run --rm -it debian:buster-slim bash
root@b153991e056a:/# apt-get update
root@b153991e056a:/# apt-get -y install curl
root@b153991e056a:/# curl -L https://bootstrap.saltstack.com -o install_salt.sh
```

### Steps to Reproduce Issue
```
root@b153991e056a:/# sh install_salt.sh -P

 *  INFO: Running version: 2019.11.04
 *  INFO: Executed by: 
 *  INFO: Command line: 'install_salt.sh -P'

 *  INFO: System Information:
 *  INFO:   OS Version:   4.9.0-11-amd64
 *  INFO:   Distribution: Debian 10.1

 *  INFO: Installing minion
 *  INFO: Found function install_debian_deps
 *  INFO: Found function install_debian_stable
 *  INFO: Found function install_debian_restart_daemons
 *  INFO: Found function install_debian_check_services
 *  INFO: Running install_debian_deps()
[...]
The following NEW packages will be installed:
  ..... python python-minimal python-yaml
  python2 python2-minimal python2.7 python2.7-minimal .....
[...]
The following NEW packages will be installed:
  apt-transport-https .....
[...]
Warning: apt-key output should not be parsed (stdout is not a terminal)
gpg: no valid OpenPGP data found.
 * ERROR: Failed to run install_debian_deps()!!!
```
Python 2.7 and wrong repository:
```
root@b153991e056a:/# cat /etc/apt/sources.list.d/saltstack.list 
deb https://repo.saltstack.com/apt/debian/10/amd64/latest buster main
```
It is not a major issue, but `apt-transport-https` is now useless for Debian 10.
apt include it.

### Idea
Since packages are only available for Python 3. I think the Python version detection is not necessary anymore  for Debian 10 and we could just set version to 3.

And we also can remove `apt-transport-https` from dependencies.

Do you agree with that ?

wolfpackmars2:
Sets python version to 3 if version to install is >= 2019.2 and python3 is available.
fixes #1388 #1418 #1350 
Also sets default to python3 version of Salt
Tested on Debian 10

### What does this PR do?
Sets python3 version of Salt to be installed if not specified on command line and Python3 is available.

#1388 #1418 #1350 and possibly others

### Previous Behavior
Remove this section if not relevant
Couldn't install on Debian 10 without adding -x option
Ubuntu 18.04 would install Python2 Salt by default.

### New Behavior
Defaults to Python3 Salt when certain conditions are met.


bootstrap-salt.sh
```diff
@@ -785,6 +785,20 @@ if [ -d "${_VIRTUALENV_DIR}" ]; then
     exit 1
 fi
 
+# Set the python exe if unset and version is >= 2019.2
+if [ "$_PY_EXE" = "" ] && [ "$(echo "$STABLE_REV" | grep -E '^(latest|2019\.2|3000)$')" != "" ]; then
+    which python3 > /dev/null
```
s0undt3ch: Use `command` instead of `which`. See source code for examples.

s0undt3ch (changes requested):

bootstrap-salt.sh
vutny: There is a special function for that: `__check_command_exists`

s0undt3ch: Even better :smile:

bryceml:
Ideally this would be able to automatically handle any new release of salt being released, i.e. 3002, 3003, etc.  It also becomes much less complicated if we only did this on 3001 and forward as we wouldn't change any working previous behavior.  If we did it on 3001 and newer we also wouldn't have to make special cases for centos6, amazon1 and debian8 only having py2 available.

bryceml:
superceded by https://github.com/saltstack/salt-bootstrap/pull/1577

PR #1443, opened 2020-04-02
Salt v3000.1 was released

s0undt3ch:
### What does this PR do?
See title


vutny (approved):
:+1:

sblaisot:
This line is only supposed to contain major version (2019.2 or 3000) and not each dot release.

IMHO this is unneeded

either you choose 3000 and will get any new dot release in the 3000 series or you choose 3000.1 and should get exact dot release from archive.

s0undt3ch:
You stand correct @sblaisot

Issue #1458, opened 2020-05-05
bootstrap script fails on KDE neon (Ubuntu 18.04 base)


ender8282:
### Description of Issue/Question

Attempting to run the bootstrap script fails on KDE Neon. The two relevant error lines from the outputs are:

salt.sh: 1: eval: neon_18_ubuntu_base: parameter not set
 * ERROR: No dependencies installation function found. Exiting...

### Steps to Reproduce Issue
Boot KDE Neon live CD
Download salt bootstrap
$ sudo sh bootstrap-salt.sh

### Versions and Systems
bootstrap 2020.02.24
KDE Neon live CD 20200430-1117

To be clear the script properly identifies the distribution.  It is just missing a line that associates neon with ubuntu 18.04.  

PR to follow.

s0undt3ch:
The required changes have been merged.
Please reopen or open a new issue of needed.

PR #1459, opened 2020-05-05
Support for newer KDE Neon

ender8282:
### What does this PR do?
Causes salt to properly support Ubuntu 18.04 based versions

https://github.com/saltstack/salt-bootstrap/issues/1458

### Previous Behavior
Salt would fail to install on KDE Neon based on Ubuntu 18.04

### New Behavior
Salt properly detects KDE Neon based on Ubuntu 18.04

This PR includes both the fix for neon based on 18.04 as well as what I assume will be the change necessary for neon based on 20.04.  I realize that salt doesn't support Ubuntu 20.04.  KDE Neon based on 20.04 isn't actually out yet but since I was changing things I figured I'd try to future proof things.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -301,7 +301,7 @@ in combination with the ``git`` installation method.
 Ubuntu and derivatives
 ~~~~~~~~~~~~~~~~~~~~~~
 
-- KDE neon (based on Ubuntu 16.04)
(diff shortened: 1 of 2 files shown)


s0undt3ch:
This has been merged on a separate PR

PR #1466, opened 2020-05-22
Add support for the upcoming Sodium release

s0undt3ch:
### What does this PR do?
Adds support for the upcoming Sodium release


myii:
Nice, I'll start using this to build pre-salted images once this is merged.  A question though: shouldn't the test matrix be updated to include `3001` as well?

s0undt3ch:
@myii Sodium is not released yet.
You might want to wait for that.
:)

myii:
@s0undt3ch I understand, we're already building images for `master` anyway, so no problem having some `3001` images around for playing around with!  At least we'll already be prepared for the Sodium release once it's in place.

s0undt3ch:
@myii For sure, just wanted to clear that out

PR #1467, opened 2020-05-23
Reference `master` instead of `develop` branch of Salt

eliasp:
### What does this PR do?
Since the default branch of saltstack/salt was changed from `develop` to
`master`, the documentation of saltstack/salt-bootstrap should follow
this change as well, otherwise users might end up dealing with outdated
installations while expecting to have bleeding-edge code running.

None


s0undt3ch:
@eliasp Thank You for your PR.
A change was merged which addressed this issue.

PR #1480, opened 2020-06-17
Default Ubuntu 20.04 to Py3

bryceml:
### What does this PR do?
Defaults ubuntu 20.04 and later to py3 for stable installs


bryceml:
conflicts logically with #1442

s0undt3ch (approved):

Issue #1436, opened 2020-02-24
Feature Request: Unable to pin minor version 3000


sblaisot:
### Description of Issue/Question

Before 3000, one could use `salt-bootstrap stable 2019.2` to install with pin-to-major repo `2019.2` and receive any patchlevel update in the 2019.2 series or use `salt-bootstrap stable 2019.2.0` to install with pin-to-minor repo `archives/2019.2.0` and stay forever on 2019.2.0 without getting any update.

This pin-to-major/pin-to-minor was implicit from different version numbers.

Now, with salt 3000+, pin-to-major and pin-to-minor repo have the same name `3000` (either plain `3000` or `archives/3000` repos) and you have no way to select pin-to-minor `archives/3000` repo using salt-bootstrap.

I think i would be better to make this selection explicit with a new parameter asking for pin-to-minor repo rather than relying on version number liek it previously was.

So I'm asking for a way to install salt 3000 with salt-bootstrap from `archives/3000` repo using a new salt-bootstrap parameter.

ghost:
The script checks for hardcoded versions here: https://github.com/saltstack/salt-bootstrap/blob/9f2f1b56fbc0d0264ed75e97cda4e39e9529792e/bootstrap-salt.sh#L606

Maybe it just need to be added that the newest version is 3000.2.

max-arnold:
5 months passed, 3001.1 is out, and again any scripted 3001 install became non-reproducible. Here is a fix: https://github.com/saltstack/salt-bootstrap/pull/1491

max-arnold:
The fix was merged, so I guess this issue can be closed

s0undt3ch:
Yep, this is fixed.

PR #1491, opened 2020-08-09
Allow pinning minor 3xxx versions (3000.0, 3001.0, etc)

max-arnold:
### What does this PR do?

Allow pinning minor 3xxx versions. This is enabled by specifying a fake `.0` version suffix:

`salt-bootstrap.sh -x python3 stable 3001.0`
  
https://github.com/saltstack/salt-bootstrap/issues/1436

Previously, it was impossible to have reproducible bootstrap process - a major 3xxx version install could mean 3xxx, 3xxx.1, 3xxx.2, etc. depending on the time of the install.

Also addressed the following comment: https://github.com/saltstack/salt-bootstrap/commit/a321d51c385db225ddc08e42852fc59cba7b1992#r37765195

Note to reviewers: not sure if I added the test properly. Is `.kitchen.yml` the right place?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -412,6 +412,52 @@ jobs:
           bundle exec kitchen destroy py3-git-3001-amazon-2
 
 
+  py3-stable-3001-0-amazon-2:
+    name: Amazon 2 v3001.0 Py3 Stable
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    needs: lint
+
+    steps:
+      - uses: actions/checkout@v1
+      - name: Setup Ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.6.x
+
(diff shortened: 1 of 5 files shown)


s0undt3ch (approved):

s0undt3ch:
Before merging we need to update the GH Actions to include minor pinning into the PR tests.

The logic is in https://github.com/saltstack/salt-bootstrap/blob/develop/.github/workflows/templates/generate.py but the logic looks complex. I'll try to take a look on Monday.

bryceml:
https://github.com/max-arnold/salt-bootstrap/compare/pin-to-minor-post-neon...bryceml:test_0_version should add the tests if @s0undt3ch agrees with how that's done.

s0undt3ch:
Yep, that's great Bryce!

bryceml (approved):

s0undt3ch (approved):

Issue #1497, opened 2020-09-10
[Feature Request] Add option to only configure repository


blarghmatey:
### Description of Issue/Question
It would be nice to be able to use the bootstrap script to set up the appropriate package repository for the given distribution without having to install the salt-minion or other packages. The use case here being that I would like to use the bootstrap script to populate the package repo using the `runcmd` stage of cloud-init and then use the `salt-minion` module subsequently to handle installing the minion package and populate the config and minion keys. When using the bootstrap script to install the minion it automatically generates keys which then fail to be accepted, vs the already accepted keys that I am passing to the salt-minion cloud-init module.

dmurphy18:
@blarghmatey What you are wanting is available from the Install Instructions, https://docs.saltproject.io/salt/install-guide/en/latest/, take a look at https://docs.saltproject.io/salt/install-guide/en/latest/topics/install-by-operating-system/rhel.html#install-salt-on-redhat-rhel-9-x86-64 for example installing for RedHat 9, and how to setup repository information to allow you to access the Salt repositories.

Closing this feature request since a method is already available to achieve the same outcome.  Please reopen if there are additional details as to the Feature Request you are requesting.

PR #1498, opened 2020-09-10
Added flag to only set up salt package repo and dependencies

blarghmatey:
### What does this PR do?
Adds a flag to only set up the salt package repository and install dependencies in order to allow the bootstrap script to coexist with the cloud-init salt-minion module.

#1497 

### New Behavior
Adds the ability to set up the Salt package repository and install dependencies without having to install the salt minion or override the configuration.

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -275,6 +275,7 @@ _POST_NEON_INSTALL=$BS_FALSE
 _MINIMUM_PIP_VERSION="9.0.1"
 _MINIMUM_SETUPTOOLS_VERSION="9.1"
 _POST_NEON_PIP_INSTALL_ARGS="--prefix=/usr"
+_INSTALL_DEPS_ONLY=$BS_FALSE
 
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -7929,7 +7935,7 @@ fi
 
 
 # Install dependencies
-if [ ${_NO_DEPS} -eq $BS_FALSE ] && [ $_CONFIG_ONLY -eq $BS_FALSE ]; then
+if [ ${_NO_DEPS} -eq $BS_FALSE ] && [ $_CONFIG_ONLY -eq $BS_FALSE ] && [ $_INSTALL_DEPS_ONLY -eq $BS_TRUE ]; then
```
bryceml, line 7975: shouldn't this be `||` not `&&` ? or just not be here at all?

s0undt3ch:
The purpose of the script is to bootstrap Salt, not prepare the system to bootstrap.
Let's not try to complicate the script further.

Issue #1513, opened 2020-12-04
Best-effort support for Debian 11 testing


jpacura:
### Description of Issue/Question
The current salt-bootstrap script will not install on Debian 11 because it will read the version as Debian 1. 

### Steps to Reproduce Issue
```
 *  INFO: Running version: 2020.10.20
 *  INFO: Executed by: sh
 *  INFO: Command line: 'bootstrap-salt.sh -D git master'

 *  INFO: System Information:
 *  INFO:   CPU:          GenuineIntel
 *  INFO:   CPU Arch:     x86_64
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   5.9.0-4-amd64
 *  INFO:   Distribution: Debian 1

 * ERROR: End of life distributions are not supported.
 * ERROR: Please consider upgrading to the next stable. See:
 * ERROR:     https://wiki.debian.org/DebianReleases
 * DEBUG: Removing the logging pipe /tmp/bootstrap-salt.logpipe
 * DEBUG: Removing the temporary apt error file /tmp/apt_error.5vZ2BL
 * DEBUG: Killing logging pipe tee's with pid(s): 9003
```

### Versions and Systems
Clean Debian 11 install using netinstall ISO
```
$ cat /etc/debian_version
bullseye/sid
$ sudo sh bootstrap-salt.sh -v
bootstrap-salt.sh -- Version 2020.10.20
```

bryceml:
Closing since support was merged in.

PR #1113, opened 2017-07-13
Introductory support for Debian 10 (Buster)

rallytime:
This support relies on installing packages for Debian 9, similarly to what the non-LTS Ubuntu versions by installing LTS Ubuntu packages.

PR #1514, opened 2020-12-04
Debian 11 (bullseye/testing) support using Debian 10 packages

jpacura:
### What does this PR do?
Adds support for Debian 11 (testing) using Debian 10 repository packages

https://github.com/saltstack/salt-bootstrap/issues/1513

### Previous Behavior
Script will fail to run because it reads the version as 'Debian 1', and will consider it end-of-life

### New Behavior
Install as if the system is Debian 10

Thanks to @rallytime for their Debian 10 introductory PR, which was used as a base for this PR. 
https://github.com/saltstack/salt-bootstrap/pull/1113

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2620,6 +2620,190 @@ jobs:
           bundle exec kitchen destroy latest-debian-10
 
 
+  py3-stable-3002-debian-11:
+    name: Debian 11 v3002 Py3 Stable
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    needs: lint
+
+    steps:
+      - uses: actions/checkout@v1
+      - name: Setup Ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.6.x
+
+      - name: Install Bundler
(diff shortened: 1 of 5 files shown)


s0undt3ch:
Are there any Debian 11 docker containers?

jpacura:
debian:bullseye will always point to the latest Debian 11 container. debian:testing will point to Debian 11 until it is released as stable, then it will point to Debian 12. Right now, these are the same since Debian 11 is testing. 

I have confirmed that the latest develop commit fails to run on the debian:testing container, while this PR will succeed, as it does with a bare-metal Debian 11 install.

s0undt3ch:
We should then add `debian:bullseye` to the platforms to run tests on, on this PR too.

s0undt3ch:
https://github.com/saltstack/salt-bootstrap/blob/develop/.github/workflows/templates/generate.py#L15 let's add `debian-11` assuming there's also a `11` tag.

jpacura:
There isn't an 11 tag. Not sure whether I should use testing or bullseye

s0undt3ch (approved):

PR #1516, opened 2020-12-12
Fix bootstrap on Gentoo.

ijansky:
### What does this PR do?
This PR fixes builds on Gentoo that have been failing recently. It addresses several issues that have accumulated:
- Gentoo Docker are not rebuilt nightly due to https://github.com/gentoo/gentoo-docker-images/issues/97, which leads to longer build times.
- Gentoo switched to Python 3.8 as the default python3 version.

### What issues does this PR fix or reference.
N/A

### Previous Behavior
- Salt 2019 GIT builds used latest msgpack even though they should be using msgpack <1.0.
- Salt 3000 GIT builds did not work with Python 3.8.
- Build time capped at 50 minutes.

### New Behavior
- Salt 2019 GIT builds use msgpack <1.0.
- Salt 3000 GIT builds use Python 3.7.
- Tests for x.0 versions on Gentoo are removed as Gentoo does not keep x.0 ebuilds.
- Running a full system update when bootstraping test containers is followed by depclean and setting default Python version to 3.8. _The full system update, depclean, and setting the Python version should be removed once Gentoo has fresh images on Docker Hub._
- The build timeout increases from 50 to 90 minutes. _This is a temporary necessity, which should be reverted once Gentoo has fresh builds on Docker Hub._

### DIFF
--- a/.github/workflows/main-branch-only.yml
+++ b/.github/workflows/main-branch-only.yml
@@ -11,7 +11,7 @@ jobs:
   py3-stable-2019-2-gentoo:
     name: Gentoo v2019.2 Py3 Stable
     runs-on: ubuntu-latest
-    timeout-minutes: 70
+    timeout-minutes: 90
 
 
 
@@ -57,7 +57,7 @@ jobs:
   py3-git-2019-2-gentoo:
     name: Gentoo v2019.2 Py3 Git
     runs-on: ubuntu-latest
-    timeout-minutes: 70
+    timeout-minutes: 90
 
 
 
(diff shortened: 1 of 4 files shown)


s0undt3ch (approved):

ijansky:
@s0undt3ch, Gentoo builds fail again and the root cause seems to be that compiling one of the packages during world upgrade fails. We can get around this by temporarily switching away from gentoo/stage3 to ksmanis/stage3 - see https://github.com/gentoo/gentoo-docker-images/pull/100. This would also significantly cut down the execution time on Gentoo. If that is ok, then I will prepare a PR.

s0undt3ch:
I guess that would be ok, yes

PR #1518, opened 2020-12-19
Gentoo - use ksmanis/stage3 image

ijansky:
Use an alternative image until gentoo/stage3 gets pushed to Docker Hub again.

### What does this PR do?
It uses the [ksmanis/stage3](https://hub.docker.com/r/ksmanis/stage3) image instead of the outdated [gentoo/stage3](https://hub.docker.com/r/gentoo/stage3). The use of the alternative image should be reverted back to gentoo/stage3 once updated images get deployed again - see https://github.com/gentoo/gentoo-docker-images/pull/100

It also sets python3 according to the PYTHON_SINGLE_TARGET portage variable.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -82,22 +82,16 @@ platforms:
         - sed -i 's/^PubkeyAcceptedKeyTypes.*$/&,ssh-rsa/' /etc/crypto-policies/back-ends/opensshserver.config
   - name: gentoo
     driver_config:
-      image: gentoo/stage3:latest
+      image: ksmanis/stage3:latest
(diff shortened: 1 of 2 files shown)


myii:
Just to add, @ijansky provided this update to our `salt-image-builder` and has got all of our Gentoo `stable` builds working again:

* https://gitlab.com/saltstack-formulas/infrastructure/salt-image-builder/-/merge_requests/73

With this change, we're expecting `git`-based builds from the `master` branch to work as well.

s0undt3ch (approved):

PR #1520, opened 2021-01-04
Handle Vagrant's `grains_config` option (same as `bootstrap-salt.sh`)

dafyddj:
### What does this PR do?

Handles Vagrant's `grains_config` option in the same way as `bootstrap-salt.sh`.
Vagrant will use this bootstrap script as part of its Salt provisioner (see hashicorp/vagrant#12127).

The same operation is performed here in `bootstrap-salt.sh`:
https://github.com/saltstack/salt-bootstrap/blob/da0f0733edf121d9b5e1bd5a8219309c38e975d9/bootstrap-salt.sh#L7600-L7605

Another commit makes the copy command used to copy the minion keys consistent with the other copy commands.

### Previous Behavior
None

### New Behavior
Copies the file `grains` to `C:\salt\conf`


s0undt3ch:
@twangboy are we good here?

twangboy (approved):

PR #1525, opened 2021-01-20
fix(opensuse): fix & add CI for openSUSE Tumbleweed

myii:
### What does this PR do?

When `systemd` is installed on Tumbleweed, `systemctl` is actually
located at `/usr/bin/systemctl`.  It appears that the same applies for
openSUSE Leap except that a symlink is also provided there, i.e.
`/bin/systemctl` => `/usr/bin/systemctl`.

Tumbleweed installs Python 3.8, which was first supported in `3001` (Sodium).
Hence, testing on `2019.2` and `3000` is not possible.

When provisioning in Kitchen, another package is supplied compared to
openSUSE Leap:

* `systemd`: It appears that this is implicitly installed in Leap as a
  dependency of the `openssh` package; this doesn't apply in Tumbleweed,
  so it is installed explicitly.

### Previous Behavior

Installation on Tumbleweed failed.

```shell
 *  INFO: Running install_opensuse_git_post()
cp: cannot create regular file '/etc/init.d/salt-minion': No such file or directory
chmod: cannot access '/etc/init.d/salt-minion': No such file or directory
Failed to get unit file state for salt-minion.service: No such file or directory
Failed to preset unit, unit salt-minion.service does not exist.
```

### ~CI~

~Since the CI seems to be failing here, this is a link to the subset of GitHub
Actions that I tested this with:~

* ~https://github.com/myii/salt-bootstrap/actions/runs/499468242~

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -6463,6 +6463,9 @@ install_opensuse_git_deps() {
         if [ "$_INSTALL_CLOUD" -eq $BS_TRUE ]; then
             __PACKAGES="${__PACKAGES} python-apache-libcloud"
         fi
+    # Check for Tumbleweed
+    elif [ "${DISTRO_MAJOR_VERSION}" -ge 20210101 ]; then
+        __PACKAGES="python3-pip"
     else
         __PACKAGES="python-pip python-setuptools gcc"
     fi
@@ -6544,7 +6547,7 @@ install_opensuse_git_post() {
         [ $fname = "minion" ] && [ "$_INSTALL_MINION" -eq $BS_FALSE ] && continue
         [ $fname = "syndic" ] && [ "$_INSTALL_SYNDIC" -eq $BS_FALSE ] && continue
 
-        if [ -f /bin/systemctl ]; then
+        if command -v systemctl; then
             use_usr_lib=$BS_FALSE
 
             if [ "${DISTRO_MAJOR_VERSION}" -ge 15 ]; then
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3264,6 +3264,190 @@ jobs:
           bundle exec kitchen destroy latest-opensuse-15
 
 
+  py3-git-3001-opensuse-tumbleweed:
+    name: Opensuse Tumbleweed v3001 Py3 Git
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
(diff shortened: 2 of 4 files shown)


bryceml:
The main.yml is generated with .github/workflows/templates/generate.py

please update that file and make sure that the committed main.py is what is generated by the generate.py script.

bootstrap-salt.sh
```diff
@@ -6538,7 +6541,7 @@ install_opensuse_git_post() {
         [ $fname = "minion" ] && [ "$_INSTALL_MINION" -eq $BS_FALSE ] && continue
         [ $fname = "syndic" ] && [ "$_INSTALL_SYNDIC" -eq $BS_FALSE ] && continue
 
-        if [ -f /bin/systemctl ]; then
+        if [ -f /bin/systemctl ] || [ -f /usr/bin/systemctl ]; then
```
myii: Actually, this is probably a nicer alternative that we're using in the `salt-image-builder` now:

```suggestion
        if command -v systemctl; then
```

myii:
> The main.yml is generated with .github/workflows/templates/generate.py
> 
> please update that file and make sure that the committed main.py is what is generated by the generate.py script.

@bryceml Done.  Must have missed the instructions to do that.

* I guess the `2019.2` tests will have to be fixed in a separate PR, due to the archiving that took place.
* I've actually already submitted a fix for the Ubuntu 16.04 failures:
  - https://github.com/saltstack/salt/pull/59340

bryceml:
https://github.com/saltstack/salt-bootstrap/pull/1526 should fix 2019.2

I'm talking to @s0undt3ch on the best way to handle pyzmq for the older releases.

bryceml:
I just merged my PR, so if you want to rebase yours, it should fix a bunch of issues.

myii:
> I just merged my PR, so if you want to rebase yours, it should fix a bunch of issues.

@bryceml Rebased, just waiting for all the tests to run.

myii:
@bryceml Arch failures have started before this PR: https://github.com/saltstack/salt-bootstrap/actions.

bryceml:
I'm looking into it.

bryceml:
@s0undt3ch as far as I can tell, the version of pip on arch right now is broken, should we just merge this now?  I suspect arch will update soon enough that we can just ignore it for now.

PR #1548, opened 2021-03-28
feat(bsd): add CI testing using Kitchen's `vagrant` driver

myii:
### What does this PR do?

Add CI testing for FreeBSD (12.2 & 11.4) and OpenBSD (6.8) using Kitchen's `vagrant` driver.  These have been configured for local testing as well.

Already using these in the SaltStack Formulas org so proposing it here as well.

I'll use inline comments to mention any noteworthy points.

### DIFF
--- a/.github/workflows/kitchen.vagrant.yml
+++ b/.github/workflows/kitchen.vagrant.yml
@@ -1,0 +1,87 @@
+---
+name: 'Kitchen Vagrant (FreeBSD & OpenBSD)'
+'on': ['push', 'pull_request']
+
+env:
+  KITCHEN_LOCAL_YAML: 'kitchen.vagrant.yml'
+
+jobs:
+  generate-actions-workflow:
+    name: 'Generate The Actions Workflow'
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: 'actions/checkout@v1'
+      - name: 'Set up Python 3.7'
+        uses: 'actions/setup-python@v1'
+        with:
+          python-version: 3.7
+      - name: 'Install Pre-Commit'
+        run: |
+          pip install -U pip
+          pip install pre-commit
+          pre-commit install
+      - name: 'Generate Workflow Actions'
+        run: |
+          pre-commit run -av generate-actions-workflow
+  lint:
+    name: 'Lint'
+    runs-on: 'ubuntu-latest'
+    needs: 'generate-actions-workflow'
+    container: 'koalaman/shellcheck-alpine:v0.6.0'
+    steps:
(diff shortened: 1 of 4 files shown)


kitchen.vagrant.yml
```diff
@@ -0,0 +1,44 @@
...
+    provisioner:
+      init_environment: |
+        echo 'auto_accept: true' > /tmp/auto-accept-keys.conf
+        sudo mkdir -p /etc/salt/master.d
+        sudo mv /tmp/auto-accept-keys.conf /etc/salt/master.d/auto-accept-keys.conf
+        sudo pkg_add coreutils
+        sh -c 't=$(gshuf -i 1-15 -n 1); echo Sleeping $t seconds; sleep $t'
+
+verifier:
+  command: pytest --cache-clear -v -s -ra --log-cli-level=debug -k "not test_ping" tests/integration/
```
myii, line 47: `test_ping` is actually working for local testing (on an Ubuntu 20.04 host).  It hangs until timeout in GitHub Actions, likely to be an issue with the macOS runner.  Haven't researched in great detail but these could be relevant:

* https://github.com/paramiko/paramiko/issues/721
* https://github.com/paramiko/paramiko/issues/735

.github/workflows/kitchen.vagrant.yml
```diff
@@ -0,0 +1,54 @@
...
+          # - py3-git-3001-openbsd-68
+          - py3-git-3002-freebsd-122
+          - py3-git-3002-freebsd-114
+          # - py3-git-3002-openbsd-68
+          - py3-git-master-freebsd-122
+          - py3-git-master-freebsd-114
+          # - py3-git-master-openbsd-68
+          - latest-freebsd-122
+          - latest-freebsd-114
+          - latest-openbsd-68
```
myii: The commented OpenBSD instances aren't currently working, since the bootstrap script needs some updating.  I've included them so that it's easier to test any changes, if there are efforts made to fix that.

krionbsd (approved):

bryceml (changes requested):
Same as the windows tests, I'd prefer shellcheck and generate-actions-workflow run first and this not run if those don't pass.

myii:
Updated with the following:

* Rebased.
* Added `generate-actions-workflow` and `lint` (isn't automated, though).
* Added FreeBSD 13.0.
* Renamed `openbsd-68` to `openbsd-6`, since `generic/openbsd6` now provides OpenBSD 6.9 (which actually installs `3002.6`, as opposed to only `3001.1` on OpenBSD 6.8).

PR #1549, opened 2021-03-28
feat(windows): add CI testing using Kitchen's `proxy` driver

myii:
### What does this PR do?

Add CI testing employing the Windows runners provided by GitHub (2019 & 2016) using Kitchen's `proxy` driver.

Already using these in the SaltStack Formulas org (thanks to the excellent contributions made by @dafyddj) so proposing it here as well.

I'll use inline comments to mention any noteworthy points.

### DIFF
--- a/.github/workflows/kitchen.windows.yml
+++ b/.github/workflows/kitchen.windows.yml
@@ -1,0 +1,150 @@
+---
+name: 'Kitchen (Windows)'
+'on': ['push', 'pull_request']
+
+env:
+  machine_user: kitchen
+  machine_pass: Password1
+  machine_port: 5985
+  KITCHEN_LOCAL_YAML: 'kitchen.windows.yml'
+
+jobs:
+  generate-actions-workflow:
+    name: 'Generate The Actions Workflow'
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: 'actions/checkout@v1'
+      - name: 'Set up Python 3.7'
+        uses: 'actions/setup-python@v1'
+        with:
+          python-version: 3.7
+      - name: 'Install Pre-Commit'
+        run: |
+          pip install -U pip
+          pip install pre-commit
+          pre-commit install
+      - name: 'Generate Workflow Actions'
+        run: |
+          pre-commit run -av generate-actions-workflow
+  lint:
+    name: 'Lint'
+    runs-on: 'ubuntu-latest'
+    needs: 'generate-actions-workflow'
+    container: 'koalaman/shellcheck-alpine:v0.6.0'
+    steps:
+      - uses: 'actions/checkout@v1'
+      - name: 'ShellCheck'
+        run: |
+          shellcheck -s sh -f tty bootstrap-salt.sh
+  test-2019:
+    runs-on: 'windows-2019'
+    timeout-minutes: 20
+    needs: 'lint'
+    strategy:
+      fail-fast: false
+      matrix:
+        instance:
+          - py3-stable-3001-windows-2019
+          - py3-stable-3002-windows-2019
+          - latest-windows-2019
+    steps:
+      - name: 'Check out code'
+        uses: 'actions/checkout@v2'
+      - name: 'Install Chef'
+        uses: 'actionshub/chef-install@1.1.0'
+        with:
+          project: 'chef'
+          version: '16.10.8'
+      - name: 'Add Chef bindir to PATH'
+        uses: 'myci-actions/export-env-var-powershell@1'
+        with:
+          name: 'PATH'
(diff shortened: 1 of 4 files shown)


kitchen.windows.yml
```diff
@@ -0,0 +1,31 @@
...
+driver:
+  name: proxy
+  host: localhost
+  reset_command: "exit 0"
+  port: 5985
+  username: kitchen
+  password: Password1
+
+provisioner:
+  salt_bootstrap_url: D:/a/salt-bootstrap/salt-bootstrap/bootstrap-salt.ps1
```
myii, line 11:
```suggestion
  salt_bootstrap_url: bootstrap-salt.ps1
```

^ I tried the obvious first but doesn't install the minion.  Probably a path issue.  Just giving the explicit path instead did the trick.

kitchen.windows.yml
```diff
@@ -0,0 +1,31 @@
...
+  init_environment: ''
+
+platforms:
+  - name: windows-2019
+  - name: windows-2016
+
+suites:
+  - name: py3-stable-3001
+    provisioner:
+      salt_version: 3001.7
```
myii, line 22: `3001` hangs while installing the minion, so went for the latest release instead.

kitchen.windows.yml
```diff
@@ -0,0 +1,31 @@
...
+  - name: py3-stable-3002
+    provisioner:
+      salt_version: 3002
+  - name: latest
+    provisioner:
+      salt_version: latest
+
+verifier:
+  command: pytest --cache-clear -v -s -ra --log-cli-level=debug -k "not test_ping" tests/integration/
```
myii, line 31: Unlike https://github.com/saltstack/salt-bootstrap/pull/1548#discussion_r602857775, we need to exclude the `test_ping` test here, since no master.

krionbsd (approved):

.github/workflows/kitchen.windows.yml
```diff
@@ -0,0 +1,121 @@
+---
```
bryceml, line 1: I'd prefer this go in .github/workflows/templates/windows.yml and end up part of the generated .github/workfows/main.yml similar to the lint.yml probably, but run at the same stage as the linux jobs.

Windows tests will be really nice to have.

.github/workflows/kitchen.windows.yml
```diff
@@ -0,0 +1,121 @@
...
+      fail-fast: false
+      matrix:
+        instance:
+          - py3-stable-3001-windows-2019
+          - py3-stable-3002-windows-2019
+          - latest-windows-2019
+    steps:
+      - name: 'Check out code'
+        uses: 'actions/checkout@v2'
+      - name: 'Install Chef'
```
bryceml, line 53: do we need chef?  can we tell kitchen to ignore it?

myii: Let's ask @dafyddj -- I'm sure he will be able to give a clear response to this query.

dafyddj: We install Chef not because we need or want Chef itself but to manage the dependencies for Test Kitchen.

In Formulas we came up against some Ruby dependency issue a short while back and at the very least we would need to upgrade Ruby in the GitHub Actions Windows image, but installing Chef Client manages all of that, and keeps us in-sync with upstream. 
In fact, Test Kitchen themselves say that, on Windows at least, the supported method of running Test Kitchen is via Chef Workstation/Chef Client.

.github/workflows/kitchen.windows.yml
myii: @bryceml Doesn't the same apply to #1548 as well?  Both PRs are using `strategy:matrix` so would have to be coerced in the way you've mentioned about `lint.yml`.

> but run at the same stage as the linux jobs

They run concurrently alongside the Linux jobs already.  I'm pretty sure having them inside the same workflow run will actually be slower overall.

.github/workflows/kitchen.windows.yml
krionbsd: yeah, let's go this way for a while, I do really hope GH actions will support win/fbsd one day, at least I've heard a lot of talks about it recently

.github/workflows/kitchen.windows.yml
bryceml: mainly I would want these not to run if lint doesn't pass.  github limits the number of concurrent things we can run at once, and I'd prefer not to waste that time if it doesn't pass lint (shellcheck).

Yeah, applies to #1548 too.

`strategy:matrix` could still work right?

myii:
Updated with the following:

* Rebased.
* Added `generate-actions-workflow` and `lint` (isn't automated, though).

Issue #1555, opened 2021-05-01
Best-effort support for Ubuntu 21.04


jpacura:
### Description of Issue/Question
The current salt-bootstrap script will not install on Ubuntu 21.04

### Steps to Reproduce Issue
Try to run the current version of bootstrap-salt.sh on a ubuntu:21.04 docker container
it will fail because the script will not override the repositories to use the 20.04 repositories
```
# sh bootstrap-salt.sh 
 *  INFO: Running version: 2021.03.02
 *  INFO: Executed by: sh
 *  INFO: Command line: 'bootstrap-salt.sh '
 *  WARN: Running the unstable version of bootstrap-salt.sh

 *  INFO: System Information:
 *  INFO:   CPU:          GenuineIntel
 *  INFO:   CPU Arch:     x86_64
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   5.10.0-6-amd64
 *  INFO:   Distribution: Ubuntu 21.04
...
Processing triggers for libc-bin (2.33-0ubuntu5) ...
 * ERROR: https://repo.saltproject.io/py3/ubuntu/21.04/amd64/latest/salt-archive-keyring.gpg failed to download to /tmp/salt-gpg-ldtAcOTA.pub
 * ERROR: Failed to run install_ubuntu_stable_deps()!!!
```

### Versions and Systems
```
# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 21.04
Release:	21.04
Codename:	hirsute

# sh bootstrap-salt.sh -v
bootstrap-salt.sh -- Version 2021.03.02
```

krionbsd:
Ubuntu 21.04 support was added recently

PR #1556, opened 2021-05-01
Add Ubuntu 21.04 support using Ubuntu 20.04 packages

jpacura:
### What does this PR do?
Adds support for Ubuntu 21.04 using Ubuntu 20.04 LTS repository packages

https://github.com/saltstack/salt-bootstrap/issues/1555

### Previous Behavior
Script will fail because it will not detect 21.04 as a non-LTS release, and will attempt to use a nonexistent repository for packages. 

### New Behavior
Script will succeed using the Ubuntu 20.04 saltstack repo. This is the expected behavior because non-LTS releases only get best-effort support using the previous LTS release's repositories

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -436,7 +436,7 @@ Debian and derivatives
 ~~~~~~~~~~~~~~~~~~~~~~
 
 - Cumulus Linux 2/3
-- Debian GNU/Linux 7/8/9/10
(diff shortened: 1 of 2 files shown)


krionbsd:
Thanks, 21.04 support added in #1559

PR #1573, opened 2021-07-14
Refactor `kitchen.yml` and update openSUSE Leap from `15.1` => `15.3`

myii:
### What does this PR do?

Clean up `kitchen.yml`, including less duplication and modernisation:

* Use the most common `run_command` of `/usr/lib/systemd/systemd` as a general setting.
* Remove unnecessary entries:
  - `image` for Fedora.
  - `salt_bootstrap_options` for openSUSE and Arch Linux.
* Use `driver` instead of `driver_config`:
  - https://github.com/test-kitchen/kitchen-azurerm/issues/69
  - https://github.com/test-kitchen/kitchen-azurerm/commit/55b768d0b3a32492c5f29b70e9490dc0be9f9570
* Update openSUSE Leap platform from `15.1` to `15.3`.
* Use YAML node anchors to remove further duplication:
  - Both Fedora and openSUSE `provision_command` entries are all duplicates that have been DRYed.

I kept the YAML node anchors change as a separate commit in case it needs to be removed; understandably, some folks prefer to avoid them.

### DIFF
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -9,6 +9,7 @@ driver:
     - sys_admin
   disable_upstart: false
   use_internal_docker_network: false
+  run_command: /usr/lib/systemd/systemd
 
 provisioner:
   name: salt_solo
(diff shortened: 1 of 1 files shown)


myii:
Only unexpected CI failures are the FreeBSD jobs.  I've submitted #1574 as a fix for those.

krionbsd (approved):

PR #1574, opened 2021-07-14
feat(freebsd): update packages from `py37-*` => `py38-*`

myii:
### What does this PR do?

Noticed the FreeBSD jobs failing when submitting #1573.  This PR updates the FreeBSD packages installed by the bootstrap to the `py38-*` versions.

### DIFF
--- a/.github/workflows/kitchen.vagrant.yml
+++ b/.github/workflows/kitchen.vagrant.yml
@@ -41,8 +41,8 @@ jobs:
       fail-fast: false
       matrix:
         instance:
-          - py3-git-3000-freebsd-130
-          - py3-git-3000-freebsd-122
+          # - py3-git-3000-freebsd-130
+          # - py3-git-3000-freebsd-122
           # - py3-git-3000-openbsd-6
           - py3-git-3001-freebsd-130
           - py3-git-3001-freebsd-122
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5843,15 +5843,15 @@ install_freebsd_git_deps() {
 
     if [ "${_POST_NEON_INSTALL}" -eq $BS_FALSE ]; then
 
-        SALT_DEPENDENCIES=$(/usr/local/sbin/pkg rquery %dn py37-salt)
+        SALT_DEPENDENCIES=$(/usr/local/sbin/pkg rquery %dn py38-salt)
         # shellcheck disable=SC2086
         /usr/local/sbin/pkg install -y ${SALT_DEPENDENCIES} python || return 1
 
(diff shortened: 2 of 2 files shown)


.github/workflows/kitchen.vagrant.yml
```diff
@@ -41,8 +41,8 @@ jobs:
       fail-fast: false
       matrix:
         instance:
-          - py3-git-3000-freebsd-130
-          - py3-git-3000-freebsd-122
+          # - py3-git-3000-freebsd-130
+          # - py3-git-3000-freebsd-122
```
myii, line 45: Before submitting this PR, I ran the tests in my fork and these two jobs were failing, so I disabled them for the time being:

* https://github.com/myii/salt-bootstrap/actions/runs/1029627093

Perhaps someone will be able to get them working again eventually.  Or maybe it doesn't even matter that much since `3000.x` is [EOL at the end of next month](https://saltproject.io/salt-platform-support/).

@krionbsd Any thoughts?

bootstrap-salt.sh
```diff
@@ -5843,15 +5843,15 @@ install_freebsd_git_deps() {
...
 
-        SALT_DEPENDENCIES=$(/usr/local/sbin/pkg rquery %dn py37-salt)
+        SALT_DEPENDENCIES=$(/usr/local/sbin/pkg rquery %dn py38-salt)
         # shellcheck disable=SC2086
         /usr/local/sbin/pkg install -y ${SALT_DEPENDENCIES} python || return 1
 
-        /usr/local/sbin/pkg install -y py37-requests || return 1
-        /usr/local/sbin/pkg install -y py37-tornado4 || return 1
+        /usr/local/sbin/pkg install -y py38-requests || return 1
+        /usr/local/sbin/pkg install -y py38-tornado4 || return 1
```
myii, line 5851: When searching for the packages, I found other options available here:

```console
# pkg search py38-tornado
py38-tornado-6.1               Python Web Server from FriendFeed
py38-tornado4-4.5.3            Python Web Server from FriendFeed
py38-tornado5-5.1.1            Python Web Server from FriendFeed
```

Can/should this be updated to use `py38-tornado5` instead?

.github/workflows/kitchen.vagrant.yml
krionbsd: We have to exclude 3000 from CI anyway, FreeBSD merged Python 3.8 to 12.2 as default python version, and I failed to use it with salt 3000

bootstrap-salt.sh
krionbsd: It's a bit more than that, we've to s/py37/py38 globally, I'd run it with tornado4 for the future

PR #1581, opened 2021-08-19
Allow this script to elevate permissions

ari:
If this script is not run as root, then run itself as root. This will work in situations where there is no root password (for example, in new EC2 FreeBSD ami)

FreeBSD does not ship with sudo, so salt-cloud when salt-cloud tries to run the bootstrap script it will fail. There is an undocumented attribute in salt-cloud profiles:

    sudo: False

and that will avoid sudo, but then this script runs on FreeBSD ami as ec2-user. So we need something to elevate permissions to root.

I'm using the t4g-micro AMI here ami-055a79b2e1da76e1e for testing. I've not tried this on other platforms and I know the su command is quite different between Linux and BSD.


bootstrap-salt.sh
```diff
@@ -638,7 +638,8 @@ fi
 
 # Root permissions are required to run this script
 if [ "$($whoami)" != "root" ]; then
-    echoerror "Salt requires root privileges to install. Please re-run this script as root."
+    echowarn "Salt requires root privileges to install. Re-running this script as root..."
+    exec "su -l root -c $0"
```
krionbsd, line 642: What if only sudo or doas are available?

krionbsd (changes requested):

bootstrap-salt.sh
ari: I guess it could go through a series of checks to see which one will work. I'm not knowledgeable enough about what common AMIs have installed on ec2 or for that matter on a range of other cloud platforms. Its not possible to have a FreeBSD system without su. What's the situation commonly on Linux?

This solves a particular problem in that its impossible to get FreeBSD to easily work with default salt-cloud options since it just assumes sudo is always available on all platforms.

ari: At any rate, this patch doesn't have to be perfect. Right now the script just fails if it isn't root. So failing only sometimes is an improvement.

ari:
Turns out that this approach doesn't work in salt-cloud, although it works standalone.

```
Uploading /tmp/tmpopxnkvhj to /tmp/.saltcloud-eab9e9e3-f2af-4d2a-9214-d49152eedb6e/deploy.sh
Warning: Permanently added '13.211.122.179' (ECDSA) to the list of known hosts.
Connection to 13.211.122.179 closed.
Warning: Permanently added '13.211.122.179' (ECDSA) to the list of known hosts.
 *  INFO: Running version: 2021.06.23
 *  INFO: Executed by: -
 *  INFO: Command line: '/tmp/.saltcloud-eab9e9e3-f2af-4d2a-9214-d49152eedb6e/deploy.sh -c /tmp/.saltcloud-eab9e9e3-f2af-4d2a-9214-d49152eedb6e'
 *  WARN: Salt requires root privileges to install. Re-running this script as root...
exec: su -l root -c /tmp/.saltcloud-eab9e9e3-f2af-4d2a-9214-d49152eedb6e/deploy.sh: not found
Connection to 13.211.122.179 closed.
Error: There was a profile error: Command 'ssh -t -t -oStrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null -oControlPath=none -oPasswordAuthentication=no -oChallengeResponseAuthentication=no -oPubkeyAuthentication=yes -oIdentitiesOnly=yes -oKbdInteractiveAuthentication=no -i /usr/local/etc/salt/cloud.providers.d/salt.pem  -p 22 ec2-user@13.211.122.179 '/tmp/.saltcloud-eab9e9e3-f2af-4d2a-9214-d49152eedb6e/deploy.sh -c '"'"'/tmp/.saltcloud-eab9e9e3-f2af-4d2a-9214-d49152eedb6e'"'"''' failed. Exit code: 127

```

It would seem that salt-cloud creates the temp executable file in such a way that it is deleted as soon as it runs, so this approach cannot work without salt-cloud being changed.

Do you want to keep this, or just abandon the pull request?

s0undt3ch:
If not a solution, abandon.

PR #1594, opened 2021-09-16
Re-add Ubuntu-16 support as it's still supported with 3001 and 3002

krionbsd:
### What does this PR do?
Re-add Ubuntu-16 support as it's still supported with 3001 and 3002

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1585,11 +1585,15 @@ __check_end_of_life_versions() {
         ubuntu)
             # Ubuntu versions not supported
             #
-            #  < 18.04
+            #  < 16.04
+            #  = 16.10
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -1585,11 +1585,16 @@ __check_end_of_life_versions() {
         ubuntu)
             # Ubuntu versions not supported
             #
-            #  < 18.04
+            #  < 16.04
+            #  = 16.10
+            #  = 17.04, 17.10
+            #  = 18.04
```
bryceml: 18.04 is supported

krionbsd: Fixed

bryceml (approved):

PR #1519, opened 2020-12-24
Add CentOS GPG key to RHEL7 builds

jeff350:
### What does this PR do?
This adds the feature requested in https://github.com/saltstack/salt-bootstrap/issues/1449 to add support for building on Red Hat's Universal Base image. The core issue here was that the GPG key for CentOS was needed for RHEL7 UBI. This MR adds the CentOS key to the saltstack repo in the __install_saltstack_rhel_repository function. 

### What issues does this PR fix or reference?
https://github.com/saltstack/salt-bootstrap/issues/1449

### Previous Behavior
the saltstack repo file did not include the CentOS GPG Key
```
[saltstack]
name=SaltStack 2019.2 Release Channel for RHEL/CentOS $releasever
baseurl=https://repo.saltstack.com/yum/redhat/7/$basearch/2019.2/
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://repo.saltstack.com/yum/redhat/7/$basearch/2019.2/SALTSTACK-GPG-KEY.pubenabled=1
enabled_metadata=1
```
### New Behavior
The key is not in place.
```
[saltstack]
name=SaltStack 2019.2 Release Channel for RHEL/CentOS $releasever
baseurl=https://repo.saltstack.com/yum/redhat/7/$basearch/2019.2/
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://repo.saltstack.com/yum/redhat/7/$basearch/2019.2/SALTSTACK-GPG-KEY.pub,https://repo.saltstack.com/yum/redhat/7/$basearch/2019.2/base/RPM-GPG-KEY-CentOS-7,
enabled=1
enabled_metadata=1
```

PR #1608, opened 2021-10-11
Add centos7 base key

bryceml:
### What does this PR do?
just the bootstrap script portion of https://github.com/saltstack/salt-bootstrap/pull/1519 since testing with ubi images doesn't seem to work currently, that portion can be handled later.


bootstrap-salt.sh
```diff
@@ -4179,21 +4192,21 @@ install_centos_stable_deps() {
         if [ "$DISTRO_MAJOR_VERSION" -ge 8 ]; then
             # YAML module is used for generating custom master/minion configs
             if [ -n "$_PY_EXE" ] && [ "$_PY_MAJOR_VERSION" -eq 3 ]; then
-                __PACKAGES="${__PACKAGES} python3-pyyaml"
+                __PACKAGES="${__PACKAGES} python3-pyyaml python3-setuptools"
```
bryceml, line 4195: @jeff350 why does python3-setuptools need to be added here?  Does it not work getting installed as a dependency of salt automatically?

jeff350: I did not test this recently but at the time I started this effort it was not getting installed automatically and the bootstrap script was failing because it was not installed

krionbsd (approved):

PR #1795, opened 2022-01-07
Enable Kali 2021

ptitdoc:
### What does this PR do?
Allowing bootstrapping Kali 2021 instances

Kali 2021 is not supported by salt-bootstrap script.


bootstrap-salt.sh
```diff
@@ -3771,6 +3772,13 @@ install_debian_git_post() {
     done
 }
 
+install_debian_2021_post() {
```
krionbsd, line 3776: Should we name the function `install_kali_2021_post` I guess it would be more appropriate

ptitdoc: I don't know if it will work, because functions are loaded based on the OS name (which is debian here because kali is a debian flavour).

krionbsd (approved):

Issue #1883, opened 2022-11-23
salt-bootstrap fails to start services when installing OneDir on AlmaLinux and Rocky Linux 9 due to presence of classic Salt packages in EPEL


ggiesen:
### Description of Issue/Question
salt-bootstrap fails to start services after installation  when using OneDir on Rocky Linux 9:

```
 *  INFO: Running daemons_running_onedir()
 * DEBUG: Waiting 3 seconds for processes to settle before checking for them
 * ERROR: /opt/saltstack/salt/run/run minion was not found running
 * ERROR: Failed to run daemons_running_onedir()!!!
```

Looks to be a similar issue to #1876 but for EL9 instead of EL8

### Setup
```
> mkdir salt-bootstrap-rockylinux-9

> cd salt-bootstrap-rockylinux-9

 salt-bootstrap-almalinux> vagrant init generic/rocky9
A `Vagrantfile` has been placed in this directory. You are now
ready to `vagrant up` your first virtual environment! Please read
the comments in the Vagrantfile as well as documentation on
`vagrantup.com` for more information on using Vagrant.

salt-bootstrap-rockylinux-9> vagrant up
Bringing machine 'default' up with 'vmware_desktop' provider...
==> default: Box 'generic/rocky9' could not be found. Attempting to find and install...
    default: Box Provider: vmware_desktop, vmware_fusion, vmware_workstation
    default: Box Version: >= 0
==> default: Loading metadata for box 'generic/rocky9'
    default: URL: https://vagrantcloud.com/generic/rocky9
==> default: Adding box 'generic/rocky9' (v4.2.2) for provider: vmware_desktop
    default: Downloading: https://vagrantcloud.com/generic/boxes/rocky9/versions/4.2.2/providers/vmware_desktop.box
    default:
    default: Calculating and comparing box checksum...
==> default: Successfully added box 'generic/rocky9' (v4.2.2) for 'vmware_desktop'!
==> default: Cloning VMware VM: 'generic/rocky9'. This can take some time...
==> default: Checking if box 'generic/rocky9' version '4.2.2' is up to date...
==> default: Verifying vmnet devices are healthy...
==> default: Preparing network adapters...
==> default: Fixed port collision for 22 => 2222. Now on port 2203.
==> default: Starting the VMware VM...
==> default: Waiting for the VM to receive an address...
==> default: Forwarding ports...
    default: -- 22 => 2203
==> default: Waiting for machine to boot. This may take a few minutes...
[...]

PR #1884, opened 2022-11-29
Removing EPEL, which is no longer needed.

garethgreenaway:
### What does this PR do?
Removing EPEL, which is no longer needed.

#1883


s0undt3ch (approved):

sachk12:
this fixed the issue I saw with rocky-linuc-9. When will this get into the stable branch?

Issue #1872, opened 2022-10-07
Can't install Salt on Debian Buster arm64


keenobi:
### Description of Issue/Question

I am trying to install salt on a Debian Buster machine with an arm64 architecture but I get an error when running the script :

```
root@host:~ # ./bootstrap-salt.sh stable 3005
[…]
*  WARN: Support for arm64 packages is experimental and might rely on architecture-independent packages from the amd64 repository.
[…]
* ERROR: https://repo.saltproject.io/py3/debian/10/arm64/3005/salt-archive-keyring.gpg failed to download to /tmp/salt-gpg-Oj5l62UU.pub
* ERROR: Failed to run install_debian_deps()!!!
```
Indeed the arm64 architecture is not supported in the official repo : [https://repo.saltproject.io/py3/debian/10/](https://repo.saltproject.io/py3/debian/10/)

This problem has already been discussed in issue #1240 and a PR (#1251) has been merged.
However, this fix was removed in a recent PR :  #1867.

I thought of several solutions :

- Add support for arm64 architectures
- Test if the architecture is supported for current distribution
  - if yes : Use current architecture
  - if no : Use amd64 architecture by default
- Other ?

### Setup

Debian 10 (arm64) host.

### Steps to Reproduce Issue

Run bootstrap script on Debian 10 with arm64 architecture.

### Versions and Systems

```
root@host:~ # ./bootstrap-salt.sh -v
./bootstrap-salt.sh -- Version 2022.10.04

root@host:~ # cat /etc/debian_version 
10.13

root@host:~ # dpkg --print-architecture
arm64
```

garethgreenaway:
@keenobi I just tested this one and it seems to be working with the latest bootstrap script.

keenobi:
Hello, I don't understand your answer. You tried to install Salt via the bootstrap script on a Debian 10 host with arm64 architecture.

I just tested and it doesn't work :

```
root@host:/tmp # wget https://bootstrap.saltproject.io -O /tmp/bootstrap-salt.sh
root@host:/tmp # /tmp/bootstrap-salt.sh -v
/tmp/bootstrap-salt.sh -- Version 2022.10.04
root@host:/tmp # /tmp/bootstrap-salt.sh stable 3005
 *  INFO: Running version: 2022.10.04
 *  INFO: Executed by: /bin/sh
 *  INFO: Command line: '/tmp/bootstrap-salt.sh stable 3005'

 *  INFO: System Information:
[...]

PR #1887, opened 2022-12-10
 Debian ARM Fix

garethgreenaway:
### What does this PR do?
Use amd64 repos for anything older than Debian 11 since the package sare arch-indepedant

#1872

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1442,9 +1442,13 @@ __check_dpkg_architecture() {
             if [ "$_CUSTOM_REPO_URL" != "null" ]; then
                 warn_msg="Support for arm64 is experimental, make sure the custom repository used has the expected structure and contents."
             else
-                # Saltstack official repository does not yet have arm64 metadata,
-                # use arm64 repositories on arm64, since all pkgs are arch-independent
-                __REPO_ARCH="arm64"
+                # Saltstack official repository has arm64 metadata beginning with Debian 11,
+                # use amd64 repositories on arm64 for anything older, since all pkgs are arch-independent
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -1442,9 +1442,13 @@ __check_dpkg_architecture() {
             if [ "$_CUSTOM_REPO_URL" != "null" ]; then
                 warn_msg="Support for arm64 is experimental, make sure the custom repository used has the expected structure and contents."
             else
-                # Saltstack official repository does not yet have arm64 metadata,
-                # use arm64 repositories on arm64, since all pkgs are arch-independent
-                __REPO_ARCH="arm64"
+                # Saltstack official repository has arm64 metadata beginning with Debian 11,
+                # use amd64 repositories on arm64 for anything older, since all pkgs are arch-independent
+                if [ "$DISTRO_NAME_L" = "debian" ] || [ "$DISTRO_MAJOR_VERSION" -lt 11 ]; then
```
dmurphy18, line 1447: Wondering how this will handle RHEL 9 on aarch64, see https://www.centos.org/download/, supported aarch64 even back in 7, not that I ever want to build classic packages for it.

Conditional needs a little work

s0undt3ch (approved):

bootstrap-salt.sh
garethgreenaway: @dmurphy18 This is only in the Debian section.

s0undt3ch: This is Debian related only.

s0undt3ch: Unless I misread the code.

PR #1920, opened 2023-04-21
Updates for Fedora and Photon

garethgreenaway:
### What does this PR do?
Updating bootstratp to install Salt built Fedora packages.  Initial support for installing on Photon via bootstrap script

### DIFF
--- a/.github/workflows/templates/generate.py
+++ b/.github/workflows/templates/generate.py
@@ -81,6 +81,9 @@ ONEDIR_DISTROS = [
     "centos-stream9",
     "debian-10",
     "debian-11",
+    "fedora-36",
+    "fedora-37",
+    "fedora-38",
     "oraclelinux-7",
     "oraclelinux-8",
     "photon-3",
@@ -574,10 +577,6 @@ def generate_test_jobs():
                     if salt_version in STABLE_VERSION_BLACKLIST:
                         continue
                     if distro not in STABLE_DISTROS:
-                        continue
-
-                    if distro.startswith("fedora") and salt_version != "latest":
-                        # Fedora does not keep old builds around
                         continue
 
                 if bootstrap_type == "git":
--- a/.github/workflows/templates/generate.py
+++ b/.github/workflows/templates/generate.py
@@ -255,9 +255,7 @@ BLACKLIST_GIT_3006 = [
     "ubuntu-2204",
 ]
 
-BLACKLIST_GIT_MASTER = [
-    "photon-4",
-]
+BLACKLIST_GIT_MASTER = []
 
 SALT_VERSIONS = [
     "3003",
--- a/.github/workflows/templates/generate.py
+++ b/.github/workflows/templates/generate.py
@@ -224,9 +224,6 @@ BLACKLIST_GIT_3005 = [
 
 BLACKLIST_3006 = [
     "arch",
-    "fedora-36",
-    "fedora-37",
-    "fedora-38",
     "gentoo",
     "gentoo-systemd",
     "opensuse-15",
--- a/.github/workflows/templates/generate.py
+++ b/.github/workflows/templates/generate.py
@@ -65,6 +65,8 @@ STABLE_DISTROS = [
     "opensuse-tumbleweed",
     "oraclelinux-7",
     "oraclelinux-8",
+    "photon-3",
+    "photon-4",
     "rockylinux-8",
     "ubuntu-2004",
     "ubuntu-2204",
@@ -81,6 +83,8 @@ ONEDIR_DISTROS = [
     "debian-11",
     "oraclelinux-7",
     "oraclelinux-8",
+    "photon-3",
+    "photon-4",
     "rockylinux-8",
     "rockylinux-9",
     "ubuntu-2004",
@@ -98,6 +102,8 @@ ONEDIR_RC_DISTROS = [
     "debian-11",
     "oraclelinux-7",
     "oraclelinux-8",
+    "photon-3",
+    "photon-4",
     "rockylinux-8",
     "rockylinux-9",
     "ubuntu-2004",
@@ -117,6 +123,8 @@ BLACKLIST_3003 = [
     "gentoo-systemd",
     "opensuse-15",
     "opensuse-tumbleweed",
+    "photon-3",
+    "photon-4",
     "rockylinux-8",
     "rockylinux-9",
     "ubuntu-2204",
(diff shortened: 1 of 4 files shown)


bootstrap-salt.sh
```diff
@@ -4185,6 +4194,62 @@ install_debian_check_services() {
...
+        gpg_key_urls=$(printf "${base_url}${key},%s" "$gpg_key_urls")
+    done
+
+    repo_file="/etc/yum.repos.d/salt.repo"
+
+    if [ ! -s "$repo_file" ] || [ "$_FORCE_OVERWRITE" -eq $BS_TRUE ]; then
+        cat <<_eof > "$repo_file"
+[saltstack]
+name=SaltStack ${repo_rev} Release Channel for Fedora \$releasever
+baseurl=${base_url}
```
s0undt3ch: Why not download the existing repo files on our packages repo?

garethgreenaway: @s0undt3ch initial pass at this, will rework the functions to use the repo files once things are semi-working.

s0undt3ch: 👍

.github/workflows/templates/generate.py
```diff
@@ -229,20 +240,25 @@
...
+    "photon-3",
+    "photon-4",
     "rockylinux-9",
     "ubuntu-2004",
     "ubuntu-2110",
     "ubuntu-2204",
 ]
 
+BLACKLIST_GIT_MASTER = [
+    "photon-4",
```
s0undt3ch: We can't install from git master on PhotonOS 4?

bootstrap-salt.sh
```diff
@@ -4185,6 +4194,43 @@ install_debian_check_services() {
...
+    else
+        repo_rev="latest"
+    fi
+
+    __PY_VERSION_REPO="yum"
+    if [ -n "$_PY_EXE" ] && [ "$_PY_MAJOR_VERSION" -eq 3 ]; then
+        __PY_VERSION_REPO="py3"
+    fi
+
+    gpg_key="SALT-PROJECT-GPG-PUBKEY-2023.pub"
```
s0undt3ch:
```suggestion
    GPG_KEY="SALT-PROJECT-GPG-PUBKEY-2023.pub"
```

bootstrap-salt.sh
```diff
@@ -4185,6 +4194,43 @@ install_debian_check_services() {
...
+
+    repo_file="/etc/yum.repos.d/salt.repo"
```
s0undt3ch:
```suggestion
    REPO_FILE="/etc/yum.repos.d/salt.repo"
```

bootstrap-salt.sh
```diff
@@ -4185,6 +4194,43 @@ install_debian_check_services() {
...
+
+    if [ ! -s "$repo_file" ] || [ "$_FORCE_OVERWRITE" -eq $BS_TRUE ]; then
+        fetch_url="${HTTP_VAL}://${_REPO_URL}/${_ONEDIR_DIR}/${__PY_VERSION_REPO}/fedora/${DISTRO_MAJOR_VERSION}/${CPU_ARCH_L}/${ONEDIR_REV}"
+        if [ "${ONEDIR_REV}" = "nightly" ] ; then
+            fetch_url="${HTTP_VAL}://${_REPO_URL}/${_ONEDIR_NIGHTLY_DIR}/${__PY_VERSION_REPO}/fedora/${DISTRO_MAJOR_VERSION}/${CPU_ARCH_L}/"
```
s0undt3ch: This will still be broken as we  will now have at least two sub paths to account to, `master` and `<release-branch>`(3006.x).
And then we'll also have a date, today's date.
We should probably have a `latest` along with the dates, so it's simpler for scripts to get latest nightly.....
Anyway, this is still being worked on, so, while it's ok to commit this now, it will be broken soon.

garethgreenaway: @s0undt3ch This brokeness is why I disabled all the nightly tests for now.

.github/workflows/templates/generate.py
garethgreenaway: @s0undt3ch There was a traceback during the install about missing libraries, I wanted to get the initial support merged in and then investigate that later.

bootstrap-salt.sh
s0undt3ch: 👍

bootstrap-salt.sh
```diff
@@ -4185,6 +4194,43 @@ install_debian_check_services() {
...
+    if [ -n "$_PY_EXE" ] && [ "$_PY_MAJOR_VERSION" -eq 3 ]; then
+        __PY_VERSION_REPO="py3"
+    fi
+
+    GPG_KEY="SALT-PROJECT-GPG-PUBKEY-2023.pub"
+
+    REPO_FILE="/etc/yum.repos.d/salt.repo"
+
+    if [ ! -s "$REPO_FILE" ] || [ "$_FORCE_OVERWRITE" -eq $BS_TRUE ]; then
+        fetch_url="${HTTP_VAL}://${_REPO_URL}/${_ONEDIR_DIR}/${__PY_VERSION_REPO}/fedora/${DISTRO_MAJOR_VERSION}/${CPU_ARCH_L}/${ONEDIR_REV}"
```
s0undt3ch:
```suggestion
        FETCH_URL="${HTTP_VAL}://${_REPO_URL}/${_ONEDIR_DIR}/${__PY_VERSION_REPO}/fedora/${DISTRO_MAJOR_VERSION}/${CPU_ARCH_L}/${ONEDIR_REV}"
```
Variables should always be uppercased to better distinguish from functions.

bootstrap-salt.sh
```diff
@@ -4185,6 +4194,43 @@ install_debian_check_services() {
...
+        if [ "${ONEDIR_REV}" = "nightly" ] ; then
+            fetch_url="${HTTP_VAL}://${_REPO_URL}/${_ONEDIR_NIGHTLY_DIR}/${__PY_VERSION_REPO}/fedora/${DISTRO_MAJOR_VERSION}/${CPU_ARCH_L}/"
+        fi
+
+        __fetch_url "${REPO_FILE}" "${fetch_url}.repo"
+
+        for key in $GPG_KEY; do
```
s0undt3ch: `$GPG_KEY` can be an array or a list of strings?
Why do we itereate?

s0undt3ch (changes requested):
Sorry, some more review comments.

bootstrap-salt.sh
garethgreenaway: I copied both the Photon and Fedora functions from the CentOS ones, which at one point had support for multiple keys.  I've updated both to only use the one.

s0undt3ch (approved):

Issue #1919, opened 2023-04-21
Bootstrap-salt.sh 3006 error


simon-xu-gao:
### Description of Issue/Question
Running bootstrap-salt.sh, would not install due to an error. Worked one I downgraded to the previous stable version (boostrap-salt.sh stable 3005)

### Setup
(Please provide relevant configs (Be sure to remove sensitive info).)

### Steps to Reproduce Issue
Running version: 2023.04.06
Executed by: sh
Command line: '/tmp/bootstrap-salt.sh '
Running the unstable version of bootstrap-salt.sh
/tmp/bootstrap-salt.sh: 620: 1: parameter not set

### Versions and Systems
Ubuntu-22.04 via Windows WSL.

connectwise-cody-krukauskas:
also experiencing this same issue trying to launch Ubuntu 22.04 EC2 machines via salt-cloud on a 3006 master.

Arderos:
Same here

garethgreenaway:
Thanks for the report.  We just released a new version of the bootstrap script, can you try the latest version and see if that resolves the issue for you?  Thanks!

neomancernz:
Good afternoon. I just downloaded redownloaded the script via curl, I am getting the same issue. (stable 3005 worked)

Debian 11 container
![image](https://user-images.githubusercontent.com/128402973/233810691-d6b06ac8-2d69-4f87-bc78-db98634d9f88.png)

simon-xu-gao:
> Thanks for the report. We just released a new version of the bootstrap script, can you try the latest version and see if that resolves the issue for you? Thanks!

Original issue was also freshly downloaded off of bootstrap.saltproject.io

Getting the same issue running version 2023.04.21 directly from the git repo:

```
$ git clone https://github.com/saltstack/salt-bootstrap.git  
$ cd salt-bootstrap/
$ bash bootstrap-salt.sh 
 *  INFO: Running version: 2023.04.21
 *  INFO: Executed by: bash
 *  INFO: Command line: 'bootstrap-salt.sh '
 *  WARN: Running the unstable version of bootstrap-salt.sh
bootstrap-salt.sh: line 620: $1: unbound variable
```

garethgreenaway:
Thanks everyone, I was able to reproduce it.  Also appears to show a missing window in our testing for bootstrap.
This diff should fix it and we'll make sure the fix is available in the next bootstrap release.

```
[...]

PR #1921, opened 2023-04-24
 Fixing defaults

garethgreenaway:
### What does this PR do?
Fixing a bug when no type or version is specified.  Adding tests forwhen no type or version is passed.

#1923 #1919

### DIFF
--- a/.github/workflows/templates/generate.py
+++ b/.github/workflows/templates/generate.py
@@ -50,6 +50,7 @@ BSD = [
 
 STABLE_DISTROS = [
     "almalinux-8",
+    "almalinux-9",
     "amazon-2",
     "arch",
     "centos-7",
@@ -174,7 +175,9 @@ BLACKLIST_3004 = [
 ]
 
 BLACKLIST_3005 = [
+    "almalinux-9",
     "arch",
+    "centos-stream9",
     "fedora-36",
     "fedora-37",
     "fedora-38",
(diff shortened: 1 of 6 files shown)


.github/workflows/templates/generate.py
```diff
@@ -482,6 +488,12 @@ def generate_test_jobs():
                 kitchen_target = f"{bootstrap_type}-{salt_version}"
                 instances.append(kitchen_target)
 
+        for bootstrap_type in ("default",):
+            if distro not in STABLE_DISTROS:
+                continue
+            kitchen_target = f"{bootstrap_type}"
+            instances.append(kitchen_target)
```
s0undt3ch:
```suggestion
            if distro not in STABLE_DISTROS:
                continue
            instances.append(bootstrap_type)
```
We don't really need another variable here.

.github/workflows/templates/generate.py
```diff
@@ -522,6 +534,12 @@ def generate_test_jobs():
...
```
s0undt3ch:
```suggestion
            if distro not in STABLE_DISTROS:
                continue
            instances.append(bootstrap_type)
```

.github/workflows/templates/generate.py
```diff
@@ -613,6 +631,12 @@ def generate_test_jobs():
...
```
s0undt3ch:
```suggestion
            if distro not in STABLE_DISTROS:
                continue
            instances.append(bootstrap_type)
```

s0undt3ch (changes requested):

s0undt3ch (approved):

PR #1969, opened 2023-10-05
Don't sort lists and dicts, as order of items matters

monkz:
### Previous Behavior
If sort_keys is not set to False, pyyaml reorders the data structure alphabetically. While this might me asthetically pleasing, it interferes (at least) with order of entries for gitfs.

### New Behavior
Keeps data in original order


dmurphy18:
@monkz Closing this PR in favor of PR https://github.com/saltstack/salt-bootstrap/pull/2006 which contains your fix.  Doing this for speed, as have a number of fixes to release, since the major cleanup in release 2024-07-12 and want to get this simple fix in.

PR #1981, opened 2024-01-02
Add arm support for Debian 10

twangboy:
### What does this PR do?
Adds arm64 support for Debian 10

https://github.com/saltstack/salt-bootstrap/issues/1980


dmurphy18 (approved):
This support is for Salt 3006 and up only, no ARM64 support for classic package and Tiamat-backend Salt 3005.x and Debian 10

Issue #1991, opened 2024-03-02
Rocky Linux 8 Container is missing find command


blu-base:
### Description of Issue/Question

For bootstraps from version 3006.0 there seems to be a scriptlet which is expecting the `find`command.
However, in the container for Rocky Linux 8 apparently is no `find` command available.

The ubiquitous `findutils` is preinstalled in all RHEL-based Distro (-containers), but Rocky Linux 8... 

https://github.com/saltstack/salt-bootstrap/actions/runs/7628876701/job/20781257386
```
        *  INFO: Running install_rocky_linux_onedir()
       Last metadata expiration check: 0:00:02 ago on Tue Jan 23 16:45:34 2024.
       Dependencies resolved.
       ================================================================================
        Package             Architecture Version                 Repository       Size
       ================================================================================
       Installing:
        salt-master         x86_64       3006.5-0                saltstack       3.4 M
        salt-minion         x86_64       3006.5-0                saltstack       235 k
       Installing dependencies:
        dmidecode           x86_64       1:3.3-5.el8             baseos           91 k
        hwdata              noarch       0.314-8.19.el8          baseos          1.8 M
        openssl             x86_64       1:1.1.1k-9.el8_7        baseos          709 k
        pciutils            x86_64       3.7.0-3.el8             baseos          104 k
        pciutils-libs       x86_64       3.7.0-3.el8             baseos           53 k
        salt                x86_64       3006.5-0                saltstack        57 M
       
       Transaction Summary
       ================================================================================
       Install  8 Packages
       
       Total download size: 64 M
       Installed size: 183 M
       Downloading Packages:
       (1/8): dmidecode-3.3-5.el8.x86_64.rpm           195 kB/s |  91 kB     00:00    
       (2/8): pciutils-3.7.0-3.el8.x86_64.rpm          880 kB/s | 104 kB     00:00    
       (3/8): pciutils-libs-3.7.0-3.el8.x86_64.rpm     449 kB/s |  53 kB     00:00    
       (4/8): hwdata-0.314-8.19.el8.noarch.rpm         1.7 MB/s | 1.8 MB     00:01    
       (5/8): openssl-1.1.1k-9.el8_7.x86_64.rpm        601 kB/s | 709 kB     00:01    
       (6/8): salt-3006.5-0.x86_64.rpm                  88 MB/s |  57 MB     00:00    
[...]

PR #1987, opened 2024-02-02
Added support for Amazon 2023, Debian 12, Ubuntu 24.04, removed EOL and BSD OSs and Python 2.7 support

dmurphy18:
### What does this PR do?
Adds support for Amazon 2023, Debian 11 and 12.
Removes support for Debian 7, 8 & 9, RHEL 6, Amazon  Linux (RHEL 6 clone) and Python 2.7

### What issues does this PR fix or reference?
Fixes: https://github.com/saltstack/salt-bootstrap/issues/1940, https://github.com/saltstack/salt-bootstrap/issues/1986

### Previous Behavior
Amazon 2023, Debian 11 and 12 failed to install with bootstrap script

### New Behavior
Amazon 2023, Debian 11 and 12 now install with bootstrap script.
Debian 7, 8 and 9, RHEL 6, Amazon Linux (RHEL 6 clone) and Python 2.7 are no longer supported

PR #1992, opened 2024-03-02
Ensure find command available in RHEL-based distros

blu-base:
### What does this PR do?

The Rocky Linux 8 container doesn't ship the package `findutils`.
Since salt 3006 there seems to be a scriptlet, in the the salt rpm and ondir package which requires the `find` command.

This PR adds the `findutils` package to the install dependencies for all RHEL-based distros.
In almost all cases, this package is pre-installed, and therefore doesn't seem to be a significant impact.

#1991

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -4808,7 +4808,7 @@ install_centos_stable_deps() {
         fi
     fi
 
-    __PACKAGES="${__PACKAGES} procps"
+    __PACKAGES="${__PACKAGES} procps findutils"
 
     # shellcheck disable=SC2086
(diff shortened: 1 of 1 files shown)


dmurphy18:
See comment https://github.com/saltstack/salt-bootstrap/issues/1991#issuecomment-1977153151
TL;DR wondering why this is an issue for Salt, and not the container, which should have **find**

dmurphy18 (approved):

dmurphy18:
@blu-base Closing this since resolved by PT https://github.com/saltstack/salt-bootstrap/pull/1987 and now available in current stable release (2024-07-12)

PR #2000, opened 2024-05-24
Fix installing salt 3007 as stable or onedir

sblaisot:
### What does this PR do?

When trying to install salt 3007 as onedir or stable specifying only the major (which means the latest minor in this major serie for salt 3005 and 3006), under debian ot ubuntu, salt-bootstrap fails because it looks for the repo signing key iin `minor/3007` instead of just `3007`

This PR fixes that behaviour and let you install salt 3007 major as onedir or stable

### Previous Behavior

salt-bootstrap onedir 3007 fails on debian or ubuntu

### New Behavior

salt-bootstrap onedir 3007 works on debian or ubuntu

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -630,7 +630,7 @@ elif [ "$ITYPE" = "stable" ]; then
         _ONEDIR_REV="latest"
         ITYPE="onedir"
     else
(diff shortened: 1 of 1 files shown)


sblaisot:
@dmurphy18 any news on this?

sblaisot:
Fix has been implemented in #1987

dmurphy18:
@sblaisot Waiting for S3 bucket to sync with web site, clearing cache to induce sync at the moment

PR #2005, opened 2024-07-15
Update bootstrap-salt.sh

javatask:
On arm64 platforms, bootstrap fails to finish installation due to missing variable 'error_msg=""' in case statement. This change adds a single line to fix the bug.

### What does this PR do?
Add a single line with 'error_msg=""' to arm64 check statment

Fix installation on arm64 architecture.

### Previous Behavior
Before, the script failed to deploy salt on the arm64 platform due to error on 1409 LoC

### New Behavior
Successfully, now I can install salt on arm64

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -1389,6 +1389,7 @@ __check_dpkg_architecture() {
             ;;
         "arm64")
             # Saltstack official repository has full arm64 support since 3006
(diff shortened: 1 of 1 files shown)


dmurphy18:
@javatask Thanks for the catch

dmurphy18 (approved):

Issue #2027, opened 2024-10-31
salt-bootstrap scripts broken: Need a new release to work with packages.broadcom.com repos


ScriptAutomate:
### Description of Issue/Question

`salt-bootstrap` needs updating to work with the new `packages.broadcom.com` repo hosting, layout, and package management layer of version pinning (apt, dnf), the removal of all Salt versions before Salt 3006 LTS, etc. since the migration (Salt Project Blog post: [Salt Project Package Repository (repo.saltproject.io) Migration and Guidance](https://saltproject.io/blog/salt-project-package-repo-migration-and-guidance/) and [Salt Install Guide Overhaul and salt-bootstrap updates](https://saltproject.io/blog/salt-install-guide-overhaul/)).

The Salt Install Guide has been updated to show how people can install on target systems without `salt-bootstrap` at the moment, and can be used as a reference point for updating `salt-bootstrap` logic:

- https://docs.saltproject.io/salt/install-guide/en/latest/

### Follow-up steps

This also means, upon completion, that documentation in both this repo and the https://github.com/saltstack/salt-install-guide will need updates based on changes in the script behavior.

- https://github.com/saltstack/salt-install-guide/issues/3

ipaqmaster:
Same here. This has been handled no less than horribly. 783 servers failed to provision today due to failing to make a connection with repo.saltproject.io. All it needed was a 302 redirect but not even that could be achieved with Broadcom.

This is a mess and good reason to migrate states to Ansible.

sanket925:
Ridiculous and completely unprofessional change from Broadcom unless this is a security issue. The infrastructure to support new repo is not updated before the closure of existing repositories.  This has impacted our production deployments resulting in  SEV-A situations in a holiday week.

ghost:
I spent half an hour trying to figure out why all my machines' DNS are blowing up and can't resolve `repo.saltproject.io`, turns out it's because Broadcom intentionally broke it. 💀

FFS, add a fucking 302 redirection.

aholyoake-bc:
A 302 wouldn't work by the looks - the deb repo structure is completely different.
Absolute shambles :rage4:

tjyang:
Waiting for a newer  bootstrap-salt.sh that use broadcom URL.

```
root@ubuntu22t01:~# bash bootstrap-salt.sh -v
bootstrap-salt.sh -- Version 2024.09.24
[...]

PR #2039, opened 2024-11-07
Update bootstrap shell CI/CD to utilize containers and clean up

dmurphy18:
### What does this PR do?
https://github.com/saltstack/salt-bootstrap/issues/2038

### What issues does this PR fix or reference?
Fixes https://github.com/saltstack/salt-bootstrap/issues/2038

Issue #2038, opened 2024-11-07
Update bootstrap shell CI/CD to utilize containers and clean up


dmurphy18:
### Description of Issue/Question
Update bootstrap shell CI/CD to utilize containers and clean up

Need to enable CI/CD for shell bootstrap-salt for Linux and MacOS. Release v2024.11.07 was tested by hand in order to speed delivery with CI/CD testing disabled.  This issue tracks fixing CI/CD to utilize containers not that the AWS infrastructure that was previously used, is no longer available.

Limits of the v2024.11.07 release:
- Support for Broadcom packaging infrastructure, packaging.broadcom.com
- Photon OS support only for latest Salt, that is, Salt 3007.1 (problem with Photon support for repo file exclude parameter)
- Linux platforms only support latest major branch versions (similar to Salt Install Guide),  for example: 3006.9, 3007.1
- Windows and MacOS support minor branch versions, for example: 3006.8
- Future fixes for Linux minor versions of Salt will be forth-coming

TODO:

- Need to adjust Photon OS minor version support
- Need to adjust Linux OS minor version support
- Optimize script to eliminate duplication
- Remove support for OS which are no longer relevant, for example: gentoo

PR #2041, opened 2024-11-08
Allow to use minor versions on Debian

didiermfb:
### What does this PR do?
Allow minor versions for Debian
Can't test on other OS

Related to #2027

### Previous Behavior
Bootstrap used only the major version to pin package on Debian

### New Behavior
Bootstrap now use the good pin version.
`./bootstrap-salt.sh stable 3006.5`
Result : 
```
$ cat /etc/apt/preferences.d/salt-pin-1001
Package: salt-*
Pin: version 3006.5
Pin-Priority: 1001
```

`./bootstrap-salt.sh stable 3006`
Result : 
```
$ cat /etc/apt/preferences.d/salt-pin-1001
Package: salt-*
Pin: version 3006.*
Pin-Priority: 1001
```

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2994,14 +2994,13 @@ __install_saltstack_ubuntu_repository() {
 
     if [ "$STABLE_REV" != "latest" ]; then
         # latest is default
-        STABLE_REV_MAJOR=$(echo "$STABLE_REV" | cut -d '.' -f 1)
-        if [ "$STABLE_REV_MAJOR" -eq "3006" ]; then
+        if [ "$(echo "$STABLE_REV" | grep -E '^(3006|3007)$')" != "" ]; then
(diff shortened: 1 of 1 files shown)


dmurphy18:
@didiermfb Thanks for that, but marking as duplicate , see https://github.com/saltstack/salt-bootstrap/pull/2039 and https://github.com/saltstack/salt-bootstrap/issues/2038

But I shall take a look at your submission for doing Debian minor versions, have RedHat family and Photon OS to do also.  But will get to that after getting the CI/CD build containers working.

Last release was done by hand in order to get a somewhat working script out, need the CI/CD testing working before I can make the additional changes, otherwise too much time testing.

dmurphy18:
@didiermfb Thanks for that, added your change and testing is good, just got to get rpm doing similar.

PR #2055, opened 2024-11-26
fix almalinux

jeff350:
### What does this PR do?
In the most recent release (2024.11.21) the `almalinux` install functions were renamed to `rockylinux`, however the necessary `rocky_linux` functions already existed for Rockylinux (see next section in code). The rename of these functions broke the ability for almalinux installs to work. This PR renames the functions back to almalinux so that bootstrap will again work on both almalinux and rockylinux.

### Previous Behavior
Bootstrap is broken on almalinux after 2024.11.21 release

### New Behavior
Installs work on both rockylinux and almalinux

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5112,79 +5112,79 @@ install_oracle_linux_check_services() {
 
 #######################################################################################################################
 #
-#   RockyLinux Install Functions
+#   AlmaLinux Install Functions
 #
-install_rockylinux_stable_deps() {
+install_almalinux_stable_deps() {
     install_centos_stable_deps || return 1
     return 0
 }
 
-install_rockylinux_git_deps() {
+install_almalinux_git_deps() {
     install_centos_git_deps || return 1
     return 0
 }
 
-install_rockylinux_onedir_deps() {
+install_almalinux_onedir_deps() {
     install_centos_onedir_deps || return 1
     return 0
 }
 
-install_rockylinux_testing_deps() {
+install_almalinux_testing_deps() {
     install_centos_testing_deps || return 1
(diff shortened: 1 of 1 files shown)


dmurphy18:
@jeff350 Not going to happen, Rocky Linux -> Alma Linux, as I stated previously, Rocky Linux is the now designated testing platform for RedHat families and that decision was made summer of 2024.

However, I will add the AlmaLinux back in to since it is causing problems, which was not the intent. Thought it would finally default into using Centos the way that Rocky Linux does.  

Working on git install issues with Rocky Linux 9, and will add the AlmaLinux support back in, in addition to Rocky Linux, similar to your PR. Cannot give timeline, since working on git install and when it passes testing.

Closing this PR as a result, but feel free to re-open if there is additional information.

dmurphy18:
See my PR https://github.com/saltstack/salt-bootstrap/pull/2057 for your changes implemented in conjunction with Rocky Linux

jeff350:
@dmurphy18 please take a closer look at this PR. This already contains both the rocky and almaLinux functions as you are stating is needed, and I agree. 

The PR you reference renamed the `almalinux` functions to `rockylinux`, This change was not necessary as the `rocky_linux` functions (with underscore) already existed. See line 5220 in PR #2057. After the  2024.11.21 release both `rocky_linux` and `rockylinux` functions exist. the `rockylinux` functions are not used or needed. This PR renames the functions back to almalinux to restore bootstrap's ability to work on both rocky and almalinux as was supported in v2024.11.07.

dmurphy18:
@jeff350 That is what I did renamed the rockylinux back to almalinux

PR #2068, opened 2024-12-04
Add support for specifying a directory other than /tmp for installation execution

edlitmus:
### What does this PR do?

Adds an option to specify something other than `/tmp` for installation execution.

Systems that have been hardened to CIS-2 benchmarks should have `/tmp` and `/var/tmp` mounted with the `noexec` option. This causes installations to fail when trying to execute anything from `/tmp`.

### New Behavior

Using the new option allows installations to proceed as expected. An example using the `ubuntu` user home directory:

```shell
bash /home/ubuntu/bootstrap-salt.sh -X -U -A localhost -T /home/ubuntu onedir "$SALT_VERSION"
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@
 minimum_pre_commit_version: 1.15.2
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.5.0
+    rev: v5.0.0
     hooks:
       - id: check-merge-conflict  # Check for files that contain merge conflict strings.
       - id: trailing-whitespace   # Trims trailing whitespace.
--- a/README.rst
+++ b/README.rst
@@ -196,6 +196,9 @@ To view the latest options and descriptions for ``salt-bootstrap``, use ``-h`` a
     -r  Disable all repository configuration performed by this script. This
         option assumes all necessary repository configuration is already present
         on the system.
+    -T  If set this overrides the use of /tmp for script execution. This is
+        to allow for systems in which noexec is applied to temp filesystem mounts
+        for security reasons
     -U  If set, fully upgrade the system prior to bootstrapping Salt
     -v  Display script version
     -V  Install Salt into virtualenv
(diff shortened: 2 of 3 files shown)


dmurphy18:
@edlitmus You need to refresh, released v2024.12.12 today, fix the lint errors by running the following in your work directory ```pre-commit install```, this way you will catch lint errors, etc when you git commit.

And lastly, you need to add tests to ensure your changes to code are tested.

edlitmus:
> @edlitmus You need to refresh, released v2024.12.12 today, fix the lint errors by running the following in your work directory `pre-commit install`, this way you will catch lint errors, etc when you git commit.

Thanks, I refreshed and fixed the lint errors.

> And lastly, you need to add tests to ensure your changes to code are tested.

Any guides on adding tests for this? I'm not sure where to start on that.

dmurphy18:
@edlitmus Adding tests using pytest, similar to Salt itself, except don't have to use nox.  Take a look at https://github.com/saltstack/salt-bootstrap/blob/develop/tests/integration/test_installation.py and also Salt test and pytest  documentation

- https://docs.saltproject.io/en/latest/topics/development/tests/index.html
- https://docs.saltproject.io/en/latest/topics/tutorials/writing_tests.html
- https://docs.pytest.org/en/stable/
- https://www.tutorialspoint.com/pytest/index.htm

Lastly, if you add a feature to Linux, you cannot ignore Windows, needs to be added there too, that is, bootstrap-salt.ps1 needs updating too.

edlitmus:
@dmurphy18 Sorry for the delay on this, I was taking time off, but I did some digging into the Windows side of this, and it doesn't look like the options for the Windows script are a match for the shell script. Also, I don't think it makes sense to add the same feature to the Windows script as I do not believe that `noexec` exists for NTFS, so it wouldn't be something that would be required for the installation on Windows systems.

As for the tests, I still need to look into how to add something to do validations.

twangboy (approved):

twangboy (approved):

ed-silva-eb:
Sorry for the lack of additional tests but I didn't see how to add any that would be effective. FWIW we have been using this version in production since January.

twangboy:
Would you mind rebasing and fixing the conflicts?

ed-silva-eb:
> Would you mind rebasing and fixing the conflicts?

I was able to merge cleanly info my fork from the upstream develop branch with no conflicts.

Issue #2079, opened 2025-07-18
[BUG] installing salt-minion on PhotonOS fails due to wrong naming in installer


erwindon:
**Description**
The bootstrap installer fails on PhotonOS 4 and 5.
The reason is that the installer assumes the name "photon".
But the installer tries to find "vmware_photon_os".
These names are used to find the internal functions in the insaller.

Changing the installer script as follow makes it work again.
`sed --in-place 's/_photon_/_vmware_photon_os_/' /tmp/bootstrap-salt.sh`

**Setup**
VirtualBox with PhotonOS 4
or
VirtualBox with PhotonOS 5

Please be as specific as possible and give set-up details.
- [ ] on-prem machine
- [X] VM (Virtualbox 7.12)
- [ ] VM running on a cloud service, please be explicit and add details
- [ ] container (Kubernetes, Docker, containerd, etc. please specify)
- [ ] or a combination, please be explicit
- [ ] jails if it is FreeBSD
- [ ] classic packaging
- [ ] onedir packaging
- [X] used bootstrap to install

**Steps to Reproduce the behavior**
* Install PhotonOS 4 or 5
* Download bootstrapsalt.sh (using https://raw.githubusercontent.com/saltstack/salt-bootstrap/refs/heads/develop/bootstrap-salt.sh)
* run bootstrap-salt.sh
This step fails.

**Expected behavior**
running bootstrap-salt.sh succeeds (and leaves a working minion).

**Screenshots**
```
 *  INFO: Running version: 2025.02.24
 *  INFO: Executed by: sh
 *  INFO: Command line: '/tmp/bootstrap-salt.sh '

 *  INFO: System Information:
 *  INFO:   CPU:          GenuineIntel
 *  INFO:   CPU Arch:     x86_64
 *  INFO:   OS Name:      Linux
 *  INFO:   OS Version:   6.1.143-2.ph5
 *  INFO:   Distribution: VMware Photon OS 5.0

 *  INFO: Installing minion
 *  INFO: Found function config_salt
 *  INFO: Found function preseed_master
 *  INFO: Found function daemons_running_onedir
 * ERROR: No dependencies installation function found. Exiting...
```

**Versions Report**
(salt-minion is not installed yet, thus salt-master version is irrelevant)

**Additional context**
(none)

twangboy:
Please create a PR with your suggested change

erwindon:
[...]

Issue #2078, opened 2025-07-18
[BUG] PhotonOS 4 install using bootstrap-salt.sh lacks WGET


erwindon:
**Description**
When installing a salt-minion on PhotonOS 4, the bootstrap installer tries to run WGET which is not installed yet.

**Setup**
Fresh minimal install of PhotonOS 4

Please be as specific as possible and give set-up details.

- [ ] on-prem machine
- [X] VM (Virtualbox 7.12)
- [ ] VM running on a cloud service, please be explicit and add details
- [ ] container (Kubernetes, Docker, containerd, etc. please specify)
- [ ] or a combination, please be explicit
- [ ] jails if it is FreeBSD
- [ ] classic packaging
- [ ] onedir packaging
- [X] used bootstrap to install

**Steps to Reproduce the behavior**
* Install PhotonOS 4 (in minimal install, I did not try other editions)
* Download bootstrap-salt.sh and run it
* See the installer fail with a missing WGET

**Expected behavior**
Installer detects missing WGET and installs it first.

**Versions Report**
No salt-minion installed yet (therefore salt-master version is also irrelevant)

**Additional context**
This problem does not occur in PhotonOS 5.

Note that the bootstrap installer also has another problem, see #68196. This report assumes that that one is fixed.

twangboy:
What is available on a minimum install of Photon OS 4 for downloading stuff?

erwindon:
I'll look into that again.
(IIRC `wget` is available, but it is not downloaded (using a package installer) before it is used)

erwindon:
the solution is indeed to just install "wget".

But I was mistaken about the fact that this did not occur for Photon5.
That does not change the solution as both share the same installer code.

I have prepared a PR for #2078 (this) and #2079 as both are needed for successful installation on Photon.

PR #2080, opened 2025-07-30
fixes for Photon4/5

erwindon:
### What does this PR do?
Fixed installation on Photon 4 and Photon 5.

https://github.com/saltstack/salt-bootstrap/issues/2078
https://github.com/saltstack/salt-bootstrap/issues/2079

1) Photon identifies itself as "vmware_photon_os". But the installer was assuming that it was just "photon". Fixed by renaming.
2) Photon (at least in minimal-install) does not yet have `wget`. But the installer tries to use it. Fixed by adding dependency on `wget`.

Note that I'm not actually a user of Photon. I have VirtualBox VMs for all operating systems that are supported by SaltStack, so that I can test my project [SaltGUI](https://github.com/erwindon/SaltGUI).

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -6610,7 +6610,7 @@ install_vmware_photon_os_onedir_deps() {
         __install_saltstack_vmware_photon_os_onedir_repository || return 1
     fi
 
-    __PACKAGES="procps-ng sudo shadow"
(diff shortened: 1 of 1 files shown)


erwindon:
@twangboy fyi

twangboy (approved):

PR #2114, opened 2026-06-01
Fix checking out a branch on in existing repo

heewa:
### What does this PR do?

In the case where a git repo already exists:
* if a branch is intended to be checked out, reset to `origin/<branch>`
    * this mirrors the behavior of the repo not existing [on these lines, a little further down](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L2214-L2217)
    * it avoids problems stemming from a local branch by that name already existing, which may not even be related, or could have diverged from `origin`
* remove the unnecessary code that rebases from `origin`, since we just directly check it out (which happens after a `fetch` updates it a few lines up)

When a user first runs `bootstrap-salt.sh` with a branch specified (without a checked out repo) works fine, but sets up `HEAD` in a detached state to `origin/<branch>`. Subsequent runs of the bootstrap script fail when it tries to `git reset --hard <branch>` , with err:

> fatal: ambiguous argument '<branch>': unknown revision or path not in the working tree.

### Previous Behavior

Remove this section if not relevant

### New Behavior

Remove this section if not relevant

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -2158,12 +2158,13 @@ __git_clone_and_checkout() {
         fi
 
         # Check if GIT_REV_ADJ is a remote branch or just a commit hash
+        __GIT_CHECKOUT_REV="$GIT_REV_ADJ"
         if git branch -r | grep -q -F -w "origin/$GIT_REV_ADJ"; then
-            GIT_REV_ADJ="origin/$GIT_REV_ADJ"
+            __GIT_CHECKOUT_REV="origin/$GIT_REV_ADJ"
         fi
 
-        echodebug "Hard reseting the cloned repository to ${GIT_REV_ADJ}"
-        git reset --hard "$GIT_REV_ADJ" || return 1
+        echodebug "Hard reseting the cloned repository to ${__GIT_CHECKOUT_REV}"
+        git reset --hard "$__GIT_CHECKOUT_REV" || return 1
     else
         if [ "$_FORCE_SHALLOW_CLONE" -eq "${BS_TRUE}" ]; then
             echoinfo "Forced shallow cloning of git repository."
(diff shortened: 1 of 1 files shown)


bootstrap-salt.sh
```diff
@@ -2145,18 +2145,13 @@ __git_clone_and_checkout() {
             git fetch --tags upstream
         fi
 
+        # Check if GIT_REV_ADJ is a remote branch or just a commit hash
+        if git branch -r | grep -q -F -w "origin/$GIT_REV_ADJ"; then
+            GIT_REV_ADJ="origin/$GIT_REV_ADJ"
+        fi
+
```
twangboy, line 2165: "While this fixes the detached HEAD issue for an existing repository, permanently mutating GIT_REV_ADJ by prepending origin/ introduces two side effects downstream:

Global Scope Pollution: GIT_REV_ADJ is used throughout the rest of the script. Changing it here means any subsequent steps that expect the raw branch name or commit SHA will now explicitly see the origin/ prefix.

Upstream Remote Breakage: Just above this block (lines 2138–2146), the script explicitly handles setups tracking upstream (the official SaltStack repo) instead of origin. If a user is tracking an upstream branch, hardcoding origin/ here will cause the git reset --hard on line 2154 to fail because origin/<branch> won't exist or won't be up to date.

Instead of mutating the global variable, we should dynamically evaluate the ref check inside the reset logic using git rev-parse, checking both origin and upstream remotes safely."

```bash
if git rev-parse --verify "origin/${GIT_REV_ADJ}" >/dev/null 2>&1; then
    git reset --hard "origin/${GIT_REV_ADJ}" || return 1
elif git rev-parse --verify "upstream/${GIT_REV_ADJ}" >/dev/null 2>&1; then
    git reset --hard "upstream/${GIT_REV_ADJ}" || return 1
else
    git reset --hard "${GIT_REV_ADJ}" || return 1
fi
```

PR #2117, opened 2026-06-03
Write alpine service scripts directly

heewa:
### What does this PR do?

Writes alpine service files for salt daemons directly, rather than trying to download them from the salt repo (which deprecated them a while ago). This mirrors [how it's done for gentoo](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L7682-L7698).

Alpine install is broken. In the `install_alpine_linux_post()` phase, it [tries to install](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L5495) OpenRC scripts from `<salt-repo>/pkg/alpine/...`, but they're not there anymore. They were moved to [`old/`](https://github.com/saltstack/salt/tree/master/pkg/old) when they got deprecated [3.5 years ago](https://github.com/saltstack/salt/commit/701633d181).

### DIFF
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5535,15 +5535,22 @@ install_alpine_linux_post() {
         [ $fname = "syndic" ] && [ "$_INSTALL_SYNDIC" -eq $BS_FALSE ] && continue
 
         if [ -f /sbin/rc-update ]; then
-            script_url="${_SALTSTACK_REPO_URL%.git}/raw/master/pkg/alpine/salt-$fname"
-            [ -f "/etc/init.d/salt-$fname" ] || __fetch_url "/etc/init.d/salt-$fname" "$script_url"
+            local script_path="/etc/init.d/salt-$fname"
+            if ! [ -f "$script_path" ]; then
+                cat <<_eof > "$script_path"
+#!/sbin/openrc-run
+command="/usr/bin/salt-${fname}"
+command_args="--daemon"
+pidfile="/var/run/salt-${fname}.pid"
+name="Salt ${fname} daemon"
 
-            # shellcheck disable=SC2181
-            if [ $? -eq 0 ]; then
-                chmod +x "/etc/init.d/salt-$fname"
-            else
-                echoerror "Failed to get OpenRC init script for $OS_NAME from $script_url."
-                return 1
+depend() {
+        need localmount
+        use net
+        after bootmisc
+}
+_eof
+                chmod +x "$script_path"
             fi
 
             # Skip salt-api since the service should be opt-in and not necessarily started on boot
--- a/bootstrap-salt.sh
+++ b/bootstrap-salt.sh
@@ -5535,7 +5535,7 @@ install_alpine_linux_post() {
         [ $fname = "syndic" ] && [ "$_INSTALL_SYNDIC" -eq $BS_FALSE ] && continue
 
         if [ -f /sbin/rc-update ]; then
-            local script_path="/etc/init.d/salt-$fname"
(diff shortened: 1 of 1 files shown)


twangboy (approved):

twangboy:
Please rebase and address the lint failures

bootstrap-salt.sh
```diff
@@ -5492,15 +5492,22 @@ install_alpine_linux_post() {
         [ $fname = "syndic" ] && [ "$_INSTALL_SYNDIC" -eq $BS_FALSE ] && continue
 
         if [ -f /sbin/rc-update ]; then
-            script_url="${_SALTSTACK_REPO_URL%.git}/raw/master/pkg/alpine/salt-$fname"
-            [ -f "/etc/init.d/salt-$fname" ] || __fetch_url "/etc/init.d/salt-$fname" "$script_url"
+            local script_path="/etc/init.d/salt-$fname"
+            if ! [ -f "$script_path" ]; then
+                cat <<_eof > "$script_path"
```
twangboy, line 5540: Because the heredoc delimiter ``_eof`` is unquoted, the current shell will evaluate and expand all the variables at the time the bootstrap script runs, instead of writing the literal variable strings into the target file.

For example, line 5499 (``command="/usr/bin/salt-${fname}"``) will literally write ``command="/usr/bin/salt-minion"`` into the init file. While that works fine for execution, lines like 5500 (``command_args="--daemon"``) are okay, but lines 5501 and 5502 will expand things prematurely based on the bootstrap execution environment rather than leaving them dynamic for OpenRC.

To pass the text completely literally and allow OpenRC to handle its own variable definitions cleanly, we should quote the delimiter.

Suggested Change:
```bash
cat << '_eof' > "$script_path"
```

bootstrap-salt.sh
```diff
@@ -5492,15 +5492,22 @@ install_alpine_linux_post() {
         [ $fname = "syndic" ] && [ "$_INSTALL_SYNDIC" -eq $BS_FALSE ] && continue
 
         if [ -f /sbin/rc-update ]; then
-            script_url="${_SALTSTACK_REPO_URL%.git}/raw/master/pkg/alpine/salt-$fname"
-            [ -f "/etc/init.d/salt-$fname" ] || __fetch_url "/etc/init.d/salt-$fname" "$script_url"
+            local script_path="/etc/init.d/salt-$fname"
```
twangboy: While ``script_path`` is correctly defined as a local variable, ``fname`` is used throughout this block without being explicitly declared local at the top of the ``install_alpine_linux_post()`` function. If it isn't already, we should ensure fname is scoped locally to avoid leaking or mutating global state.

bootstrap-salt.sh
```diff
@@ -5492,15 +5492,22 @@ install_alpine_linux_post() {
...
-            if [ $? -eq 0 ]; then
-                chmod +x "/etc/init.d/salt-$fname"
-            else
-                echoerror "Failed to get OpenRC init script for $OS_NAME from $script_url."
-                return 1
+depend() {
+        need localmount
+        use net
+        after bootmisc
+}
```
twangboy, line 5551: Once the heredoc delimiter is quoted, these lines will write nicely to the file. However, standard ShellCheck guidelines for OpenRC scripts recommend ensuring variables inside the definitions are safe. It looks clean, but just double-check that the indentation within the ``depend()`` block (lines 5504-5508) matches the standard 4-space tab or 2-space layout preferred by Alpine's upstream openrc styles.

twangboy (changes requested):

twangboy (approved):

PR #2124, opened 2026-08-03
fix(bootstrap): rewrite salt.sources URIs to honor -R custom repo URL on Debian/Ubuntu

### LINKED ISSUE #2123: Custom repository URL not working for Debian/Ubuntu
The *custom repository URL* option does not install a suitable sources file for Ubuntu.

```
-R  Specify a custom repository URL. Assumes the custom repository URL
        points to a repository that mirrors Salt packages located at
        packages.broadcom.com. [...]
```

The `-R` option propagates its value into `_REPO_URL` via `_CUSTOM_REPO_URL`, overriding the default Broadcom Artifactory. For Fedora, RHEL, Amazon Linux, Photon, SUSE and MacOS, `_REPO_URL` is then used to compose a repo definition file, but for Debian/Ubuntu, it simply downloads a `salt.sources` file from GitHub and pulls only the apt signing key from `_REPO_URL`, leaving the system still dependent on the Broadcom Artifactory repo.

It's already running one post-processing step with `sed` to change the keyring file suffix, but likely needs to make replacements of `packages.broadcom.com/artifactory` with the custom URL.

(glossing over the naming of this variable as a URL when it isn't, which can also cause confusion)

twangboy:
### What does this PR do?
The Debian/Ubuntu repo functions downloaded a salt.sources template that hardcodes packages.broadcom.com/artifactory in its URIs line. Only the GPG key fetch used $_REPO_URL, so -R had no effect on which apt repo packages actually came from. Add a sed rewrite of the URIs line to $_REPO_URL, mirroring how RHEL/SUSE/Photon already parameterize their repo files.

Fixes #2123

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -63,6 +63,26 @@ jobs:
         run: |
           /usr/bin/docker start ${{ github.run_id }}_salt-test
 
+      - name: "Wait for container ${{ inputs.container-slug }} to finish booting"
+        run: |
+          # systemd's default tmpfiles.d rules wipe /tmp during early boot
+          # (systemd-tmpfiles-setup.service). Running pip/bootstrap before
+          # that finishes races the wipe and fails with spurious
+          # "No such file or directory" errors on /tmp/pip-*-tracker-*.
+          for i in $(seq 1 90); do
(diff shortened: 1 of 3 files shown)


hackery:
Awesome turnaround, tyvm.

